PR #17, opened 2015-06-05
Update TableMapReduceUtil.java

bertpassek:
TableMapReduceUtil is an utility class. I can't get it why these handy methods convertStringToScan and convertScanToString are package protected? 

We need to setup server side filters for mapreduce jobs so we must provide a scan object to TableInputFormat and we need those methods to put it into the job configuration. It just does not make sense.


busbey:
Could you please open an issue on our project tracker describing the issue and attach a patch:

https://issues.apache.org/jira/browse/HBASE

for more information on submitting contributions to the project, please read through our new contributor section:

http://hbase.apache.org/book.html#getting.involved

busbey:
This same change was made in [HBASE-15223](https://issues.apache.org/jira/browse/HBASE-15223), and will be available in HBase 1.3.0+ and 2.0.0+

PR #26, opened 2016-02-04
HBASE-11262

ashubhumca:
In bulk import, it just stores everything in the table but some times we want to avoid empty column values to save some disk space. So, this code snippet to skip those empty columns values while bulk load based on the user's choice. By default, this feature will be disabled. User has to set SKIP_EMPTY_COLUMNS true to use this.

### DIFF
--- a/hbase-resource-bundle/src/main/resources/supplemental-models.xml
+++ b/hbase-resource-bundle/src/main/resources/supplemental-models.xml
@@ -1195,4 +1195,22 @@ Copyright (c) 2007-2011 The JRuby project
       </licenses>
     </project>
   </supplement>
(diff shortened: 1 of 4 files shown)


tedyu:
HBase infra hasn't switched to accommodating pull request.

Review board is currently used.

Patch looks good.

ashubhumca:
Thanks for the acknowledgement. Please let me know if I've to do something else in order to add this patch.

busbey:
yes, please attach your patch to the jira for HBASE-11262 and close this PR.

ashubhumca:
Closing this PR as patch has already been added via jira.

PR #40, opened 2016-10-13
HBASE-16469 Several log refactoring/improvement suggestions 

nemo9cby:
(This patch originally focused on release 1.2.2, but these following three files changed since then and the original logging code with issues does not exist. We modified the patch to be against master)

hbase-client\src\main\java\org\apache\hadoop\hbase\client\AsyncProcess.java
hbase-server\src\main\java\org\apache\hadoop\hbase\ipc\RpcServer.java
hbase-server\src\main\java\org\apache\hadoop\hbase\master\AssignmentManager.java


dimaspivak:
Hi @nemo9cby,

Thanks so much for the patch. Would you mind resubmitting it following [our normal patch process](https://hbase.apache.org/book.html#submitting.patches)? We don't accept GitHub PRs.

nemo9cby:
Hi @dimaspivak 

Sorry for the trouble. Sure, I have already uploaded the patch in JIRA following the process. Thanks!

busbey:
apologies, I accidentally closed this PR instead of #42. please continue with this on [HBASE-16469](https://issues.apache.org/jira/browse/HBASE-16469).

PR #41, opened 2016-10-29
HBASE-NNNN Clarify Scan comment about caching

jkleckner:
The Scan class has an integer field, caching, that defaults to -1.
The meaning of -1 is that the option was not overriden (aka not set) and
therefore that the default configuration should be used.

This trivial patch corrects the comment of that field to reflect this
meaning rather than misleading the reader to believe that caching is
disabled.


dimaspivak:
Hey @jklecker,

Thanks so much for creating a patch! On Apache HBase, we require all contributors to use our issue tracker and to submit patches by attaching them to individual JIRAs, as described in our ref guide. Would you mind closing this PR and posting your patch through that channel?

Thanks again,
Dima

jkleckner:
Happy to do so.  I went there first and couldn't find "create issue" in the places I usually see it in JIRA.  Finally found it by going to up to the root.

Created as https://issues.apache.org/jira/browse/HBASE-16970

PR #48, opened 2017-04-12
When a huge amount of data writing to hbase through thrift2, there wi…

AlbertLee1:
…ll be a deadlock error.

Because of the conneciton cache and htablepool cache does not time out at the same time.

Here is my patch https://patch-diff.githubusercontent.com/raw/apache/hbase/pull/48.patch


busbey:
duplicate of #49 I think?

PR #49, opened 2017-04-12
When a huge amount of data writing to hbase through thrift2, there wi…

AlbertLee1:
…ll be a deadlock error.

Because the conneciton cache and htablepool cache use the same timeout.
So we need a refresh mechanism to clean connectioncache.


busbey:
Hi Albert!

The HBase project uses the [HBASE tracker in ASF JIRA instance](https://issues.apache.org/jira/browse/HBASE) to review and apply work. Would you mind filing a request there that describes the problem and proposed solution?

Our [contributor guide section includes instructions on how to attach a patch to your issue once it exists](http://hbase.apache.org/book.html#submitting.patches).

Thanks for your interest in improving Apache HBase!

PR #56, opened 2017-07-28
HBASE-18470: Fix a bug in `RetriesExhaustedWithDetailsException#getDesc` describe

asdf2014:
The describe from `RetriesExhaustedWithDetailsException#getDesc` is `
org.apache.hadoop.hbase.client.RetriesExhaustedWithDetailsException: Failed 3 actions: FailedServerException: 3 times, `, there is a not need ', ' in the tail.


asdf2014:
HBase project doesn't use PR model on github... Closed!

PR #59, opened 2017-09-03
HBASE-18743 Fixed TableName#valueOf issue

wenbang:
HBASE-18743
TableName#valueOf  issue may cause hfiles missing
fix it


wenbang:
close  request

PR #64, opened 2017-10-11
HBASE-18901 [C++] Provide CMAKE infrastructure

phrocker:
* Provided cmake files for packages in which a default module
did not exist.
* Moved tests to a location where we could automatically build
the test suite.
* Resolved minor issues with tests
* Tested across OSX, RHEL7, and Ubuntu16

HBASE-18901 [C++] Cleanup CMAKE


phrocker:
Note that some of the cmake modules are derivative of previous work that I've done, which was likely derivative work of someone else's work years past. How would you like me to handle this? 

Even if I didn't immediately reference or copy their cmake module I'd be happy to document somewhere that something may be a derivative work as a result of someone else's efforts that I've propagated via memory.

busbey:
This PR is slated to be closed as a part of [HBASE-20391](https://issues.apache.org/jira/browse/HBASE-20391) because the referenced JIRA has since closed.

If this is incorrect and you are still interested in working on the issue please comment there and/or reopen the PR.

phrocker:
I think @joshelser can confirm that this was merged via another PR. I will close as a result. Thanks

phrocker:
I guess I could confirm too, but I don't really know how to use git....what's the command?

PR #70, opened 2017-12-19
HBASE-19547: HBase fails building on AArch64 due to asciidoctor-maven…

guyuqi:
This error took place because asciidoctor-maven-plugin used an old version
of Jruby (1.7.26) which didn’t work with AArch64 (due to the fact that
Jffi 1.2.12 which Jruby 1.7.26 depends on has no native AArch64 libraries).
This patch resolves the issue by upgradeing Jruby to version 9.1.10.0 that supports AArch64.

Change-Id: I17c2372434cc536f9436aa9e101ecf27571a5623
Signed-off-by: Yuqi Gu <yuqi.gu@arm.com>

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -1234,6 +1234,11 @@
             <artifactId>asciidoctorj-pdf</artifactId>
             <version>${asciidoctorj.pdf.version}</version>
           </dependency>
+          <dependency>
+            <groupId>org.jruby</groupId>
+            <artifactId>jruby-complete</artifactId>
+            <version>${jruby.version}</version>
+          </dependency>
         </dependencies>
         <configuration>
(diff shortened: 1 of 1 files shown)


vielmetti:
Was this closed and then not merged and then closed with no explanation?

busbey:
not at all. the comment on the closing ref:

> * closes #70 - the referenced JIRA, HBASE-19547, has already closed.

that jira (the same one in the title ): [HBASE-19547](https://issues.apache.org/jira/browse/HBASE-19547) says the change was merged and made it into the 2.0.0 release.

busbey:
FTR, it was merged in April 2018.

vielmetti:
My apologies, I had looked at only a portion of the referenced closing document!

PR #73, opened 2018-03-02
HBASE-20070 refactor website generation

busbey:
* rely on git plumbing commands when checking if we've built the site for a particular commit already
* switch to forcing '-e' for bash
* add command line switches for: path to hbase, working directory, and publishing
* only export JAVA/MAVEN HOME if they aren't already set.
* add some docs about assumptions
* Update javadoc plugin to consistently be version 3.0.0
* avoid duplicative site invocations on reactor modules
* update use of cp command so it works both on linux and mac
* manually skip enforcer plugin during build
* still doing install of all jars due to MJAVADOC-490, but then skip rebuilding during aggregate reports.
* avoid the pager on git-diff by teeing to a log file, which also helps later reviewing in the case of big changesets.

### DIFF
--- a/dev-support/jenkins-scripts/generate-hbase-website.sh
+++ b/dev-support/jenkins-scripts/generate-hbase-website.sh
@@ -23,74 +23,176 @@
 #
 # It needs to be built on a Jenkins server with the label git-websites
 #
-# It expects to have the hbase repo cloned to the directory hbase
+# Allows specifying options for working directory, maven repo, and publishing to git
+# run with --help for usage.
 #
 # If there is a build error, the Jenkins job is configured to send an email
 
-LOCAL_REPO=${WORKSPACE}/.m2/repo
-# Nuke the local maven repo each time, to start with a known environment
-rm -Rf "${LOCAL_REPO}"
-mkdir -p "${LOCAL_REPO}"
+declare CURRENT_HBASE_COMMIT
+declare PUSHED
+declare FILE
+declare WEBSITE_COMMIT_MSG
+declare -a FILES_TO_REMOVE
 
-# Clean any leftover files in case we are reusing the workspace
-rm -Rf -- *.patch *.patch.zip hbase/target target *.txt hbase-site
+set -e
+function usage {
+  echo "Usage: ${0} [options] /path/to/hbase/checkout"
+  echo ""
+  echo "    --working-dir /path/to/use  Path for writing logs and a local checkout of hbase-site repo."
+  echo "                                if given must exist."
+  echo "                                defaults to making a directory in /tmp."
+  echo "    --local-repo /path/for/maven/.m2  Path for putting local maven repo."
+  echo "                                if given must exist."
+  echo "                                defaults to making a clean directory in --working-dir."
+  echo "    --publish                   if given, will attempt to push results back to the hbase-site repo."
+  echo "    --help                      show this usage message."
+  exit 1
+}
+# if no args specified, show usage
+if [ $# -lt 1 ]; then
+  usage
+fi
+
+# Get arguments
+declare component_dir
+declare working_dir
+declare local_repo
+declare publish
+while [ $# -gt 0 ]
+do
+  case "$1" in
+    --working-dir) shift; working_dir=$1; shift;;
+    --local-repo) shift; local_repo=$1; shift;;
+    --publish) shift; publish="true";;
+    --) shift; break;;
+    -*) usage ;;
+    *)  break;;  # terminate while loop
+  esac
+done
+
+# should still have where component checkout is.
+if [ $# -lt 1 ]; then
+  usage
+fi
+component_dir="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
+
+if [ -z "${working_dir}" ]; then
+  echo "[DEBUG] defaulting to creating a directory in /tmp"
+  working_dir=/tmp
+  while [[ -e ${working_dir} ]]; do
+    working_dir=/tmp/hbase-generate-website-${RANDOM}.${RANDOM}
+  done
+  mkdir "${working_dir}"
+else
+  # absolutes please
+  working_dir="$(cd "$(dirname "${working_dir}")"; pwd)/$(basename "${working_dir}")"
+  if [ ! -d "${working_dir}" ]; then
+    echo "passed working directory '${working_dir}' must already exist."
+    exit 1
+  fi
+fi
+
+echo "You'll find logs and temp files in ${working_dir}"
+
+if [ -z "${local_repo}" ]; then
+  echo "[DEBUG] defaulting to creating a local repo within '${working_dir}'"
+  local_repo="${working_dir}/.m2/repo"
+  # Nuke the local maven repo each time, to start with a known environment
+  rm -Rf "${local_repo}"
+  mkdir -p "${local_repo}"
+else
+  # absolutes please
+  local_repo="$(cd "$(dirname "${local_repo}")"; pwd)/$(basename "${local_repo}")"
+  if [ ! -d "${local_repo}" ]; then
+    echo "passed directory for storing the maven repo '${local_repo}' must already exist."
+    exit 1
+  fi
+fi
 
 # Set up the environment
-export JAVA_HOME=$JDK_1_8_LATEST__HOME
-export PATH=$JAVA_HOME/bin:$MAVEN_3_3_3_HOME/bin:$PATH
-export MAVEN_OPTS="-XX:MaxPermSize=256m -Dmaven.repo.local=${LOCAL_REPO}"
+if [ -z "${JAVA_HOME}" ]; then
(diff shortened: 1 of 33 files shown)


mistyhacks:
Nit: The path to HBase is not a switch but a positional argument. This actually tripped me up the first time I tried using the new syntax.

Reviewing now.

busbey:
I did that since it's required. old habit.

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -23,73 +23,173 @@
...
+# should still have where component checkout is.
+if [ $# -lt 1 ]; then
+  usage
+fi
+component_dir="$(cd "$(dirname "$1")"; pwd)/$(basename "$1")"
+
+if [ -z "${working_dir}" ]; then
+  working_dir=/tmp
+  while [[ -e ${working_dir} ]]; do
+    working_dir=/tmp/hbase-generate-website-${RANDOM}.${RANDOM}
```
mistyhacks: Maybe emit that we are using a default value, for every time we use a default value because we didn’t find a flag?

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -23,73 +23,173 @@
...
+if [ -z "${local_repo}" ]; then
+  local_repo="${working_dir}/.m2/repo"
+  # Nuke the local maven repo each time, to start with a known environment
+  rm -Rf "${local_repo}"
+  mkdir -p "${local_repo}"
+else
+  # absolutes please
+  local_repo="$(cd "$(dirname "${local_repo}")"; pwd)/$(basename "${local_repo}")"
+  if [ ! -d "${local_repo}" ]; then
+    echo "passed directory for storing the maven repo  '${local_repo}' must already exist."
```
mistyhacks: Not extra space after “repo”

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -23,73 +23,173 @@
...
+  JAVA_HOME="${JDK_1_8_LATEST__HOME}"
+  export JAVA_HOME
+  export PATH="${JAVA_HOME}/bin:${PATH}"
+fi
+if [ -z "${MAVEN_HOME}" ]; then
+  MAVEN_HOME="${MAVEN_3_3_3_HOME}"
+  export MAVEN_HOME
+  export PATH="${MAVEN_HOME}/bin:${PATH}"
+fi
+export MAVEN_OPTS="-XX:MaxPermSize=256m -Dmaven.repo.local=${local_repo}"
```
mistyhacks: I think the MaxPermSize is deprecated and the script bitches about it each time it runs.

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -23,73 +23,173 @@
...
+# Fail if it's empty
+if [ -z "${CURRENT_HBASE_COMMIT}" ]; then
+  echo "Got back a blank answer for the current HEAD on the remote hbase repository. failing."
+  exit 1
+fi
 echo "Current HBase commit: $CURRENT_HBASE_COMMIT"
 
 # Clone the hbase-site repo manually so it doesn't trigger spurious
 # commits in Jenkins.
 git clone --depth 1 --branch asf-site https://git-wip-us.apache.org/repos/asf/hbase-site.git
```
mistyhacks: We don’t verify that the clone worked and yet we do stuff below assuming that it did work.

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -23,73 +23,173 @@
...
+    --batch-mode \
+    -Denforcer.skip=true \
+    -Dmaven.main.skip=true \
+    -Dmaven.test.skip=true \
+    -DskipTests \
+    --log-file="${working_dir}/hbase-site-log-${CURRENT_HBASE_COMMIT}.txt"; then
+  echo "Successfully built site."
+else
+  status=$?
   echo "Failure: mvn clean site"
```
mistyhacks: Nit this is out of date now.

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -125,46 +225,62 @@ FILES_TO_REMOVE=("hbase-*"
...
-  # Create a patch, which Jenkins can save as an artifact and can be examined for debugging
-  git format-patch --stdout origin/asf-site > "${WORKSPACE}/${CURRENT_HBASE_COMMIT}.patch"
-  echo "Change set saved to patch ${WORKSPACE}/${CURRENT_HBASE_COMMIT}.patch"
-  # Push the real commit
-  git push origin asf-site || (echo "Failed to push to asf-site. Website not updated." && exit -1)
-  # Create an empty commit to work around INFRA-10751
+fi
+
+# Add all the changes
+echo "Adding all the files we know about"
```
mistyhacks: Nit: redundant with the comment above.

dev-support/jenkins-scripts/generate-hbase-website.sh
```diff
@@ -125,46 +225,62 @@ FILES_TO_REMOVE=("hbase-*"
...
+# Create the commit message and commit the changes
+WEBSITE_COMMIT_MSG="Published site at $CURRENT_HBASE_COMMIT."
+echo "WEBSITE_COMMIT_MSG: $WEBSITE_COMMIT_MSG"
+git commit -m "${WEBSITE_COMMIT_MSG}" -a
+# Dump a little report
+echo "This commit changed these files (excluding Modified files):"
+git diff --name-status --diff-filter=ADCRTXUB origin/asf-site | tee "${working_dir}/hbase-file-diff-summary-${CURRENT_HBASE_COMMIT}.txt"
+# Create a patch, which Jenkins can save as an artifact and can be examined for debugging
+git format-patch --stdout origin/asf-site > "${working_dir}/${CURRENT_HBASE_COMMIT}.patch"
+if [ ! -s "${working_dir}/${CURRENT_HBASE_COMMIT}.patch" ]; then
```
mistyhacks: Is this really the test you want? I think `-s` is testing whether something is a string, but I think you want a file match.

dev-support/jenkins-scripts/generate-hbase-website.sh
busbey: verify it worked how? if the command failed the script will end with an error exit code because we used `set -e` above.

dev-support/jenkins-scripts/generate-hbase-website.sh
busbey: out of date how?

dev-support/jenkins-scripts/generate-hbase-website.sh
busbey: so don't set it at all or try to check the java version or something?

dev-support/jenkins-scripts/generate-hbase-website.sh
busbey: `-s` is "file exists and is not zero size"

dev-support/jenkins-scripts/generate-hbase-website.sh
busbey: Misty says

> we're not really running `mvn clean site`

I'll switch this to a descriptive string.

dev-support/jenkins-scripts/generate-hbase-website.sh
mistyhacks: Got it.

dev-support/jenkins-scripts/generate-hbase-website.sh
mistyhacks: LOL at the IRL code review.

dev-support/jenkins-scripts/generate-hbase-website.sh
mistyhacks: Yeah we determined i was thinking of `-n`.

dev-support/jenkins-scripts/generate-hbase-website.sh
mistyhacks: Apparently not needed anymore: http://openjdk.java.net/jeps/122

busbey: do we only ever build the website with the master branch? If so then I think we're good since that's always built with java 8.

mistyhacks: Correct. We probably should review that because I think it means that the other branches don’t have up to date Javadoc available anywhere for browsing, but that’s out of scope for this.

busbey: FWIW, when I do a branch-1.2 release I just run `mvn javadoc:aggregate` to get the javadocs to copy in place.

PR #74, opened 2018-03-09
HBASE-20162 [nightly] depending on pipeline execution we sometimes re…

busbey:
…fer to the wrong workspace

### DIFF
--- a/dev-support/Jenkinsfile
+++ b/dev-support/Jenkinsfile
@@ -30,9 +30,6 @@ pipeline {
     skipDefaultCheckout()
   }
   environment {
-    TOOLS = "${env.WORKSPACE}/tools"
-    // where we check out to across stages
-    BASEDIR = "${env.WORKSPACE}/component"
     YETUS_RELEASE = '0.7.0'
     // where we'll write everything from different steps. Need a copy here so the final step can check for success/failure.
     OUTPUT_DIR_RELATIVE_GENERAL = 'output-general'
@@ -42,6 +39,7 @@ pipeline {
 
     PROJECT = 'hbase'
     PROJECT_PERSONALITY = 'https://raw.githubusercontent.com/apache/hbase/master/dev-support/hbase-personality.sh'
+    PERSONALITY_FILE = 'tools/personality.sh'
     // This section of the docs tells folks not to use the javadoc tag. older branches have our old version of the check for said tag.
     AUTHOR_IGNORE_LIST = 'src/main/asciidoc/_chapters/developer.adoc,dev-support/test-patch.sh'
     WHITESPACE_IGNORE_LIST = '.*/generated/.*'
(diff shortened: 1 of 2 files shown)


dev-support/Jenkinsfile
```diff
@@ -101,7 +98,8 @@ if [ ! -d "${YETUS_DIR}" ]; then
   gunzip -c yetus.tar.gz | tar xpf - -C "${YETUS_DIR}" --strip-components 1
 fi
         '''
-        dir ("${env.TOOLS}") {
+        // Set up the file we need at PERSONALITY_FILE location
+        dir ("tools") {
```
madrob: This will always be relative to root? To $BASEDIR?

dev-support/Jenkinsfile
```diff
@@ -408,6 +410,9 @@ curl -L  -o personality.sh "${env.PROJECT_PERSONALITY}"
             // this needs to be set to the jdk that ought to be used to build releases on the branch the Jenkinsfile is stored in.
             jdk "JDK 1.8 (latest)"
           }
+          environment {
+            BASEDIR = "${env.WORKSPACE}/component"
```
madrob: What's the difference between ${BASEDIR} and ${env.BASEDIR} [below](https://github.com/apache/hbase/pull/74/files#diff-070c3739c8fa8192e0741b4219d30de1R429)

Do they both read this value, or does only env.BASEDIR read this and the other one comes from ~magic

dev-support/Jenkinsfile
```diff
@@ -130,6 +128,7 @@ curl -L  -o personality.sh "${env.PROJECT_PERSONALITY}"
             }
           }
           environment {
+            BASEDIR = "${env.WORKSPACE}/component"
```
madrob: I... don't understand how these are different from what we had previously?

dev-support/Jenkinsfile
busbey: it's relative to the workspace root.

dev-support/Jenkinsfile
busbey: env.BASEDIR gets replaced with the literal value by jenkins. BASEDIR is a bash variable that will be set to a value of the same thing that Jenkins would rewrite.

dev-support/Jenkinsfile
busbey: the captured context is different. jenkins captures things like WORKSPACE per executor. how it turns the given stages and initial environment into executors is magic AFAICT.

busbey: so down here, because we had a `node` on the stage we get a new executor and a new WORKSPACE gets made.

dev-support/Jenkinsfile
madrob: this one doesn't have a node? so where does it exec?

dev-support/Jenkinsfile
busbey: at the end of this stage we stash the tools directory and in future stages we unstash it relative to those stages workspace root.

dev-support/Jenkinsfile
busbey: stages that don't have their own node label exec on the same node as the task that launches the entire pipeline (so the one that runs the "install yetus" and "init health results" steps)

busbey:
ugh, lack of jira integration / notifications.

PR #78, opened 2018-06-18
Update HBaseContext.scala

CPorrot:
Added a bulkLoadCustomVersions method:
* flatMap requires an Iterator of a Triple[KeyFamilyQualifier, Array[Byte], Long], with the third element being the version
* writeValueToHFile will use this version, or nowTimeStamp if the version is negative

Reason:
Our project uses spark-hbase, but we cannot use bulkLoad for our RDD, because we need custom versions for our cells (1, 2, 3, 4), instead of the timestamp, because we can receive the same version of the object several times through our SparkStreaming job. This additional method would allow the use of a bulkLoad in this specific context.


chia7712:
HBase does not use the PR flow to do the code update and discussion. Please use jira to report and fix this issue. Take a look at https://hbase.apache.org/book.html#developer for more details.

Cheers,
chia-ping

PR #79, opened 2018-06-23
HBASE-20557 Backport HBASE-17215 to branch-1

taklwu:
The second backport of HBASE-20555

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/cleaner/HFileCleaner.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/cleaner/HFileCleaner.java
@@ -17,22 +17,32 @@
  */
 package org.apache.hadoop.hbase.master.cleaner;
 
+import com.google.common.annotations.VisibleForTesting;
+import java.io.IOException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-
-import org.apache.hadoop.hbase.classification.InterfaceAudience;
+import java.util.concurrent.BlockingQueue;
+import java.util.concurrent.LinkedBlockingQueue;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileStatus;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.Stoppable;
+import org.apache.hadoop.hbase.classification.InterfaceAudience;
+import org.apache.hadoop.hbase.conf.ConfigurationObserver;
 import org.apache.hadoop.hbase.io.HFileLink;
 import org.apache.hadoop.hbase.regionserver.StoreFileInfo;
+
 /**
  * This Chore, every time it runs, will clear the HFiles in the hfile archive
  * folder that are deletable for each HFile cleaner in the chain.
  */
 @InterfaceAudience.Private
-public class HFileCleaner extends CleanerChore<BaseHFileCleanerDelegate> {
+public class HFileCleaner extends CleanerChore<BaseHFileCleanerDelegate> implements
+    ConfigurationObserver {
 
   public static final String MASTER_HFILE_CLEANER_PLUGINS = "hbase.master.hfilecleaner.plugins";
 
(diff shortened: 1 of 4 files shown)


taklwu:
ran `mvn clean install -DskipTest` and executed unit test `TestHFileCleaner` successfully, pinged on [HBASE-20557](https://issues.apache.org/jira/browse/HBASE-20557)

z-york:
Can you fix the import order checkstyle warning?

taklwu:
I tried to used the `hbase_eclipse_formatter.xml` and followed the setting in `hbase-checkstyle/src/main/resources/hbase/checkstyle.xml`, submitted a new patch for reorganizing the imports. However, it's not working as expected and I don't know what is missing from the checkstyle, you can also see those non-changed classes `MergeNormalizationPlan` and `SplitNormalizationPlan` is importing correctly as well but failed with the same issue, I'm wondered if I'm hitting a bug or something for checkstyle, I need more time to look into it but if you can point out the problem that would be helpful 

```

  [ERROR] /local/home/wutaklon/workspace-tmp/hbase/hbase-server/src/main/java/org/apache/hadoop/hbase/master/normalizer/MergeNormalizationPlan.java:27: Wrong order for 'java.io.IOException' import. [ImportOrder]
  [ERROR] /local/home/wutaklon/workspace-tmp/hbase/hbase-server/src/main/java/org/apache/hadoop/hbase/master/normalizer/SplitNormalizationPlan.java:27: Wrong order for 'java.io.IOException' import. [ImportOrder]
...
[ERROR] /local/home/wutaklon/workspace-tmp/hbase/hbase-server/src/main/java/org/apache/hadoop/hbase/master/cleaner/HFileCleaner.java:33: Wrong order for 'java.io.IOException' import. [ImportOrder]

```

taklwu:
I figured it out at the end that checkstyle of the IDE in `hbase_eclipse_formatter.xml` and checkstyle while running maven-checkstyle-plugin with conf in `hbase-checkstyle/src/main/resources/hbase/checkstyle.xml` does not match.  

with the following the imports order checkstyle will pass

```

import com.google.common.annotations.VisibleForTesting;
import java.io.IOException;
import java.util.ArrayList;
import java.util.List;
import java.util.Map;
import java.util.concurrent.BlockingQueue;
import java.util.concurrent.LinkedBlockingQueue;
import org.apache.commons.logging.Log;
import org.apache.commons.logging.LogFactory;
import org.apache.hadoop.conf.Configuration;
import org.apache.hadoop.fs.FileStatus;
import org.apache.hadoop.fs.FileSystem;
import org.apache.hadoop.fs.Path;
import org.apache.hadoop.hbase.Stoppable;
import org.apache.hadoop.hbase.classification.InterfaceAudience;
import org.apache.hadoop.hbase.conf.ConfigurationObserver;
import org.apache.hadoop.hbase.io.HFileLink;
import org.apache.hadoop.hbase.regionserver.StoreFileInfo;

```

So, I'm wondered if we should have an separate JIRA (I will open it and fix it) to fix and sync between these two setting files (with [HBASE-19262](https://issues.apache.org/jira/browse/HBASE-19262) and [HBASE-19552](https://issues.apache.org/jira/browse/HBASE-19552)) still does not work well, so we need to backport it branch-1 and also fix the problem we found here.

For this changes, I personally see the current commit matched what  `hbase_eclipse_formatter.xml` setup, so, it's safe to commit. but if you insist we need to fix it within this PR, please let me know if we should fix it here instead of a new JIRA/PR.

z-york:
Let's update the order related to the checkstyle.xml for now and open a JIRA to discuss this mismatch.

taklwu:
sure, I have done that by updated the commit and patch on the HBASE-20557, the latest auto-build show the checkstyle has been passed.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/cleaner/HFileCleaner.java
```diff
@@ -17,22 +17,32 @@
  */
 package org.apache.hadoop.hbase.master.cleaner;
 
+import com.google.common.annotations.VisibleForTesting;
```
z-york: It's weird that this wants it all in one block... but if checkstyle likes it who am I to disagree? :)

taklwu:
any update ? do I need to change anything else?

z-york:
merged. @busbey how can I close the PR? It looks like I don't have the permissions required.

taklwu:
this has been merged as https://github.com/apache/hbase/commit/a52c6bf2f3f742ae4bec29c5d10dfc339c1e3f0d

PR #83, opened 2018-07-19
HBASE-20856 PITA having to set WAL provider in two places

taklwu:
With this change if hbase.wal.meta_provider is not explicitly set,
it uses whatever set with hbase.wal.provider. this change avoids a use
case of unexpectedly using two different providers when only
hbase.wal.provider is set to non-default but not hbase.wal.meta_provider.

This change also include document (architecture.adoc) update


taklwu:
merged as https://github.com/apache/hbase/commit/1d0fca370bf56a41fc62b72bebe86a7185a2b0c2

PR #85, opened 2018-07-23
 Fixing a well known binary search bug, -ve bounds

bashrc-real:
There are couple of issues in the code being fixed:
*  `>>>` operator would mess the values if `low` + `high` end up being negative.  This shouldn't happen but I don't see anything to prevent this from happening. 

* The code fails around boundary values of `low` and `high`. This is a well known binary search catch. https://ai.googleblog.com/2006/06/extra-extra-read-all-about-it-nearly.html


busbey:
Hi!

Thanks for taking the time to put together this PR. The HBase project uses the [HBASE tracker in ASF JIRA instance](https://issues.apache.org/jira/browse/HBASE) to review and apply work. Would you mind filing a request there that describes the problem and proposed solution?

Our [contributor guide section includes instructions on how to attach a patch to your issue once it exists](http://hbase.apache.org/book.html#submitting.patches).

Thanks for your interest in improving Apache HBase!

bashrc-real:
This has been fixed with commit https://github.com/apache/hbase/commit/3a3855aade174d89087c5959320b7662347c1275

PR #86, opened 2018-07-26
HBASE-20734 Colocate recovered edits directory with hbase.wal.dir

z-york:


hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
```diff
@@ -330,6 +330,9 @@
   private final int rowLockWaitDuration;
   static final int DEFAULT_ROWLOCK_WAIT_DURATION = 30000;
 
+  private Path regionDir;
+  private FileSystem walFS;
```
tedyu: Add a comment for the role of this field

PR #87, opened 2018-07-29
node-b should be node-c

liganglee1122:
node-b should be node-c, please check this.


liganglee1122:
please check other version also!

PR #107, opened 2019-01-23
HBASE-21760 Unify the upper/lower case for title

xubo245:


Apache9:
Hi, the workflow for HBase has not been migrated to Github based yet, so please upload your patch in HBASE-21760 directly. I've already assign the issue to you.

PR #108, opened 2019-03-14
HBASE-22042 add missing @Override annotation for RawAsyncTableImpl.scan

jainrish:


willshen:
@jainrish I dont think we use PR to accept changes for HBase yet. Please see http://hbase.apache.org/book.html#submitting.patches

PR #261, opened 2019-05-28
HBASE-22467 UI fixes to enable Knox proxying

joshelser:


hbase-http/src/main/java/org/apache/hadoop/hbase/http/ProfileOutputServlet.java
```diff
@@ -48,11 +51,25 @@ protected void doGet(final HttpServletRequest req, final HttpServletResponse res
...
+      resp.setHeader("Refresh", REFRESH_PERIOD + ";" + refreshUrl);
       resp.getWriter().write("This page will be auto-refreshed every " + REFRESH_PERIOD +
-        " seconds until the output file is ready.");
+        " seconds until the output file is ready. Redirecting to " + refreshUrl);
     } else {
       super.doGet(req, resp);
     }
   }
+
+  static String sanitize(String input) {
```
busbey: Any particular rationale for this approach rather than using the existing `org.apache.hadoop.hbase.http.HtmlQuoting.quoteHtmlChars(String)` method?

I guess the quoting approach doesn't protect against an attacker setting the `Accept` header to javascript at the same time they include a query parameter that would cause the browser to execute said javascript in the echoed page.

I'm surprised there isn't already a utility method for this.

hbase-http/src/main/java/org/apache/hadoop/hbase/http/ProfileOutputServlet.java
```diff
@@ -37,6 +38,8 @@
   private static final long serialVersionUID = 1L;
   private static final Logger LOG = LoggerFactory.getLogger(ProfileOutputServlet.class);
   private static final int REFRESH_PERIOD = 2;
+  // Alphanumeric characters, plus percent (url-encoding), equals, and ampersand
+  private static final Pattern ALPHA_NUMERIC = Pattern.compile("[a-zA-Z0-9\\%\\=\\&]*");
```
busbey: nit:should also accept a literal `+` for spaces I think?

hbase-http/src/main/java/org/apache/hadoop/hbase/http/ProfileOutputServlet.java
joshelser: Nope! Just ignorant of that class.

hbase-http/src/main/java/org/apache/hadoop/hbase/http/ProfileOutputServlet.java
joshelser: Agree!

joshelser:
Pushed edaae03 which uses HtmlQuoting. Don't see a need to re-invent the wheel and I think the implementation of that method is fine.

hbase-http/src/test/java/org/apache/hadoop/hbase/http/TestProfileOutputServlet.java
```diff
@@ -0,0 +1,55 @@
...
+import org.junit.experimental.categories.Category;
+
+@Category({SmallTests.class})
+public class TestProfileOutputServlet {
+  @ClassRule
+  public static final HBaseClassTestRule CLASS_RULE =
+      HBaseClassTestRule.forClass(TestProfileOutputServlet.class);
+
+  @Test
+  public void testSanitization() {
```
busbey: won't these tests fail now?

joshelser: Ugh, yes.

joshelser:
Ripping out edaae03 -- doesn't work for what I'm trying to do here. Will leave a comment on commit.

PR #366, opened 2019-07-09
HBASE-22658 region_mover.rb should choose same rsgroup servers as target servers

a516072575:
There are many retries when i am using graceful_stop.sh to shutdown region server after using regroup, because the target server in a different rsgroup. This makes it slow to graceful shutdown a regionserver. So I think that region_mover.rb  should only choose same rsgroup servers as target servers.


a516072575:
I chose the wrong branch.

PR #377, opened 2019-07-12
HBASE-22686 ZkSplitLogWorkerCoordination doesn't allow a regionserver to pick up all of the split work it is capable of

apurtell:
The regionserver side distributed split coordination logic only allows a regionserver to grab one task one at a time. Nearby javadoc says "This policy puts an upper-limit on the number of simultaneous log splitting that could be happening in a cluster." That upper limit will be the number of currently live regionservers. However, a regionserver can handle more than one task at once and in fact the max number of concurrent split tasks it can accept is configurable.


apurtell:
Updated to remove now invalid unit test.

apurtell:
Updated to remove now invalid unit test.

apurtell:
Latest change moves debug log lines to TRACE level which might otherwise be logged 10 times a second.

PR #397, opened 2019-07-21
HBASE-22676 Move all the code in hbase-rsgroup to hbase-server and re…

Apache9:
…move hbase-rsgroup module


Apache9:
The pre commit failed and I think this is a bug for yetus. We just removed the hbase-rsgroup module but in the javadoc check we still wants to test the module.

> 22:41:57  /testptch/hbase/hbase-rsgroup no longer exists. Skipping.
22:41:57  /testptch/patchprocess/precommit/test-patch.sh: line 1690: pushd: /testptch/hbase/hbase-rsgroup: No such file or directory
22:41:57  
22:41:57  
22:41:57  
22:41:57  ASSERT: no offset passed to offset_clock: unit

PR #640, opened 2019-09-19
HBASE-23008 put batch with updated position into queue even empty

bsglz:
In order to trigger cleanOldLogs


bsglz:
@saintstack Could you help to review it ?

HorizonNet:
@bsglz Is this a branch-2-only problem? If not, can you please target master and we then start backporting it.

saintstack:
But like @HorizonNet says, Master branch is different. Does it have this issue?

bsglz:
Yeah, branch-2 only, master is ok.
Close it.

PR #699, opened 2019-10-08
HBASE-23136 PartionedMobFileCompactor bulkloaded files shouldn't get …

wchevreuil:
…replicated (addressing buklload replication related issue raised in HBASE-22380)


petersomogyi:
Can we use a shorter commit message? :)

wchevreuil:
> Should this new option be exposed on command-line?

Definitely! Will address that on a separate issue, since on this one I'm most concerned with avoiding getting compacted mob files replicated.

> Can we use a shorter commit message? :)

We sure can, I will edit the message prior to merging it.

PR #733, opened 2019-10-18
HBASE-23191 EOFE log spam

karthikhw:
If no new active writes in WAL edit, then WALEntryStream#hasNext -> ReaderBase -> ProtobufLogReader#readNext will reach end of file. It would be a good idea for changing the log level from INFO to DEBUG.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/ProtobufLogReader.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/ProtobufLogReader.java
@@ -418,7 +418,7 @@ public class ProtobufLogReader extends ReaderBase {
               "current position and original position match at {}", originalPosition);
             seekOnFs(0);
           } else {
-            LOG.info("Reached the end of file at position {}", originalPosition);
+            LOG.debug("Reached the end of file at position {}", originalPosition);
           }
         } else {
           // Else restore our position to original location in hope that next time through we will


wchevreuil:
As someone who had been troubleshooting _Replication stuck_ issues for the last 5+ years, I really think it's useful having that information right away, without needing to reset

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/ProtobufLogReader.java
```diff
@@ -418,7 +418,7 @@ protected boolean readNext(Entry entry) throws IOException {
               "current position and original position match at {}", originalPosition);
             seekOnFs(0);
           } else {
-            LOG.info("Reached the end of file at position {}", originalPosition);
+            LOG.debug("Reached the end of file at position {}", originalPosition);
```
wchevreuil: How much overhead would you think this creates on the logs? As someone who had been troubleshooting _replication stuck_ issues for the last 5+ years, I really think it's useful having that information right away, without needing to reset log level.

karthikhw: Thank you very much @wchevreuil for checking this.  There is no heavy overhead and it generates only 800 KB/hour. I filed this JIRA because likely possible that we miss some log due to log rollthe ing when customer maintenance goes longer(I.e few days).  And, this info is added recently HBASE-22502 (hbase-2.2.0). IMO, I don't think it is useful log.

karthikhw: Cc: @Apache9

karthikhw: @wchevreuil  @Apache9 (Duo Zhang) Appreciate if you would confirm the change.

saintstack:
Merged this. The log was added in 2.2.0.

PR #780, opened 2019-10-31
HBASE-23238 Additional test and checks for null references on Scanner…

wchevreuil:
…CallableWithReplicas


wchevreuil:
Thanks @busbey ! 
> would the test also get forward ported to branches-2/master? or is there something providing equivalent coverage already?

I was planning to add the test in all branches, as I don't see any existing one checking on this particular condition. The null checks would also apply to branch-2, but not to master, so for master we should probably have a patch with the test only, but yeah, let's wait for qabot.

PR #781, opened 2019-11-01
HBASE-18095: Zookeeper-less client connection implementation

bharathv:
Generally when a HBaseClient tries to create a cluster
Connection, it fetches a bunch of metadata from Zookeeper
(like active master address, clusterID, meta locations etc)
before it creates the underlying transport. However exposing
ZK to all the clients is a DDOS risk and ZK connections in
the past have caused issues by not timing out on blocking
RPCs (more context in the JIRA).

This patch attempts to remove this ZK dependency by making
the client fetch all the meta information directly from list
of client configured HMaster endpoints. The patch adds a
a new AsyncRegistry implementation that encapsulates this logic
of fetching this meta information from the provided master
end points. New RPCs are added to the HMasters to help fetch
this information.

Meta HRL caching:
----------------
One critical piece of metadata needed by clients to query tables
is meta HRegionLocations. These are fetched from ZK by default.
Since this patch moves away from ZK, it adds an in-memory cache
of these locations on both Active/StandBy HMasters. ZK Listeners
are registered to keep the cache up-to-date.

New client configs:
------------------
- 'hbase.client.asyncregistry.masteraddrs' Should be set to a
list of comma separated HMaster host:port addresses.
- Should be used in conjunction with 'hbase.client.registry.impl'
set to HMasterAsyncRegistry class.

* Testing is still a WIP. Error paths are missing (ex: a master is
not accessible etc).

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/MetaRegionsNotAvailableException.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/MetaRegionsNotAvailableException.java
@@ -1,4 +1,4 @@
-/**
+/*
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
  * distributed with this work for additional information
@@ -15,29 +15,21 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package org.apache.hadoop.hbase.client;
+package org.apache.hadoop.hbase;
 
-import org.apache.hadoop.conf.Configuration;
 import org.apache.yetus.audience.InterfaceAudience;
-import org.apache.hadoop.hbase.util.ReflectionUtils;
+
+import java.io.IOException;
 
 /**
- * Get instance of configured Registry.
+ * Thrown by master when meta region locations are not cached whatever reason.
+ * Client is expected to retry when running into this.
  */
 @InterfaceAudience.Private
-final class AsyncRegistryFactory {
+public class MetaRegionsNotAvailableException extends IOException  {
+    private static final long serialVersionUID = (1L << 14) - 1L;
 
-  public static final String REGISTRY_IMPL_CONF_KEY = "hbase.client.registry.impl";
-
-  private AsyncRegistryFactory() {
-  }
-
-  /**
-   * @return The cluster registry implementation to use.
-   */
-  static AsyncRegistry getRegistry(Configuration conf) {
-    Class<? extends AsyncRegistry> clazz =
-        conf.getClass(REGISTRY_IMPL_CONF_KEY, ZKAsyncRegistry.class, AsyncRegistry.class);
-    return ReflectionUtils.newInstance(clazz, conf);
-  }
+    public MetaRegionsNotAvailableException(String msg) {
+        super(msg);
+    }
 }
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
@@ -394,4 +394,7 @@ class AsyncConnectionImpl implements AsyncConnection {
   Optional<MetricsConnection> getConnectionMetrics() {
     return metrics;
   }
+
+  @VisibleForTesting
+  AsyncRegistry getRegistry() { return registry; }
 }
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
@@ -1,0 +1,236 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.client;
+
+import static org.apache.hadoop.hbase.HConstants.DEFAULT_HBASE_RPC_TIMEOUT;
+import static org.apache.hadoop.hbase.HConstants.HBASE_RPC_TIMEOUT_KEY;
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.HRegionLocation;
+import org.apache.hadoop.hbase.MasterNotRunningException;
+import org.apache.hadoop.hbase.MetaRegionsNotAvailableException;
+import org.apache.hadoop.hbase.RegionLocations;
+import org.apache.hadoop.hbase.ServerName;
+import org.apache.hadoop.hbase.ipc.HBaseRpcController;
+import org.apache.hadoop.hbase.ipc.RpcClient;
+import org.apache.hadoop.hbase.ipc.RpcClientFactory;
+import org.apache.hadoop.hbase.ipc.RpcControllerFactory;
+import org.apache.hadoop.hbase.security.User;
+import org.apache.hadoop.hbase.shaded.protobuf.ProtobufUtil;
+
+import org.apache.hadoop.hbase.shaded.protobuf.generated.HBaseProtos;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetClusterIdRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetClusterIdResponse;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetMetaLocationsRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetMetaLocationsResponse;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsActiveRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsActiveResponse;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.MasterService;
+import org.apache.hbase.thirdparty.com.google.common.base.Preconditions;
+import org.apache.hbase.thirdparty.com.google.protobuf.ServiceException;
+
+import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Fetches the meta information directly from HMaster by making relevant RPCs. HMaster RPC end
+ * points are looked up via client configuration 'hbase.client.asyncregistry.masteraddrs' as
+ * comma separated list of host:port values. Should be set in conjunction with
+ * 'hbase.client.registry.impl' set to this class impl.
+ *
+ * The class does not cache anything. It is the responsibility of the callers to cache and
+ * avoid repeated requests.
+ */
+@InterfaceAudience.Private
+public class HMasterAsyncRegistry implements AsyncRegistry {
+  private static final Logger LOG = LoggerFactory.getLogger(HMasterAsyncRegistry.class);
+  public static final String CONF_KEY = "hbase.client.asyncregistry.masteraddrs";
+  private static final String DEFAULT_HOST_PORT = "localhost:" + HConstants.DEFAULT_MASTER_PORT;
+
+  // Parsed list of host and ports for masters from hbase-site.xml
+  private final List<ServerName> masterServers;
+  final Configuration conf;
+  // RPC client used to talk to the master servers. This uses a stand alone RpcClient instance
+  // because AsyncRegistry is created prior to creating a cluster Connection. The client is torn
+  // down in close().
+  final RpcClient rpcClient;
+  final int rpcTimeout;
+
+  public HMasterAsyncRegistry(Configuration config) {
+    masterServers = new ArrayList<>();
+    conf = config;
+    parseHortPorts();
+    // Passing the default cluster ID means that the token based authentication does not work for
+    // certain client implementations.
+    // TODO(bharathv): Figure out a way to fetch the CLUSTER ID using a non authenticated way.
+    rpcClient = RpcClientFactory.createClient(conf, HConstants.CLUSTER_ID_DEFAULT);
+    rpcTimeout = (int) Math.min(Integer.MAX_VALUE,
+        TimeUnit.MILLISECONDS.toNanos(conf.getLong(HBASE_RPC_TIMEOUT_KEY,
+            DEFAULT_HBASE_RPC_TIMEOUT)));
+  }
+
+  private void parseHortPorts() {
+    String hostPorts = conf.get(CONF_KEY, DEFAULT_HOST_PORT);
+    for (String hostPort: hostPorts.split(",")) {
+      masterServers.add(ServerName.valueOf(hostPort, ServerName.NON_STARTCODE));
+    }
+    Preconditions.checkArgument(!masterServers.isEmpty(), String.format("%s is empty", CONF_KEY));
+    // Randomize so that not every client sends requests in the same order.
+    Collections.shuffle(masterServers);
+  }
+
+  /**
+   * Util that generates a master stub for a given ServerName.
+   */
+  private MasterService.BlockingInterface getMasterStub(ServerName server) throws IOException  {
+    return MasterService.newBlockingStub(
+        rpcClient.createBlockingRpcChannel(server, User.getCurrent(), rpcTimeout));
+  }
+
+  /**
+   * Blocking RPC to fetch the meta region locations using one of the masters from the parsed list.
+   */
+  private RegionLocations getMetaRegionLocationsHelper() throws MetaRegionsNotAvailableException {
+    List<HBaseProtos.RegionLocation> result = null;
+    for (ServerName sname: masterServers) {
+      try {
+        MasterService.BlockingInterface stub = getMasterStub(sname);
+        HBaseRpcController rpcController = RpcControllerFactory.instantiate(conf).newController();
+        GetMetaLocationsResponse resp = stub.getMetaLocations(rpcController,
+            GetMetaLocationsRequest.getDefaultInstance());
+        result = resp.getLocationsList();
+      } catch (Exception e) {
+        LOG.warn("Error fetch meta locations from master {}. Trying others.", sname, e);
+      }
+    }
+    if (result == null || result.isEmpty()) {
+      throw new MetaRegionsNotAvailableException(String.format(
+          "Meta locations not found. Probed masters: %s", conf.get(CONF_KEY, DEFAULT_HOST_PORT)));
+    }
+    List<HRegionLocation> deserializedResult = new ArrayList<>();
+    result.stream().forEach(
+        location -> deserializedResult.add(ProtobufUtil.toRegionLocation(location)));
+    return new RegionLocations(deserializedResult);
+  }
+
+  /**
+   * Picks the first master entry from 'masterHortPorts' to fetch the meta region locations.
+   */
+  @Override
+  public CompletableFuture<RegionLocations> getMetaRegionLocation() {
+    CompletableFuture<RegionLocations> result = new CompletableFuture<>();
+    CompletableFuture.runAsync(() -> {
+      try {
+        result.complete(this.getMetaRegionLocationsHelper());
+      } catch (Exception e) {
+        result.completeExceptionally(e);
+      }
+    });
+    return result;
+  }
+
+  /**
+   * Blocking RPC to get the cluster ID from the parsed master list. Returns null if no active
+   * master found.
+   */
+  private String getClusterIdHelper() throws  MasterNotRunningException {
+    // Loop through all the masters serially. We could be hitting some standby masters which cannot
+    // process this RPC, so we just skip them.
+    for (ServerName sname: masterServers) {
+      try {
+        MasterService.BlockingInterface stub = getMasterStub(sname);
+        HBaseRpcController rpcController = RpcControllerFactory.instantiate(conf).newController();
+        GetClusterIdResponse resp =
+            stub.getClusterId(rpcController, GetClusterIdRequest.getDefaultInstance());
+        return resp.getClusterId();
+      } catch (IOException e) {
+        LOG.warn("Error fetching cluster ID from master: {}", sname, e);
+      } catch (ServiceException e) {
+        // This is probably a standby master, can be ignored.
+        LOG.debug("Error fetching cluster ID from server: {}" , sname, e);
+      }
+    }
+    // If it comes to this point, no active master could be found.
+    throw new MasterNotRunningException(String.format(
+        "No active master found. Probed masters: %s", conf.get(CONF_KEY, DEFAULT_HOST_PORT)));
+  }
+
+  @Override
+  public CompletableFuture<String> getClusterId() {
+    CompletableFuture<String> result = new CompletableFuture<>();
+    CompletableFuture.runAsync(() -> {
+      try {
+        result.complete(this.getClusterIdHelper());
+      } catch (Exception e) {
+        result.completeExceptionally(e);
+      }
+    });
+    return result;
+  }
+
+  /**
+   * Blocking RPC to get the active master address from the parsed list of master servers.
+   */
+  private ServerName getMasterAddressHelper() throws MasterNotRunningException  {
+    for (ServerName sname: masterServers) {
+      try {
+        MasterService.BlockingInterface stub = getMasterStub(sname);
+        HBaseRpcController rpcController = RpcControllerFactory.instantiate(conf).newController();
+        IsActiveResponse resp = stub.isActive(rpcController, IsActiveRequest.getDefaultInstance());
+        if (resp.getIsMasterActive()) {
+          return ServerName.valueOf(sname.getHostname(), sname.getPort(), resp.getStartCode());
+        }
+      } catch (Exception e) {
+
+      }
+    }
+    throw new MasterNotRunningException(String.format("No active master found. Probed masters: %s",
+        conf.get(CONF_KEY, DEFAULT_HOST_PORT)));
+  }
+
+  /**
+   * @return the active master among the configured master addresses in 'masterHortPorts'.
+   */
+  @Override
+  public CompletableFuture<ServerName> getMasterAddress() {
+    CompletableFuture<ServerName> result = new CompletableFuture<>();
+    CompletableFuture.runAsync(() -> {
+      try {
+        result.complete(this.getMasterAddressHelper());
+      } catch (Exception e) {
+        result.completeExceptionally(e);
+      }
+    });
+    return result;
+  }
+
+  @Override
+  public void close() {
+    if (rpcClient != null) {
+      rpcClient.close();
+    }
+  }
+}
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/shaded/protobuf/ProtobufUtil.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/shaded/protobuf/ProtobufUtil.java
@@ -3188,6 +3188,7 @@ public final class ProtobufUtil {
   }
 
   public static HBaseProtos.RegionLocation toRegionLocation(HRegionLocation loc) {
+    if (loc == null) return null;
     HBaseProtos.RegionLocation.Builder builder = HBaseProtos.RegionLocation.newBuilder();
     builder.setRegionInfo(toRegionInfo(loc.getRegion()));
     if (loc.getServerName() != null) {
--- a/hbase-protocol-shaded/src/main/protobuf/Master.proto
+++ b/hbase-protocol-shaded/src/main/protobuf/Master.proto
@@ -585,6 +585,36 @@ message GetProceduresResponse {
   repeated Procedure procedure = 1;
 }
 
+message IsActiveRequest {
+}
+
+message IsActiveResponse {
+  required bool is_master_active = 1;
+  required uint64 start_code = 2;
+}
+
+message GetMetaLocationsRequest {
+}
+
+message GetMetaLocationsResponse {
+  repeated RegionLocation locations = 1;
+}
+
+message GetPortInfoRequest {
+}
+
+message GetPortInfoResponse {
+  required uint32 info_port = 1;
+  required uint32 master_port = 2;
+}
+
+message GetClusterIdRequest {
+}
+
+message GetClusterIdResponse {
+    required string cluster_id = 1;
+}
+
 message GetLocksRequest {
 }
 
@@ -706,6 +736,19 @@ service MasterService {
   /** Return cluster status. */
   rpc GetClusterStatus(GetClusterStatusRequest)
     returns(GetClusterStatusResponse);
+
+  /** Returns whether this master is active or not. Served on both active/standby masters.*/
+  rpc IsActive(IsActiveRequest) returns(IsActiveResponse);
+
+  /** Used by client to get the location of meta replica(s). Served on both active/standby masters.*/
+  rpc GetMetaLocations(GetMetaLocationsRequest)
+    returns(GetMetaLocationsResponse);
+
+  /** Returns the ClusterId UUID for this cluster */
+  rpc GetClusterId(GetClusterIdRequest) returns(GetClusterIdResponse);
+
+  /** Returns the port numbers used by HMaster instances. Served on both active/standby masters.*/
+  rpc GetPortInfo(GetPortInfoRequest) returns(GetPortInfoResponse);
 
   /** return true if master is available */
   rpc IsMasterRunning(IsMasterRunningRequest) returns(IsMasterRunningResponse);
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterRpcServices.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterRpcServices.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.ClusterMetricsBuilder;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.MetaRegionsNotAvailableException;
 import org.apache.hadoop.hbase.MetaTableAccessor;
 import org.apache.hadoop.hbase.NamespaceDescriptor;
 import org.apache.hadoop.hbase.Server;
@@ -184,14 +185,20 @@ import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.ExecProced
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.ExecProcedureResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.FixMetaRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.FixMetaResponse;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetClusterIdRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetClusterIdResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetClusterStatusRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetClusterStatusResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetCompletedSnapshotsRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetCompletedSnapshotsResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetLocksRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetMetaLocationsRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetMetaLocationsResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetLocksResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetNamespaceDescriptorRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetNamespaceDescriptorResponse;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetPortInfoRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetPortInfoResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetProcedureResultRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetProcedureResultResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetProceduresRequest;
@@ -205,6 +212,8 @@ import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetTableNa
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetTableStateRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.GetTableStateResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.HbckService;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsActiveRequest;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsActiveResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsBalancerEnabledRequest;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsBalancerEnabledResponse;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.IsCatalogJanitorEnabledRequest;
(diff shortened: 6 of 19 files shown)


bharathv:
@apurtell / @wchevreuil FYI. Appreciate if you can give any initial feedback.

Apache-HBase:
:broken_heart: **-1 overall**






| :blue_heart: |  reexec  |   0m 44s |  Docker mode activated.  |
||| _ Prechecks _ |
| :green_heart: |  dupname  |   0m  1s |  No case conflicting files found.  |
| :blue_heart: |  prototool  |   0m  0s |  prototool was not available.  |
| :green_heart: |  hbaseanti  |   0m  0s |  Patch does not have any anti-patterns.  |
| :green_heart: |  @author  |   0m  0s |  The patch does not contain any @author tags.  |
| :green_heart: |  test4tests  |   0m  0s |  The patch appears to include 6 new or modified test files.  |
| :blue_heart: |  mvndep  |   0m 39s |  Maven dependency ordering for branch  |
| :green_heart: |  mvninstall  |   6m 40s |  master passed  |
| :green_heart: |  compile  |   2m 10s |  master passed  |
| :green_heart: |  checkstyle  |   2m 18s |  master passed  |
| :green_heart: |  shadedjars  |   4m 59s |  branch has no errors when building our shaded downstream artifacts.  |
| :green_heart: |  javadoc  |   1m 17s |  master passed  |
| :blue_heart: |  spotbugs  |   4m 57s |  Used deprecated FindBugs config; considering switching to SpotBugs.  |
| :green_heart: |  findbugs  |   8m 46s |  master passed  |
||| _ Patch Compile Tests _ |
| :blue_heart: |  mvndep  |   0m 16s |  Maven dependency ordering for patch  |
| :green_heart: |  mvninstall  |   5m 20s |  the patch passed  |
| :green_heart: |  compile  |   2m 10s |  the patch passed  |
| :green_heart: |  cc  |   2m 10s |  the patch passed  |
| :green_heart: |  javac  |   2m 10s |  the patch passed  |
| :broken_heart: |  checkstyle  |   0m 41s |  hbase-client: The patch generated 145 new + 206 unchanged - 0 fixed = 351 total (was 206)  |
| :broken_heart: |  checkstyle  |   1m 35s |  hbase-server: The patch generated 320 new + 322 unchanged - 0 fixed = 642 total (was 322)  |
| :green_heart: |  whitespace  |   0m  0s |  The patch has no whitespace issues.  |
| :green_heart: |  shadedjars  |   4m 42s |  patch has no errors when building our shaded downstream artifacts.  |
| :green_heart: |  hadoopcheck  |  15m 56s |  Patch does not cause any errors with Hadoop 2.8.5 2.9.2 or 3.1.2.  |
| :green_heart: |  hbaseprotoc  |   2m  3s |  the patch passed  |
| :green_heart: |  javadoc  |   1m 15s |  the patch passed  |
| :broken_heart: |  findbugs  |   4m 24s |  hbase-server generated 2 new + 0 unchanged - 0 fixed = 2 total (was 0)  |
||| _ Other Tests _ |
| :green_heart: |  unit  |   0m 44s |  hbase-protocol-shaded in the patch passed.  |
| :green_heart: |  unit  |   1m 53s |  hbase-client in the patch passed.  |
| :broken_heart: |  unit  | 162m 37s |  hbase-server in the patch failed.  |
| :green_heart: |  asflicense  |   1m 25s |  The patch does not generate ASF License warnings.  |
|  |   | 244m 19s |   |


| Reason | Tests |
|-------:|:------|
| FindBugs | module:hbase-server |
|  |  Dead store to isActiveMaster in org.apache.hadoop.hbase.master.MasterRpcServices.isActive(RpcController, MasterProtos$IsActiveRequest)  At MasterRpcServices.java:org.apache.hadoop.hbase.master.MasterRpcServices.isActive(RpcController, MasterProtos$IsActiveRequest)  At MasterRpcServices.java:[line 1006] |
|  |  Return value of org.apache.hadoop.hbase.zookeeper.ZKWatcher.getZNodePaths() ignored, but method has no side effect  At MetaRegionLocationCache.java:but method has no side effect  At MetaRegionLocationCache.java:[line 83] |
| Failed junit tests | hadoop.hbase.client.TestHMasterAsyncRegistry |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | Client=19.03.4 Server=19.03.4 base: https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/artifact/out/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/781 |
| JIRA Issue | HBASE-18095 |
| Optional Tests | dupname asflicense javac javadoc unit spotbugs findbugs shadedjars hadoopcheck hbaseanti checkstyle compile cc hbaseprotoc prototool |
| uname | Linux 43004a648844 4.15.0-58-generic #64-Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64 GNU/Linux |
| Build tool | maven |
| Personality | /home/jenkins/jenkins-slave/workspace/HBase-PreCommit-GitHub-PR_PR-781/out/precommit/personality/provided.sh |
| git revision | master / ea5c572963 |
| Default Java | 1.8.0_181 |
| checkstyle | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/artifact/out/diff-checkstyle-hbase-client.txt |
| checkstyle | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/artifact/out/diff-checkstyle-hbase-server.txt |
| findbugs | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/artifact/out/new-findbugs-hbase-server.html |
| unit | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/artifact/out/patch-unit-hbase-server.txt |
|  Test Results | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/testReport/ |
| Max. process+thread count | 4919 (vs. ulimit of 10000) |
| modules | C: hbase-protocol-shaded hbase-client hbase-server U: . |
| Console output | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/1/console |
| versions | git=2.11.0 maven=2018-06-17T18:33:14Z) findbugs=3.1.11 |
| Powered by | Apache Yetus 0.11.0 https://yetus.apache.org |


hbase-client/src/main/java/org/apache/hadoop/hbase/shaded/protobuf/ProtobufUtil.java
```diff
@@ -3188,6 +3188,7 @@ public static String toLockJson(List<LockServiceProtos.LockedResource> lockedRes
   }
 
   public static HBaseProtos.RegionLocation toRegionLocation(HRegionLocation loc) {
+    if (loc == null) return null;
```
wchevreuil: just curious, did you face NPE conditions while testing this?

hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
```diff
@@ -213,6 +213,10 @@
 
   private volatile AsyncClusterConnection asyncConnection;
 
+  // Tracks any other connections created with custom  client config. Used for testing clients with custom
+  // configurations. Tracked here so that they can be cleaned up on close() / restart.
+  private List<AsyncClusterConnection> customConnections = Collections.synchronizedList(new ArrayList<>());
+
```
wchevreuil: I didn't quite follow why you needed to define these extra connections. Are you using both types of registry implementation on a single util instance?

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestHMasterAsyncRegistry.java
```diff
@@ -0,0 +1,120 @@
...
+    @BeforeClass
+    public static void setUp() throws Exception {
+        TEST_UTIL.getConfiguration().setInt(META_REPLICAS_NUM, 3);
+        TEST_UTIL.startMiniCluster(3);
+        Configuration conf = new Configuration(TEST_UTIL.getConfiguration());
+        // Set the master host:port in the client config used to create a custom connection.
+        String masterHostName = TEST_UTIL.getMiniHBaseCluster().getMaster().getServerName().getHostname();
+        int masterPort = TEST_UTIL.getMiniHBaseCluster().getMaster().getServerName().getPort();
+        conf.set(HMasterAsyncRegistry.CONF_KEY, HostAndPort.fromParts(masterHostName, masterPort).toString());
+        conf.set(AsyncRegistryFactory.REGISTRY_IMPL_CONF_KEY, HMasterAsyncRegistry.class.getName());
```
wchevreuil: Could be set on _TEST_UTIL_ config before starting the cluster, so that its async connection impl already uses _HMasterAsyncRegistry_?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestHMasterAsyncRegistryRPCs.java
```diff
@@ -0,0 +1,201 @@
...
+        for (int i = 1; i < 10; i++) {
+            assertEquals(i, zk.getZNodePaths().getMetaReplicaIdFromPath(metaZnode + "-" +i));
+        }
+        for (String suffix : new String[]{"foo", "1234", "foo-123", "123-foo-234"}) {
+            try {
+                final String input = metaZnode + suffix;
+                zk.getZNodePaths().getMetaReplicaIdFromZnode(input);
+                fail("Exception not hit getMetaReplicaIdFromZnode(): " +  input);
+            } catch (NumberFormatException e) {
+                // Expected
```
virajjasani: nit: may be we can use assertEquals with NFE's error message if available?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncRegistryFactory.java
```diff
@@ -27,7 +27,7 @@
 @InterfaceAudience.Private
 final class AsyncRegistryFactory {
 
-  static final String REGISTRY_IMPL_CONF_KEY = "hbase.client.registry.impl";
+  public static final String REGISTRY_IMPL_CONF_KEY = "hbase.client.registry.impl";
```
virajjasani: IMHO unless some source code requires this to be public, we can keep this unchanged. For test case, we can directly use "hbase.client.registry.impl". Should be fine?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,230 @@
...
+ * Fetches the meta information directly from HMaster by making relevant RPCs. HMaster RPC end points are looked up
+ * via client configuration 'hbase.client.asyncregistry.masteraddrs' as comma separated list of <host>:<port>
+ * values. Should be set in conjunction with 'hbase.client.registry.impl' set to this class impl.
+ *
+ * The class does not cache anything. It is the responsibility of the callers to cache and avoid repeated requests.
+ */
+@InterfaceAudience.Private
+public class HMasterAsyncRegistry implements AsyncRegistry {
+    private static final Logger LOG = LoggerFactory.getLogger(ZKAsyncRegistry.class);
+    public static final String CONF_KEY = "hbase.client.asyncregistry.masteraddrs";
```
virajjasani: Same here, good to keep it private unless source code needs it?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,230 @@
...
+    public static final String DEFAULT_HOST_PORT = "localhost:" + HConstants.DEFAULT_MASTER_PORT;
```
virajjasani: This can also be private

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetaRegionLocationCache.java
```diff
@@ -0,0 +1,167 @@
...
+ * a listener on ZK to track changes to the meta table znodes. Clients are expected to retry if the meta
+ * information is stale. This class is thread-safe.
+ */
+@InterfaceAudience.Private
+public class MetaRegionLocationCache extends ZKListener {
+
+    private static final Logger LOG = LoggerFactory.getLogger(MetaRegionLocationCache.class);
+
+    // Maximum number of times we retry when ZK operation times out. Should this be configurable?
+    private static final int MAX_ZK_META_FETCH_RETRIES = 10;
```
virajjasani: I think better to make it configurable and may be keep 10(not sure about right value) as default in the absence of site config.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestHMasterAsyncRegistryRPCs.java
```diff
@@ -0,0 +1,201 @@
...
+    @Test
+    public void TestClusterIdRPC() throws Exception {
+        assertEquals(activeMaster.getClusterId(), getClusterId(activeMaster));
+        try {
+            assertTrue(standByMasters.size() > 0);
+            getClusterId(standByMasters.get(0));
+            fail("No exception thrown while fetching ClusterId for standby master: "
+                + standByMasters.get(0).getServerName().toString());
+        } catch (ServiceException e) {
+            // Expected.
```
virajjasani: nit: assertEquals with error message if available?

bharathv:
I configured the checkstyle plugin and fixed most of the checkstyle issues and messy indents. Still figuring out the import order mess. For some reason it is never happy, whatever way I fix the imports.

Apache-HBase:
:broken_heart: **-1 overall**






| :blue_heart: |  reexec  |   0m 35s |  Docker mode activated.  |
||| _ Prechecks _ |
| :green_heart: |  dupname  |   0m  0s |  No case conflicting files found.  |
| :blue_heart: |  prototool  |   0m  1s |  prototool was not available.  |
| :green_heart: |  hbaseanti  |   0m  0s |  Patch does not have any anti-patterns.  |
| :green_heart: |  @author  |   0m  0s |  The patch does not contain any @author tags.  |
| :green_heart: |  test4tests  |   0m  0s |  The patch appears to include 8 new or modified test files.  |
| :blue_heart: |  mvndep  |   0m 35s |  Maven dependency ordering for branch  |
| :green_heart: |  mvninstall  |   5m 31s |  master passed  |
| :green_heart: |  compile  |   2m  3s |  master passed  |
| :green_heart: |  checkstyle  |   2m 22s |  master passed  |
| :green_heart: |  shadedjars  |   4m 38s |  branch has no errors when building our shaded downstream artifacts.  |
| :green_heart: |  javadoc  |   1m 16s |  master passed  |
| :blue_heart: |  spotbugs  |   4m  2s |  Used deprecated FindBugs config; considering switching to SpotBugs.  |
| :green_heart: |  findbugs  |   7m 39s |  master passed  |
||| _ Patch Compile Tests _ |
| :blue_heart: |  mvndep  |   0m 14s |  Maven dependency ordering for patch  |
| :green_heart: |  mvninstall  |   4m 54s |  the patch passed  |
| :green_heart: |  compile  |   2m  2s |  the patch passed  |
| :green_heart: |  cc  |   2m  2s |  the patch passed  |
| :green_heart: |  javac  |   2m  2s |  the patch passed  |
| :broken_heart: |  checkstyle  |   0m 38s |  hbase-client: The patch generated 10 new + 206 unchanged - 0 fixed = 216 total (was 206)  |
| :broken_heart: |  checkstyle  |   1m 26s |  hbase-server: The patch generated 27 new + 322 unchanged - 0 fixed = 349 total (was 322)  |
| :green_heart: |  whitespace  |   0m  0s |  The patch has no whitespace issues.  |
| :green_heart: |  shadedjars  |   4m 49s |  patch has no errors when building our shaded downstream artifacts.  |
| :green_heart: |  hadoopcheck  |  15m 41s |  Patch does not cause any errors with Hadoop 2.8.5 2.9.2 or 3.1.2.  |
| :green_heart: |  hbaseprotoc  |   2m  1s |  the patch passed  |
| :green_heart: |  javadoc  |   1m 12s |  the patch passed  |
| :green_heart: |  findbugs  |   8m 15s |  the patch passed  |
||| _ Other Tests _ |
| :green_heart: |  unit  |   0m 42s |  hbase-protocol-shaded in the patch passed.  |
| :green_heart: |  unit  |   1m 54s |  hbase-client in the patch passed.  |
| :green_heart: |  unit  | 161m 39s |  hbase-server in the patch passed.  |
| :green_heart: |  asflicense  |   1m 32s |  The patch does not generate ASF License warnings.  |
|  |   | 238m 53s |   |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | Client=19.03.4 Server=19.03.4 base: https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/2/artifact/out/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/781 |
| JIRA Issue | HBASE-18095 |
| Optional Tests | dupname asflicense javac javadoc unit spotbugs findbugs shadedjars hadoopcheck hbaseanti checkstyle compile cc hbaseprotoc prototool |
| uname | Linux 7587f3deda10 4.15.0-58-generic #64-Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64 GNU/Linux |
| Build tool | maven |
| Personality | /home/jenkins/jenkins-slave/workspace/HBase-PreCommit-GitHub-PR_PR-781/out/precommit/personality/provided.sh |
| git revision | master / 90007b7e80 |
| Default Java | 1.8.0_181 |
| checkstyle | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/2/artifact/out/diff-checkstyle-hbase-client.txt |
| checkstyle | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/2/artifact/out/diff-checkstyle-hbase-server.txt |
|  Test Results | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/2/testReport/ |
| Max. process+thread count | 5005 (vs. ulimit of 10000) |
| modules | C: hbase-protocol-shaded hbase-client hbase-server U: . |
| Console output | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-781/2/console |
| versions | git=2.11.0 maven=2018-06-17T18:33:14Z) findbugs=3.1.11 |
| Powered by | Apache Yetus 0.11.0 https://yetus.apache.org |


hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
wchevreuil: > because we need the service side meta/master tracking to happen using ZK and just the client side to use HMasterAsyncRegistry

Oh, didn't know that was the intention. I thought it was just fine to leave even server side processes acting as client to use the master registry. Is the concern here related to an rpc overload?

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestHMasterAsyncRegistry.java
wchevreuil: > This is purely a client side config

Yeah, that confused me. Am just wandering now what would be the impacts of having it also set at server side? Because that may become a common config mistake, if it's critical, maybe it's worth start thinking about split the connection creation path from client and server side?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestHMasterAsyncRegistryRPCs.java
wchevreuil: I think it's fine as it is, or if you prefer you can use mockito's expected functionality:

`@Test(expected = NumberFormatException.class)`

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetaRegionLocationCache.java
virajjasani: Oh, the control of no of retries is not supposed to be given to user?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncRegistryFactory.java
wchevreuil: I think we should keep it as a public constant, as per @bharathv explanation on the other comment.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
wchevreuil: Yeah, let's keep this constant public.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+  final RpcClient rpcClient;
+  final int rpcTimeout;
+
+  public HMasterAsyncRegistry(Configuration config) {
+    masterServers = new ArrayList<>();
+    conf = config;
+    parseHortPorts();
+    // Passing the default cluster ID means that the token based authentication does not work for
+    // certain client implementations.
+    // TODO(bharathv): Figure out a way to fetch the CLUSTER ID using a non authenticated way.
```
apurtell: One thought from the JIRA discussion was to add a servlet which does not require authentication. So, a separate instance of the infoserver stack, simply serving the cluster ID. A bit heavyweight, but an alternative like a separate instance of the RPC stack with auth requirements disabled serving only one iface/method isn't any less heavy.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
```diff
@@ -394,4 +394,7 @@ public Hbck getHbck(ServerName masterServer) throws IOException {
   Optional<MetricsConnection> getConnectionMetrics() {
     return metrics;
   }
+
+  @VisibleForTesting
+  AsyncRegistry getRegistry() { return registry; }
```
apurtell: Hmm. Fine for master and branch-2, but note a headache here for branch-1 (if a backport is desired). The precursor implementation in branch-1 is ClusterRegistry. AsyncRegistry was a big refactor via HBASE-16835. Not suggesting this needs be different, and some simple substitutions may get you most of the way there, but you may want to look at this.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncRegistryFactory.java
apurtell: So the idea here is the clients will instantiate one type of REGISTRY_IMPL, which contacts masters via RPC, and the masters and presumably all server side code will instantiate another type of REGISTRY_IMPL that talks to zookeeper directly? How is the distinction managed? Maybe I'll answer my own question upon further review...

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+ * comma separated list of host:port values. Should be set in conjunction with
+ * 'hbase.client.registry.impl' set to this class impl.
+ *
+ * The class does not cache anything. It is the responsibility of the callers to cache and
+ * avoid repeated requests.
+ */
+@InterfaceAudience.Private
+public class HMasterAsyncRegistry implements AsyncRegistry {
+  private static final Logger LOG = LoggerFactory.getLogger(HMasterAsyncRegistry.class);
+  public static final String CONF_KEY = "hbase.client.asyncregistry.masteraddrs";
```
apurtell: Nit: This doesn't need "asynchregistry" in the key name. We can anticipate multiple consumers of "hbase.client.master.addrs" (suggestion): HMasterAsyncRegistry, HMasterClusterRegistry (in branch-1), something else...

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+            DEFAULT_HBASE_RPC_TIMEOUT)));
+  }
+
+  private void parseHortPorts() {
+    String hostPorts = conf.get(CONF_KEY, DEFAULT_HOST_PORT);
+    for (String hostPort: hostPorts.split(",")) {
+      masterServers.add(ServerName.valueOf(hostPort, ServerName.NON_STARTCODE));
+    }
+    Preconditions.checkArgument(!masterServers.isEmpty(), String.format("%s is empty", CONF_KEY));
+    // Randomize so that not every client sends requests in the same order.
```
apurtell: Good

hbase-client/src/main/java/org/apache/hadoop/hbase/MetaRegionsNotAvailableException.java
```diff
@@ -0,0 +1,35 @@
...
+import org.apache.yetus.audience.InterfaceAudience;
+
+import java.io.IOException;
+
+/**
+ * Thrown by master when meta region locations are not cached whatever reason.
+ * Client is expected to retry when running into this.
+ */
+@InterfaceAudience.Private
+public class MetaRegionsNotAvailableException extends IOException  {
```
apurtell: Is there going to be other information served by this registry beside meta regions? Maybe not right now but do you think this mechanism may be overloaded in the future? It seems generally useful... masters as intermediary between clients (or servers) and what is direct access to zookeeper today. It would be better to not have a checked exception for each bit of information served via this means. Would IOException be sufficient?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
apurtell: Or even "hbase.master.addrs"

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+    });
+    return result;
+  }
+
+  /**
+   * Blocking RPC to get the cluster ID from the parsed master list. Returns null if no active
+   * master found.
+   */
+  private String getClusterIdHelper() throws  MasterNotRunningException {
+    // Loop through all the masters serially. We could be hitting some standby masters which cannot
```
apurtell: It's worth asking: Why do we need an active master to tell us what the cluster ID is? All the masters can look at the file in HDFS regardless of role. Standbys aren't special here, they can only serve one cluster, just like the active. If we fix this so any master can respond to this query then it's a small improvement in overall availability.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterRpcServices.java
```diff
@@ -977,6 +986,53 @@ public GetClusterStatusResponse getClusterStatus(RpcController controller,
...
+        HConstants.DEFAULT_MASTER_INFOPORT));
+    return response.build();
+  }
+
+  @Override
+  public GetClusterIdResponse getClusterId(RpcController controller, GetClusterIdRequest req)
+      throws ServiceException {
+    GetClusterIdResponse.Builder response = GetClusterIdResponse.newBuilder();
+    try {
+      master.checkInitialized();
```
apurtell: See my above comment. Initialize just this bit of info from HDFS if needed. Does not require active role.

hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
apurtell: I have the same question. Only the masters need to do things differently. All other clients, including embedded server side clients, should use the new service? So we get a nice common interposition point between all clients and the source of truth. Except the masters themselves, of course, which need to go directly to the source of truth.

hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
```diff
@@ -3067,6 +3075,19 @@ public Connection getConnection() throws IOException {
     return this.asyncConnection.toConnection();
   }
 
+  /**
+   * Creates a new Connection instance for this mini cluster with a given input config. Use it wisely since
+   * connection creation is expensive. For all practical purposes @link(getConnection()) should be good enough. This
+   * helper should only used if one wants to test a custom client side configuration that differs from the conf used to
+   * spawn the mini-cluster.
+   */
+  public AsyncClusterConnection getCustomConnection(Configuration conf) throws IOException {
```
apurtell: This is confusing.
"Custom"?

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestHMasterAsyncRegistry.java
```diff
@@ -0,0 +1,140 @@
...
+    hostAndPorts.add(HostAndPort.fromParts(masterHostName, 10001));
+    // 2. Invalid hostname but valid port
+    hostAndPorts.add(HostAndPort.fromParts("foo.bar", masterPort));
+    // 3. Invalid hostname and port.
+    hostAndPorts.add(HostAndPort.fromParts("foo.bar", 10003));
+    // 4. Finally valid host:port
+    hostAndPorts.add(HostAndPort.fromParts(masterHostName, masterPort));
+    final String config = Joiner.on(",").join(hostAndPorts);
+    conf.set(HMasterAsyncRegistry.CONF_KEY, config);
+    conf.set(AsyncRegistryFactory.REGISTRY_IMPL_CONF_KEY, HMasterAsyncRegistry.class.getName());
```
apurtell: Ok, I see where this is set for testing, but not where it's established as a default for clients.

apurtell: Should somehow distinguish between master and not-master roles and choose the right one, allowing for custom config override.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncRegistryFactory.java
wchevreuil: Same concern here. My understanding now, is that both server and client will rely on same config property, which may seem error prone to me. That's why I mention on my previous comment that maybe we should think on split the connection creation path between server and client based callers?

bharathv: Agree with the comments. Like @apurtell mentioned in another comment, I will work towards making this the default registry implementation for both clients and internal service connections (except Master which obviously goes to ZK, the source of truth). There is nothing in the implementation that prevents this. It was only done because the HBaseMiniCluster used in tests picks random ports (for running concurrent tests) and the clients don't know before hand what would the correct master port to use in the config. 

So if you see the pattern in tests, we wait for the mini cluster to be up, get the running master and it's port and then create a new Connection object based on that config. Once I figure out a way to force the mini-cluster to use certain known ports (without affecting the test concurrency ofcourse), we can get rid of the whole custom-config / split-config business. I'm looking into it. Hope it clarifies the intention.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
bharathv: Thats a good point. I'm looking into it.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterRpcServices.java
bharathv: Ack.

hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
bharathv: @apurtell Agree, that makes the whole design simpler. That is my end goal.

hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
bharathv: Will try to get rid of this.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+ * 'hbase.client.registry.impl' set to this class impl.
+ *
+ * The class does not cache anything. It is the responsibility of the callers to cache and
+ * avoid repeated requests.
+ */
+@InterfaceAudience.Private
+public class HMasterAsyncRegistry implements AsyncRegistry {
+  private static final Logger LOG = LoggerFactory.getLogger(HMasterAsyncRegistry.class);
+  public static final String CONF_KEY = "hbase.client.asyncregistry.masteraddrs";
+  private static final String DEFAULT_HOST_PORT = "localhost:" + HConstants.DEFAULT_MASTER_PORT;
```
xcangCRM: Is this "localhost" being default value for DEFAULT_HOST_PORT useful?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+import org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos.MasterService;
+import org.apache.hbase.thirdparty.com.google.common.base.Preconditions;
+import org.apache.hbase.thirdparty.com.google.protobuf.ServiceException;
+
+import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Fetches the meta information directly from HMaster by making relevant RPCs. HMaster RPC end
```
xcangCRM: nit: endpoint should be one word.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
xcangCRM: can you make this CONF_KEY more specific/ meaningful?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+  // RPC client used to talk to the master servers. This uses a stand alone RpcClient instance
+  // because AsyncRegistry is created prior to creating a cluster Connection. The client is torn
+  // down in close().
+  final RpcClient rpcClient;
+  final int rpcTimeout;
+
+  public HMasterAsyncRegistry(Configuration config) {
+    masterServers = new ArrayList<>();
+    conf = config;
+    parseHortPorts();
```
xcangCRM: typo here?
HortPort?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+    // Passing the default cluster ID means that the token based authentication does not work for
+    // certain client implementations.
+    // TODO(bharathv): Figure out a way to fetch the CLUSTER ID using a non authenticated way.
+    rpcClient = RpcClientFactory.createClient(conf, HConstants.CLUSTER_ID_DEFAULT);
+    rpcTimeout = (int) Math.min(Integer.MAX_VALUE,
```
xcangCRM: So you do toNanos here, is int good enough?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+        TimeUnit.MILLISECONDS.toNanos(conf.getLong(HBASE_RPC_TIMEOUT_KEY,
+            DEFAULT_HBASE_RPC_TIMEOUT)));
+  }
+
+  private void parseHortPorts() {
```
xcangCRM: method name typo

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+    String hostPorts = conf.get(CONF_KEY, DEFAULT_HOST_PORT);
+    for (String hostPort: hostPorts.split(",")) {
+      masterServers.add(ServerName.valueOf(hostPort, ServerName.NON_STARTCODE));
+    }
+    Preconditions.checkArgument(!masterServers.isEmpty(), String.format("%s is empty", CONF_KEY));
+    // Randomize so that not every client sends requests in the same order.
+    Collections.shuffle(masterServers);
```
xcangCRM: is this shuffle pseudo random that will return same or diff result from different hosts?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+    List<HBaseProtos.RegionLocation> result = null;
+    for (ServerName sname: masterServers) {
+      try {
+        MasterService.BlockingInterface stub = getMasterStub(sname);
+        HBaseRpcController rpcController = RpcControllerFactory.instantiate(conf).newController();
+        GetMetaLocationsResponse resp = stub.getMetaLocations(rpcController,
+            GetMetaLocationsRequest.getDefaultInstance());
+        result = resp.getLocationsList();
+      } catch (Exception e) {
+        LOG.warn("Error fetch meta locations from master {}. Trying others.", sname, e);
```
xcangCRM: nit: error fetching

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+    if (result == null || result.isEmpty()) {
+      throw new MetaRegionsNotAvailableException(String.format(
+          "Meta locations not found. Probed masters: %s", conf.get(CONF_KEY, DEFAULT_HOST_PORT)));
```
xcangCRM: nit:"Meta location not found"

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+    }
+    List<HRegionLocation> deserializedResult = new ArrayList<>();
+    result.stream().forEach(
+        location -> deserializedResult.add(ProtobufUtil.toRegionLocation(location)));
+    return new RegionLocations(deserializedResult);
+  }
+
+  /**
+   * Picks the first master entry from 'masterHortPorts' to fetch the meta region locations.
```
xcangCRM: typo:
masterHortPorts

hbase-client/src/main/java/org/apache/hadoop/hbase/client/HMasterAsyncRegistry.java
```diff
@@ -0,0 +1,236 @@
...
+    // Loop through all the masters serially. We could be hitting some standby masters which cannot
+    // process this RPC, so we just skip them.
+    for (ServerName sname: masterServers) {
+      try {
+        MasterService.BlockingInterface stub = getMasterStub(sname);
+        HBaseRpcController rpcController = RpcControllerFactory.instantiate(conf).newController();
+        GetClusterIdResponse resp =
+            stub.getClusterId(rpcController, GetClusterIdRequest.getDefaultInstance());
+        return resp.getClusterId();
+      } catch (IOException e) {
```
xcangCRM: so, what's the point separating IOException and ServiceException here?
 can you combine them?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetaRegionLocationCache.java
```diff
@@ -0,0 +1,185 @@
...
+    watcher = zkWatcher;
+    cachedMetaLocations = new CopyOnWriteArrayMap<>();
+    watcher.registerListener(this);
+    // Populate the initial snapshot of data from meta znodes.
+    // This is needed because stand-by masters can potentially start after the initial znode
+    // creation.
+    populateInitialMetaLocations();
+  }
+
+  private void populateInitialMetaLocations() {
```
xcangCRM: should we do exponential backoff retry here?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetaRegionLocationCache.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+    LOG.info("Meta znode for path {}: {}", path, opType.name());
+    int replicaId = watcher.getZNodePaths().getMetaReplicaIdFromPath(path);
+    if (opType == ZNodeOpType.DELETED) {
+      cachedMetaLocations.remove(replicaId);
+      return;
+    }
+    RegionState state = null;
+    int retries = 0;
+    while (retries++ < MAX_ZK_META_FETCH_RETRIES) {
```
xcangCRM: same as above : should we do exponential backoff retry here?

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestHMasterAsyncRegistry.java
```diff
@@ -0,0 +1,140 @@
...
+    HMaster master = TEST_UTIL.getMiniHBaseCluster().getMaster();
+    assertEquals(REGISTRY.getClusterId().get(), master.getClusterId());
+    assertEquals(REGISTRY.getMasterAddress().get(), master.getServerName());
+  }
+
+  /**
+   * Tests basic create, put, scan operations using the connection.
+   */
+  @Test
+  public void testCustomConnectionBasicOps() throws Exception {
```
ndimiduk: Can we not parameterize the various `TestFromClientSide*` classes to use both the new and the old methods?

hbase-protocol-shaded/src/main/protobuf/Master.proto
```diff
@@ -707,6 +737,19 @@ service MasterService {
   rpc GetClusterStatus(GetClusterStatusRequest)
     returns(GetClusterStatusResponse);
 
+  /** Returns whether this master is active or not. Served on both active/standby masters.*/
+  rpc IsActive(IsActiveRequest) returns(IsActiveResponse);
```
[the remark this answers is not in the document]
bharathv: https://github.com/apache/hbase/pull/812 - You are already a reviewer on that :-)

hbase-server/src/test/java/org/apache/hadoop/hbase/client/RegionReplicaTestHelper.java
```diff
@@ -89,6 +90,13 @@ static ServerName moveRegion(HBaseTestingUtility util, HRegionLocation currentLo
       .map(t -> t.getRegionServer().getServerName()).filter(sn -> !sn.equals(serverName)).findAny()
       .get();
     util.getAdmin().move(regionInfo.getEncodedNameAsBytes(), newServerName);
+    if (regionInfo.isMetaRegion()) {
+      // Invalidate the meta cache forcefully to avoid test races. Otherwise there might be a
```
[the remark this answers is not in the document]
bharathv: Undone.

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestHMasterAsyncRegistry.java
bharathv: https://github.com/apache/hbase/pull/807 is a blocker to parameterize this. Will do it once that is committed.

ndimiduk:
@bharathv this PR can be closed in favor of the individual subtasks, yeah?

bharathv:
@ndimiduk Yes.

Thanks to all the reviewers for the initial feedback. Those who are interested, please follow the PRs for subtasks of HBASE-18095.

PR #802, opened 2019-11-06
HBASE-23212 : Dynamically reload configs for Region Recovery chore

virajjasani:
Backport PR for branch-2
Jira: https://issues.apache.org/jira/browse/HBASE-23212


apurtell:
Testing locally, will apply if good

PR #879, opened 2019-11-26
HBASE-23343 Thrift Resolve FQDN from SPNEGO principal

karthikhw:
We need to manage different config groups in Ambari to run multiple thrift servers. This is because hbase thrift server will not able to resolve _HOST pattern from hbase.thrift.spnego.principal. so we explicitly specify the hostname for each group now.


karthikhw:
Thank you very much, @saintstack sir. 
 Sorry if am wrong. I simply thought to put all security-related stuff in one commonplace in HBase SecurityUtil.

I think we can close this pull request @saintstack. The same issue addressed via HBASE-23312 which is pushed on the same day.

saintstack:
Fixed by HBASE-23312 . Closing. Thanks @karthikhw

PR #966, opened 2019-12-26
HBASE-23620 KeyPrefixRegionSplitPolicy may cause split to fail when s…

wchevreuil:
…pecified prefix length is short enough that resulting split point ends up with same value of the parent start key.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/KeyPrefixRegionSplitPolicy.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/KeyPrefixRegionSplitPolicy.java
@@ -19,13 +19,18 @@ package org.apache.hadoop.hbase.regionserver;
 
 import java.util.Arrays;
 
+import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.yetus.audience.InterfaceAudience;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
  * A custom RegionSplitPolicy implementing a SplitPolicy that groups
- * rows by a prefix of the row-key
+ * rows by a prefix of the row-key.
+ *
+ * Note: If the row range is too big, it may not be possible to guarantee that the split point
+ * will have just the specified length, as the resulting key might be the same as the starting key
+ * of the parent.
  *
  * This ensures that a region is not split "inside" a prefix of a row key.
  * I.e. rows can be co-located in a region by their prefix.
@@ -79,8 +84,16 @@ public class KeyPrefixRegionSplitPolicy extends IncreasingToUpperBoundRegionSpli
     byte[] splitPoint = super.getSplitPoint();
     if (prefixLength > 0 && splitPoint != null && splitPoint.length > 0) {
       // group split keys by a prefix
-      return Arrays.copyOf(splitPoint,
+      int length = prefixLength;
+      byte[] result = Arrays.copyOf(splitPoint,
           Math.min(prefixLength, splitPoint.length));
+      while(Bytes.equals(result, region.getRegionInfo().getStartKey()) &&
+          length < splitPoint.length){
+        length++;
+        result = Arrays.copyOf(splitPoint,
+          Math.min(length, splitPoint.length));
+      }
+      return result;
     } else {
       return splitPoint;
     }
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestRegionSplitPolicy.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestRegionSplitPolicy.java
@@ -21,6 +21,7 @@ import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.when;
 
 import java.io.IOException;
 import java.util.ArrayList;
@@ -33,6 +34,7 @@ import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.HRegionInfo;
 import org.apache.hadoop.hbase.HTableDescriptor;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.RegionInfo;
 import org.apache.hadoop.hbase.testclassification.RegionServerTests;
 import org.apache.hadoop.hbase.testclassification.SmallTests;
 import org.apache.hadoop.hbase.util.Bytes;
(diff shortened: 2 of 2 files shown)


Apache-HBase:






| +0 :ok: |  reexec  |   1m 58s |  Docker mode activated.  |
||| _ Prechecks _ |
| +1 :green_heart: |  test4tests  |   0m  0s |  The patch appears to include 1 new or modified test files.  |
| +1 :green_heart: |  mvninstall  |   5m 33s |  master passed  |
| +1 :green_heart: |  compile  |   0m 57s |  master passed  |
| +1 :green_heart: |  checkstyle  |   1m 20s |  master passed  |
| +1 :green_heart: |  shadedjars  |   4m 43s |  branch has no errors when building our shaded downstream artifacts.  |
| +1 :green_heart: |  javadoc  |   0m 37s |  master passed  |
| +0 :ok: |  spotbugs  |   4m 31s |  Used deprecated FindBugs config; considering switching to SpotBugs.  |
| +1 :green_heart: |  findbugs  |   4m 30s |  master passed  |
||| _ Patch Compile Tests _ |
| +1 :green_heart: |  mvninstall  |   5m  3s |  the patch passed  |
| +1 :green_heart: |  compile  |   0m 58s |  the patch passed  |
| +1 :green_heart: |  javac  |   0m 58s |  the patch passed  |
| +1 :green_heart: |  checkstyle  |   1m 17s |  the patch passed  |
| +1 :green_heart: |  shadedjars  |   4m 36s |  patch has no errors when building our shaded downstream artifacts.  |
| +1 :green_heart: |  hadoopcheck  |  16m  2s |  Patch does not cause any errors with Hadoop 2.8.5 2.9.2 or 3.1.2.  |
| +1 :green_heart: |  javadoc  |   0m 36s |  the patch passed  |
| +1 :green_heart: |  findbugs  |   4m 35s |  the patch passed  |
||| _ Other Tests _ |
| +1 :green_heart: |  unit  | 151m 34s |  hbase-server in the patch passed.  |
| +1 :green_heart: |  asflicense  |   0m 32s |  The patch does not generate ASF License warnings.  |
|  |   | 211m 21s |   |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | Client=19.03.5 Server=19.03.5 base: https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-966/1/artifact/out/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/966 |
| Optional Tests | dupname asflicense javac javadoc unit spotbugs findbugs shadedjars hadoopcheck hbaseanti checkstyle compile |
| uname | Linux 77e2299bf25a 4.15.0-58-generic #64-Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64 GNU/Linux |
| Build tool | maven |
| Personality | /home/jenkins/jenkins-slave/workspace/HBase-PreCommit-GitHub-PR_PR-966/out/precommit/personality/provided.sh |
| git revision | master / 94346d8623 |
| Default Java | 1.8.0_181 |
|  Test Results | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-966/1/testReport/ |
| Max. process+thread count | 4383 (vs. ulimit of 10000) |
| modules | C: hbase-server U: hbase-server |
| Console output | https://builds.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-966/1/console |
| versions | git=2.11.0 maven=2018-06-17T18:33:14Z) findbugs=3.1.11 |


Apache9:
This breaks the assumption of this policy? The length of the split point could be greater than the prefixLength?

wchevreuil:
> This breaks the assumption of this policy? The length of the split point could be greater than the prefixLength?

Yeah, maybe worth log a WARN? Or fallback to the default policy in such condition? Leaving it as it is now, where it's completely breaking split seems the worse option to me.

Apache9:
This is a very important assumption, as it make sure that the rows with same prefix are in the same region. For example, I add user id as the prefix of the row key, and I want to do local secondary index on the user data, then it is very important to me that all the data belong to a user are in the same region.

So if you do not need this assumption, just do not use this split policy?

wchevreuil:
> For example, I add user id as the prefix of the row key, and I want to do local secondary index on the user data, then it is very important to me that all the data belong to a user are in the same region.

Yeah, most of the time the assumption would still be guaranteed. And some cases might not be so strict on the specific prefix value. But if the given row key prefix range can't fit in a region, what then, simply leave it grow indefinitely, and keep attempting split it even knowing it will fail?

Apache9:
Assumption means always right, there is no 'mostly right' assumption... Users choose this split policy means they want this assumption.

I think for an end user, the most important thing is monitoring, you need to know that a region is too large with this split policy, you need check whether this is expected, and try to fix it.

saintstack:
On the one hand, this fix is good; it caters for an issue the operator did not foresee. On the otherhand, as per @Apache9 , it does the 'fix' in violation of this KPRSP. Minimally I would think the class comment would change and we'd log the policy violations. Better would be a new policy that subclasses this one that is explicit that it will violate KPRSP if it fails to find an obliging split?

Apache9:
> On the one hand, this fix is good; it caters for an issue the operator did not foresee. On the otherhand, as per @Apache9 , it does the 'fix' in violation of this KPRSP. Minimally I would think the class comment would change and we'd log the policy violations. Better would be a new policy that subclasses this one that is explicit that it will violate KPRSP if it fails to find an obliging split?

If you do not need this assumption then just do not use this policy? It is not the default one I'd say...

wchevreuil:
This specific case was a customer using [yarn timeline service v2](https://hadoop.apache.org/docs/current/hadoop-yarn/hadoop-yarn-site/TimelineServiceV2.html). Internally, it defines a KPRSP with prefix length of 4, which in this case had problems to split regions for rows representing entries for 'hdfs' user.  YARN-10077 has been filled to increase the prefix length in the table definition.

> If you do not need this assumption then just do not use this policy?

Some applications may not need it as a strict rule. @Apache9 arguments, though, are valid. Maybe a better solution then is to implement this flexible behaviour as another policy (like @saintstack suggested). But for KPRSP, I think it should at least apply the validation and throw an exception if it can't get a valid split point, other than leaving the procedure moving forward and failing further to perform the split.

joshelser:
>  Leaving it as it is now, where it's completely breaking split seems the worse option to me.

> But for KPRSP, I think it should at least apply the validation and throw an exception if it can't get a valid split point, other than leaving the procedure moving forward and failing further to perform the split.

So, I can see the exception on Jira, but how does this manifest? Is this a side-effect of pv2 where the split procedure just keeps getting retried? Is it a simple thing for a SplitPolicy to "defer" making a split when we detect the case where we'd generate an invalid split point?

Agree with Duo that we can't change KPRSP's default functionality as that could seriously break users. I could see a new SplitPolicy which encapsulates your change, Wellington.

saintstack:
Was going to close this out @wchevreuil if good by you sir?

PR #1006, opened 2020-01-09
HBASE-23045.currentPath may be stitched in a loop in replication source code.

Gkkkk302:
…ce code.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/ReplicationSource.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/ReplicationSource.java
@@ -801,9 +801,9 @@ public class ReplicationSource extends Thread
                 Path p = rs.getPath();
                 FileStatus[] logs = fs.listStatus(p);
                 for (FileStatus log : logs) {
-                  p = new Path(p, log.getPath().getName());
-                  if (p.getName().equals(currentPath.getName())) {
-                    currentPath = p;
+                  String logName = log.getPath().getName();
+                  if (logName.equals(currentPath.getName())) {
+                    currentPath = new Path(p, logName);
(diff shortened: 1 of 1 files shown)


virajjasani:
@Gkkkk302 Thanks for the PR, changes look good, please incorporate Unit Test for this fix.

virajjasani:
This seems applicable to all active branch-1s only and not to branch-2+

virajjasani:
Can you please also confirm @Gkkkk302 ?

Gkkkk302:
Yes, I agree. I will try to provide unit tests next. Thank you @virajjasani

virajjasani:
Thanks for the fix @Gkkkk302 
Could you please also validate branch-1 patch to ensure if we can get this fix to branch-1 since it's good to have this fixed as part of 1.6 release.
Thanks

branch-1 patch:

```
diff --git a/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/ReplicationSource.java b/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/ReplicationSource.java
index 5acb70922f..a26bbd00cd 100644
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/ReplicationSource.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/ReplicationSource.java
@@ -938,8 +938,9 @@ public class ReplicationSource extends Thread implements ReplicationSourceInterf
         Path p = rs.getPath();
         FileStatus[] logs = fs.listStatus(p);
         for (FileStatus log : logs) {
-          p = new Path(p, log.getPath().getName());
-          if (p.getName().equals(path.getName())) {
+          String logName = log.getPath().getName();
+          if (logName.equals(path.getName())) {
+            p = new Path(p, log.getPath().getName());
             LOG.info("Log " + p.getName() + " found at " + p);
             return p;
           }
```

PR #1087, opened 2020-01-24
HBASE-23734 Backport [HBASE-21874 Bucket cache on Persistent memory] …

wchevreuil:
…to branch-2.2


shenshengli:
Only one directory can be configured here, but multiple hardware is usually used, such as 1T (8*128G) App Direct, in which case I want to configure 512G bucketCache, must poolSet be used?

PR #1120, opened 2020-02-04
HBASE-23787: TestSyncTimeRangeTracker fails quite easily and allocate…

markrmiller:
…s a very expensive array.

Here is a first pass, says it can't merge for some reason so I will be back for draft 2.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
@@ -20,6 +20,8 @@ package org.apache.hadoop.hbase.regionserver;
 import static org.junit.Assert.assertTrue;
 
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicLong;
+import java.util.concurrent.atomic.AtomicLongArray;
 import org.apache.hadoop.hbase.HBaseClassTestRule;
 import org.apache.hadoop.hbase.testclassification.RegionServerTests;
 import org.apache.hadoop.hbase.testclassification.SmallTests;
@@ -34,7 +36,7 @@ public class TestSyncTimeRangeTracker extends TestSimpleTimeRangeTracker {
   public static final HBaseClassTestRule CLASS_RULE =
       HBaseClassTestRule.forClass(TestSyncTimeRangeTracker.class);
 
-  private static final int NUM_KEYS = 10000000;
+  private static final int NUM_KEYS = 1000000;
   private static final int NUM_OF_THREADS = 20;
 
   @Override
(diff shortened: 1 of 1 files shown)


saintstack:
Your branch up-to-date Mr. Miller?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
```diff
@@ -84,23 +86,23 @@ public void run() {
...
   }
 
-  static class RandomTestData {
-    private long[] keys = new long[NUM_KEYS];
-    private long min = Long.MAX_VALUE;
-    private long max = 0;
+ static class RandomTestData {
+    private final AtomicLongArray keys = new AtomicLongArray(NUM_KEYS);
+    private long min = Long.MAX_VALUE; // effectively final
+    private long max = 0; // effectively final
```
saintstack: Don't want to make them final?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
```diff
@@ -84,23 +86,23 @@ public void run() {
...
 
     public RandomTestData() {
       if (ThreadLocalRandom.current().nextInt(NUM_OF_THREADS) % 2 == 0) {
         for (int i = 0; i < NUM_KEYS; i++) {
-          keys[i] = i + ThreadLocalRandom.current().nextLong(NUM_OF_THREADS);
-          if (keys[i] < min) min = keys[i];
-          if (keys[i] > max) max = keys[i];
+          keys.set(i, i + ThreadLocalRandom.current().nextLong(NUM_OF_THREADS));
+          if (keys.get(i) < min) min = keys.get(i);
```
saintstack: See your checkstyle. Needs parents around the 'min = key[i];'

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
```diff
@@ -84,23 +86,23 @@ public void run() {
     assertTrue(trr.getMin() == 0);
   }
 
-  static class RandomTestData {
-    private long[] keys = new long[NUM_KEYS];
-    private long min = Long.MAX_VALUE;
-    private long max = 0;
+ static class RandomTestData {
+    private final AtomicLongArray keys = new AtomicLongArray(NUM_KEYS);
```
saintstack: This is the 'fix'? Using atomic array. Sounds good.

Apache9: But this array will only be accessed by one thread? Why do we need to use AtomicLongArray?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
```diff
@@ -34,7 +36,7 @@
   public static final HBaseClassTestRule CLASS_RULE =
       HBaseClassTestRule.forClass(TestSyncTimeRangeTracker.class);
 
-  private static final int NUM_KEYS = 10000000;
+  private static final int NUM_KEYS = 1000000;
```
Apache9: So this is the actual fix right?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
busbey: they get assigned like right below here.

saintstack:
@markrmiller See comments above?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
markrmiller: Basically I ran this test a lot before and after these changes to try and confirm a fix and this is what seemed to work.

Technically, even if a single thread accesses a data structure, if it is a different thread than the one that created it, you still need to publish objects correctly for multiple threads - concurrency is not required. So they need a mem barrier or to be final or to be effectively final or a 32-bit primitive(I think?).

markrmiller: Yeah, so AtomicLongArray to ensure we see an up to date value.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
markrmiller: This is kind of separate. I noticed this huge array took a lot of RAM and it did not seem necessary.

I also noticed that the array is so large that the threads working on it don't overlap much as is.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestSyncTimeRangeTracker.java
markrmiller: Right, would be ideal to make them final, settling for marking them effectively final.

markrmiller:
Sorry, been chasing a local clean test run.

The AtomicLongArray was need for these tests to pass more regularly for me to ensure we always look at an up to date value.

The memory issue seemed wasteful and a bit off given the goals of the test and also counter productive to running test like this in parallel.

I've updated my pr.

saintstack:
No response on above. Closing this out.

PR #1732, opened 2020-05-18
HBASE-24312 porting HBASE-24199 to 2.1 branches

gkanade:


ramkrish86:
Thanks for the patch @gkanade .

PR #2165, opened 2020-07-29
[Subtask of HBASE-20717] Chunk and ChunkPool structure

Reidddddd:
I'm going to further breakdown the HBASE-20717 into several tasks.
The first one will be this: Chunk and ChunkPool structure.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/ccsmap/ChunkPool.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/ccsmap/ChunkPool.java
@@ -1,0 +1,199 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ * <p>
+ * http://www.apache.org/licenses/LICENSE-2.0
+ * <p>
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hbase.ccsmap;
+
+import java.util.Map;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicLong;
+
+import com.google.common.annotations.VisibleForTesting;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.hadoop.conf.Configuration;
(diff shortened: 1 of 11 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/ccsmap/HugeHeapChunk.java
```diff
@@ -0,0 +1,29 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.hadoop.hbase.ccsmap;
+
+
+public class HugeHeapChunk extends OnHeapChunk {
```
Reidddddd, line 22: Oh, this one is redundant currently, let's not consider the huge situations and focus on general situation.

carp84: Do you mean we will add `HugeHeapChunk` in another PR? If so, could you remove this class from this PR to make it clear? Thanks.

Reidddddd: Removed.

hbase-server/src/main/java/org/apache/hadoop/hbase/ccsmap/ChunkPool.java
```diff
@@ -0,0 +1,208 @@
...
+
+  private static final Log LOG = LogFactory.getLog(ChunkPool.class);
+  private static final Object initLock = new Object();
+  private static volatile ChunkPool globalInstance;
+
+  private final HeapMode heapMode;
+  private final int maxChunkCount;
+  private final int chunkSize;
+  private final ChunkAllocator chunkAllocator;
+  private final AbstractHeapChunk[] chunkArray;
```
Reidddddd, line 48: I have no idea what this array is for, but it can provide a O(1) time search. Thinking it will be up to the stage when we proceed further.

hbase-server/src/main/java/org/apache/hadoop/hbase/ccsmap/ChunkPool.java
```diff
@@ -0,0 +1,208 @@
...
+    long capacity;
+    int chunkSize;
+    int initialCount;
+    HeapMode heapMode;
+
+    ChunkPoolParameters(Configuration conf) {
+      capacity = conf.getLong(CCSMapUtils.CHUNK_CAPACITY_KEY, Long.MIN_VALUE);
+      chunkSize = conf.getInt(CCSMapUtils.CHUNK_SIZE_KEY, Integer.MIN_VALUE);
+      initialCount = conf.getInt(CCSMapUtils.INITIAL_CHUNK_COUNT_KEY, Integer.MIN_VALUE);
+      heapMode = conf.getBoolean(CCSMapUtils.USE_OFFHEAP, false) ?
```
Reidddddd, line 197: These parts are not very sensible by now. I plan to do a sanity check at the higher level (after we proceed there), so we can make sure when CCSMap is on, all needed configurations must be set. Something like this:
```
sanityCheck();
ChunkPool. initialize(conf);
```

Reidddddd: That's why the default value looks ugly.

PR #2170, opened 2020-07-29
HBASE-24788: Fix the connection leaks on getting hbase admin from unc…

bharathv:
…losed connection

Rewrote the patch for branch-1 since master has significanly diverged.

(cherry picked from commit dc5ef7af1f8b9e386495a73924c9442203f65a77)

Co-authored-by: Bharath Vissapragada <bharathv@apache.org>


bharathv:
It looks like this pattern is not as frequent in branch-1. Fixed the actual leak in TableOutputFormat and cleaned up a few other places to not make unnecessary connections.

PR #2159, opened 2020-07-28
HBASE-20226: Parallelize region manifest deletes

bharathv:
We observed this delete call to be a bottleneck for table with lots of
regions. Patch attempts to parrallelize them.

PR #2171, opened 2020-07-29
HBASE-20226: Parallelize region manifest deletes (#2159)

bharathv:
We observed this delete call to be a bottleneck for table with lots of
regions. Patch attempts to parallelize them.

Signed-off-by: Andrew Purtell <apurtell@apache.org>
(cherry picked from commit f07f30ae24e6d0e9151bb76aebf78928ec9572e3)


bharathv:
branch-1 precommit is broken, I ran the full test suite in a private jenkins environment. Aside from the usual flakes, no test failures.

PR #2180, opened 2020-07-31
HBASE-24805 HBaseTestingUtility.getConnection should be threadsafe

busbey:
This maintains *most* compatibility while changing the handling of the shared connection to threadsafe. Note about compat as comments.

PR #2188, opened 2020-08-03
HBASE-24805 HBaseTestingUtility.getConnection should be threadsafe (branch-1)

busbey:
* refactor how we use connection to rely on the access method
* refactor initialization and cleanup of the shared connection
* incompatibly change HCTU's Configuration member variable to be final so it can be safely accessed from multiple threads.

Closes #2180

adapted for jdk7

(cherry picked from commit 86ebbdd8a2df89de37c2c3bd50e64292eaf28b11)
(cherry picked from commit 0806349adab338330428c900588234d7f6fcfcc2)

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
@@ -46,6 +46,7 @@ import java.util.Random;
 import java.util.Set;
 import java.util.TreeSet;
 import java.util.UUID;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.io.FileUtils;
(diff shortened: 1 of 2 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtility.java
```diff
@@ -3020,16 +3011,26 @@ public HBaseCluster getHBaseClusterInterface() {
...
+    while (connection == null) {
+      connection = ConnectionFactory.createConnection(this.conf);
+      if (! this.connection.compareAndSet(null, connection)) {
+        try {
+          connection.close();
+        } catch (IOException exception) {
+          LOG.debug("Ignored failure while closing connection on contended connection creation.",
+              exception);
+        }
+        connection = this.connection.get();
```
virajjasani, line 3030: This is perfect, but for a while I got confused with `connection` being Connection and `this.connection` being AtomicReference. 
Can we rename `connection` to `connectionRef` to indicate AtomicReference?

busbey: sure! should I update master and branches-2 to similarly use `asyncConnectionRef` instead of `asyncConnection`? Or less confusing there because the instance and local names are already different?

virajjasani: I think branch-2+ seem good because names are different already 👍

PR #2289, opened 2020-08-20
HBASE-24918 Make RegionInfo#UNDEFINED IA.Private

saintstack:
Mark RegionInfo#UNDEFINED IA.Private and deprecated;
it is for internal use only and likely to be removed in hbase4.

Move MutableRegionInfo out of RegionInfoBuilder and have it as a
stanadlone task; a nice-to-have.


saintstack:
This patch is same as last PR attached to HBASE-24896 only no claims it fixes HBASE-24896....It is deprecation, annotation, and moving an inner-class out to be standalone only.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfo.java
```diff
@@ -69,7 +66,15 @@
  */
 @InterfaceAudience.Public
 public interface RegionInfo extends Comparable<RegionInfo> {
-  RegionInfo UNDEFINED = RegionInfoBuilder.newBuilder(TableName.valueOf("__UNDEFINED__")).build();
+  /**
+   * @deprecated since 2.3.2/3.0.0; to be removed in 4.0.0 with no replacement (for internal use).
+   */
+  @Deprecated
+  @InterfaceAudience.Private
+  // Not using RegionInfoBuilder intentionally to avoid a static loading deadlock: HBASE-24627
```
bharathv, line 74: I think this statement can be omitted since this is being handled separately in the parent task.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfo.java
```diff
@@ -69,7 +66,15 @@
...
+  RegionInfo UNDEFINED = new MutableRegionInfo(0, TableName.valueOf("__UNDEFINED__"),
```
bharathv, line 75: Switch back to RegionInfoBuilder? (same below?)

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfo.java
saintstack: Makes sense.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfo.java
saintstack: Done

saintstack:
Push addresses @bharathv review

bharathv:
@saintstack Looks like the push didn't address the comments or is it just me?

saintstack:
@bharathv  my fault... loaded but not committed. Thanks.

PR #2291, opened 2020-08-22
HBASE-24776 [hbtop] Support Batch mode

brfrn169:
Added the following command line parameters to hbtop:
| Argument | Description | 
|---|---|
| -n,--numberOfIterations <arg> | The number of iterations |
| -O,--outputFieldNames | Print each of the available field names on a separate line, then quit |
| -f,--fields <arg> | Show only the given fields. Specify comma separated fields to show multiple fields |
| -s,--sortField <arg> | The initial sort field. You can prepend a `+' or `-' to the field name to also override the sort direction. A leading `+' will force sorting high to low, whereas a `-' will ensure a low to high ordering |
| -i,--filters <arg> | The initial filters. Specify comma separated filters to set multiple filters |
| -b,--batchMode | Starts hbtop in Batch mode, which could be useful for sending output from hbtop to other programs or to a file. In this mode, hbtop will not accept input and runs until the iterations limit you've set with the `-n' command-line option or until killed |

### DIFF
--- a/hbase-hbtop/src/main/java/org/apache/hadoop/hbase/hbtop/HBTop.java
+++ b/hbase-hbtop/src/main/java/org/apache/hadoop/hbase/hbtop/HBTop.java
@@ -17,11 +17,17 @@
  */
 package org.apache.hadoop.hbase.hbtop;
 
+import java.util.ArrayList;
+import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
+import java.util.stream.Collectors;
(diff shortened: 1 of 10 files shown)


brfrn169:
@joshelser Can you please review this when you get a chance? Thanks.

saintstack (approved):
Skimmed.

brfrn169:
@saintstack Thank you very much for reviewing this! I will merge it.

PR #2296, opened 2020-08-23
HBASE-24932 Reduce the number of builds to keep for flaky find job

Apache9:


virajjasani (commented):
Do we want 10 maybe? 1/5th reduction for find should be good?

virajjasani:
I just saw Sean's comments over Jira reg holding on to this change.

PR #2467, opened 2020-09-28
HBASE-25109 Add MR Counters to WALPlayer; currently hard to tell if i…

saintstack:
…t is doing anything

Add MR counters so operator can see if WALPlayer run actually did
anything. Fix bugs in usage (it enforced two args though usage
describes allowing one arg only). Clean up usage output. In
particular add note on wal file separator as hbase by default uses
the ',' in its WAL file names which could befuddle operator
trying to do simple import.


saintstack:
Close in favor of 2468

PR #2481, opened 2020-09-30
HBASE-25115 HFilePrettyPrinter can't seek to the row which is the first row of a hfile

ZhongChaoqiang:
…st row of a hfile


saintstack:
Looks good but why do checks show as failed?

virajjasani:
Previous build issue didn't seem relevant to the patch, just re-triggered the build.

PR #2547, opened 2020-10-15
HBASE-25185 Remove the usage of MetaTableAccessor in hbase-balancer

saintstack:
Undoes dependency on MetaTableAccessor from hbase-balancer.
Moves MetaTableAccessor from hbase-balancer to hbase-server.
Adds a new hbase-tool module to host hbase tooling. In this
commit, we move favored nodes tooling here. In follow-on will
move other tooling here.

hbase-balancer/src/main/java/org/apache/hadoop/hbase/favored/FavoredNodeAssignmentHelper.java
hbase-server/src/main/java/org/apache/hadoop/hbase/favored/FavoredNodesManager.java
 Moved the write-to-meta methods over to hbase-server into
 FavoredNodeManager from FavoredNodesAssignmentHelper where they are needed.
 Made the methods priivate.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/SnapshotOfRegionAssignmentFromMeta.java
 Moved this class back into hbase-server. It wants to scan meta
 to calculate data structures already present in Master. Needs
 rewrite to use Master if Favored Nodes feature is revived.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestMasterOperationsForRegionReplicas.java
 Bug fix. Was using wrong logger.

hbase-tool/src/main/java/org/apache/hadoop/hbase/tool/favored/AssignmentVerificationReport.java
hbase-tool/src/main/java/org/apache/hadoop/hbase/tool/favored/RegionPlacementMaintainer.java
 Moved favored node tool and the reporter it uses here.
 Updated class comments to explain it unused, needs work.

hbase-tool/src/test/java/org/apache/hadoop/hbase/tool/favored/TestRegionPlacement.java
 Moved test for RegionPlacemenentMaintainer here.


Apache9:
What is the scope of the new hbase-tool module? We want to move all the classes under the tool package in hbase-server to this module in the future?

saintstack:
> What is the scope of the new hbase-tool module? We want to move all the classes under the tool package in hbase-server to this module in the future?

Thanks for asking. I should have been more explicit. It might be better to add hbase-tool as a separate issue so I do it proper justification. Let me do that, HBASE-25190. Meantime, we'd move standalone tools and their tests here especially those from hbase-server: for example under the 'tool' package in hbase-server we currently have at least BulkLoadHFilesTool (and associated files) and the Canary. These seem like good candidates.

Let me see if I can land HBASE-25190 first and then come back here.

PR #113, opened 2019-04-06
HBASE-22174 Remove error prone from our precommit javac check

Apache9:
The output of the error prone check is not stable, sometimes we miss several warnings, and sometimes the order of the warnings are changed. Remove it for now. Can add back later after we fix all these problems. Or another solution is to add it back as a separated check in pre commit and add it to the filter list, so it will only generate a -0 instead of -1.

PR #2570, opened 2020-10-20
HBASE-25126 Add load balance logic in hbase-client to distribute read…

huaxiangsun:
… load over meta replica regions

It adds load balance support. With "hbase.meta.replicas.use" set to true, client support meta replica feature.
The existing mode is called HighAvailable, client sends scan request to the primary meta replica region first,
if response is not back within a configured amount of time, it will send scan requests to all meta replica regions and
take the first response. On top of the existing mode, a new mode LoadBalance is added. In this mode, client first
choose a meta replica region to send scan request. If the response is stale, it will send the scan request to the primary
meta region. In this mode, all meta replica regions serve scan requests.

Two new config knobs are added:

1. hbase.meta.replicas.mode
   Valid value is "HighAvailable" and "LoadBalance".

2. hbase.meta.replicas.mode.loadbalance.replica.chooser
   Implementation class for MetaReplicaChooser. Only org.apache.hadoop.hbase.client.MetaReplicaLoadBalanceReplicaSimpleChooser.class
   is supported for now, which is the default as well.

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
@@ -17,8 +17,11 @@
  */
 package org.apache.hadoop.hbase.client;
 
+import static org.apache.hadoop.hbase.HConstants.DEFAULT_USE_META_REPLICAS;
+import static org.apache.hadoop.hbase.HConstants.META_REPLICAS_MODE;
 import static org.apache.hadoop.hbase.HConstants.STATUS_PUBLISHED;
 import static org.apache.hadoop.hbase.HConstants.STATUS_PUBLISHED_DEFAULT;
+import static org.apache.hadoop.hbase.HConstants.USE_META_REPLICAS;
 import static org.apache.hadoop.hbase.client.ClusterStatusListener.DEFAULT_STATUS_LISTENER_CLASS;
 import static org.apache.hadoop.hbase.client.ClusterStatusListener.STATUS_LISTENER_CLASS;
 import static org.apache.hadoop.hbase.client.ConnectionUtils.NO_NONCE_GENERATOR;
@@ -116,7 +119,7 @@ class AsyncConnectionImpl implements AsyncConnection {
   private final Optional<ServerStatisticTracker> stats;
   private final ClientBackoffPolicy backoffPolicy;
 
-  private ChoreService authService;
+  private ChoreService choreService;
 
   private final AtomicBoolean closed = new AtomicBoolean(false);
 
@@ -130,8 +133,12 @@ class AsyncConnectionImpl implements AsyncConnection {
       SocketAddress localAddress, User user) {
     this.conf = conf;
     this.user = user;
-    if (user.isLoginFromKeytab()) {
-      spawnRenewalChore(user.getUGI());
+
+    if (user.isLoginFromKeytab() || isMetaReplicaLBMode()) {
+      startChoreService();
+      if (user.isLoginFromKeytab()) {
+        spawnRenewalChore(user.getUGI());
+      }
     }
     this.connConf = new AsyncConnectionConfiguration(conf);
     this.registry = registry;
@@ -181,9 +188,16 @@ class AsyncConnectionImpl implements AsyncConnection {
     this.clusterStatusListener = listener;
   }
 
+  private void startChoreService() {
+    choreService = new ChoreService("AsyncConn Chore Service");
+  }
+
   private void spawnRenewalChore(final UserGroupInformation user) {
-    authService = new ChoreService("Relogin service");
-    authService.scheduleChore(AuthUtil.getAuthRenewalChore(user));
+    choreService.scheduleChore(AuthUtil.getAuthRenewalChore(user));
+  }
+
+  public ChoreService getChoreService() {
+    return this.choreService;
   }
 
   @Override
@@ -208,8 +222,8 @@ class AsyncConnectionImpl implements AsyncConnection {
     IOUtils.closeQuietly(clusterStatusListener);
     IOUtils.closeQuietly(rpcClient);
     IOUtils.closeQuietly(registry);
-    if (authService != null) {
-      authService.shutdown();
+    if (choreService != null) {
+      choreService.shutdown();
     }
     metrics.ifPresent(MetricsConnection::shutdown);
     ConnectionOverAsyncConnection c = this.conn;
@@ -390,6 +404,16 @@ class AsyncConnectionImpl implements AsyncConnection {
       this.conn = c;
     }
     return c;
+  }
+
+  private boolean isMetaReplicaLBMode() {
+    if (conf.getBoolean(USE_META_REPLICAS, DEFAULT_USE_META_REPLICAS)) {
+      String metaReplicaMode = conf.get(META_REPLICAS_MODE, "");
+      if (MetaReplicaMode.LoadBalance.toString().equals(metaReplicaMode)) {
+        return true;
+      }
+    }
+    return false;
   }
 
   @Override
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
@@ -17,8 +17,12 @@
  */
 package org.apache.hadoop.hbase.client;
 
+import static org.apache.hadoop.hbase.HConstants.DEFAULT_META_REPLICA_NUM;
 import static org.apache.hadoop.hbase.HConstants.DEFAULT_USE_META_REPLICAS;
 import static org.apache.hadoop.hbase.HConstants.EMPTY_END_ROW;
+import static org.apache.hadoop.hbase.HConstants.META_REPLICAS_MODE;
+import static org.apache.hadoop.hbase.HConstants.META_REPLICAS_MODE_LOADBALANCE_REPILCA_CHOOSER;
+import static org.apache.hadoop.hbase.HConstants.META_REPLICAS_NUM;
 import static org.apache.hadoop.hbase.HConstants.NINES;
 import static org.apache.hadoop.hbase.HConstants.USE_META_REPLICAS;
 import static org.apache.hadoop.hbase.HConstants.ZEROES;
@@ -34,6 +38,7 @@ import static org.apache.hadoop.hbase.util.Bytes.BYTES_COMPARATOR;
 import static org.apache.hadoop.hbase.util.ConcurrentMapUtils.computeIfAbsent;
 
 import java.io.IOException;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.Iterator;
@@ -90,6 +95,11 @@ class AsyncNonMetaRegionLocator {
   private final int locatePrefetchLimit;
 
   private final boolean useMetaReplicas;
+
+  // When useMetaReplicas is true, the mode tells if HighAvailable, LoadBalance mode is supported.
+  // The default mode is HighAvailable.
+  private final MetaReplicaMode metaReplicaMode;
+  private MetaReplicaLoadBalanceReplicaChooser metaReplicaChooser;
 
   private final ConcurrentMap<TableName, TableCache> cache = new ConcurrentHashMap<>();
 
@@ -198,6 +208,36 @@ class AsyncNonMetaRegionLocator {
       conn.getConfiguration().getInt(LOCATE_PREFETCH_LIMIT, DEFAULT_LOCATE_PREFETCH_LIMIT);
     this.useMetaReplicas =
       conn.getConfiguration().getBoolean(USE_META_REPLICAS, DEFAULT_USE_META_REPLICAS);
+
+    if (this.useMetaReplicas) {
+      if (conn.getConfiguration()
+        .get(META_REPLICAS_MODE, "").equals(MetaReplicaMode.LoadBalance.toString())) {
+        this.metaReplicaMode = MetaReplicaMode.LoadBalance;
+        int numOfMetaReplicas = conn.getConfiguration().getInt(
+          META_REPLICAS_NUM, DEFAULT_META_REPLICA_NUM);
+        if (numOfMetaReplicas <= 1) {
+          LOG.error("Configured to support meta replica load balance mode,"
+            + " but there is no meta replica configured");
+          this.metaReplicaChooser = null;
+        } else {
+          String metaReplicaChooserImpl = conn.getConfiguration().get(
+            META_REPLICAS_MODE_LOADBALANCE_REPILCA_CHOOSER,
+            MetaReplicaLoadBalanceReplicaSimpleChooser.class.getName());
+          try {
+            this.metaReplicaChooser = Class.forName(metaReplicaChooserImpl)
+              .asSubclass(MetaReplicaLoadBalanceReplicaChooser.class)
+              .getDeclaredConstructor(AsyncConnectionImpl.class).newInstance(this.conn);
+          } catch (InstantiationException | IllegalAccessException | ClassNotFoundException
+            | NoSuchMethodException | InvocationTargetException e) {
+            throw new IllegalArgumentException(e);
+          }
+        }
+      } else {
+        this.metaReplicaMode = MetaReplicaMode.HighAvailable;
+      }
+    } else {
+      this.metaReplicaMode = MetaReplicaMode.None;
+    }
   }
 
   private TableCache getTableCache(TableName tableName) {
@@ -435,6 +475,14 @@ class AsyncNonMetaRegionLocator {
       .setReadType(ReadType.PREAD);
     if (useMetaReplicas) {
       scan.setConsistency(Consistency.TIMELINE);
+
+      // If it is meta replica LoadBalance mode, pick up a meta replica to scan. It will try to
+      // pick up meta replica region to scan, in certain error cases, it may pick primary meta
+      // region to scan.
+      if (isMetaReplicaLBMode()) {
+        scan.setReplicaId(this.metaReplicaChooser.chooseReplicaToGo(tableName,
+          req.row, req.locateType));
+      }
     }
     conn.getTable(META_TABLE_NAME).scan(scan, new AdvancedScanResultConsumer() {
 
@@ -444,6 +492,8 @@ class AsyncNonMetaRegionLocator {
 
       @Override
       public void onError(Throwable error) {
+        // TODO: if it fails, with meta replica load balance, it may try with another meta
+        // replica. This improvement will be done later.
         complete(tableName, req, null, error);
       }
 
@@ -577,6 +627,15 @@ class AsyncNonMetaRegionLocator {
       if (!canUpdateOnError(loc, oldLoc)) {
         return;
       }
+      // Tell metaReplicaChooser that the location is stale. It will create a stale entry
+      // with timestamp internally. Next time the client looks up the same location,
+      // it will pick a different meta replica region. For the current implementation,
+      // the metaReplicaId is not used, so the primary one is passed in.
+      if (isMetaReplicaLBMode()) {
+        // TODO: for now, metaReplicaId is not used in simplerChooser, default to the primary one.
+        metaReplicaChooser.updateCacheOnError(loc, RegionInfo.DEFAULT_REPLICA_ID);
+      }
+
       RegionLocations newLocs = removeRegionLocation(oldLocs, loc.getRegion().getReplicaId());
       if (newLocs == null) {
         if (tableCache.cache.remove(startKey, oldLocs)) {
@@ -590,6 +649,11 @@ class AsyncNonMetaRegionLocator {
         }
       }
     }
+  }
+
+  public boolean isMetaReplicaLBMode() {
+    return this.useMetaReplicas && (this.metaReplicaMode == MetaReplicaMode.LoadBalance) &&
+      (this.metaReplicaChooser != null);
   }
 
   private void addLocationToCache(HRegionLocation loc) {
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
@@ -1,0 +1,53 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.client;
+
+import org.apache.hadoop.hbase.HRegionLocation;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * There are two modes with meta replica support.
+ *   HighAvailable    - Client sends requests to the primary meta region first, within a
+ *                      configured amount of time, if  there is no response coming back,
+ *                      client sends requests to all replica regions and takes the first
+ *                      response.
+ *
+ *   LoadBalance      - Client sends requests to meta replica regions in a round-robin mode,
+ *                      if results from replica regions are stale, next time, client sends requests for
+ *                      these stable locations to the primary meta region. In this mode, scan
+ *                      requests are load balanced across all replica regions.
+ */
+enum MetaReplicaMode {
+  None,
+  HighAvailable,
+  LoadBalance
+}
+
+/**
+ * A Meta replica chooser decides which meta replica to go for scan requests.
+ */
+@InterfaceAudience.Private
+public interface MetaReplicaLoadBalanceReplicaChooser {
+
+  void updateCacheOnError(final HRegionLocation loc, final int fromMetaReplicaId);
+  int chooseReplicaToGo(final TableName tablename, final byte[] row,
+    final RegionLocateType locateType);
+
+
+}
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
@@ -1,0 +1,248 @@
+/**
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.client;
+
+import org.apache.commons.lang3.builder.ToStringBuilder;
+import org.apache.hadoop.hbase.HRegionLocation;
+import org.apache.hadoop.hbase.ScheduledChore;
+import org.apache.hadoop.hbase.Stoppable;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import java.util.Iterator;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.ConcurrentNavigableMap;
+import java.util.concurrent.ConcurrentSkipListMap;
+import java.util.concurrent.ThreadLocalRandom;
+import static org.apache.hadoop.hbase.HConstants.DEFAULT_META_REPLICA_NUM;
+import static org.apache.hadoop.hbase.HConstants.META_REPLICAS_NUM;
+import static org.apache.hadoop.hbase.client.ConnectionUtils.isEmptyStopRow;
+import static org.apache.hadoop.hbase.util.Bytes.BYTES_COMPARATOR;
+import static org.apache.hadoop.hbase.util.ConcurrentMapUtils.computeIfAbsent;
+
+/**
+ * MetaReplicaLoadBalanceReplicaSimpleChooser implements a simple meta replica load balancing
+ * algorithm. It maintains a stale location cache for each table. Whenever client looks up meta,
+ * it first check if the row is the stale location cache, if yes, this means the the location from
+ * meta replica is stale, it will go to the primary meta to look up update-to-date location;
+ * otherwise, it will randomly pick up a meta replica region for meta lookup. When clients receive
+ * RegionNotServedException from region servers, it will add these region locations to the stale
+ * location cache. The stale cache will be cleaned up periodically by a chore.
+ */
+
+/**
+ * StaleLocationCacheEntry is the entry when a stale location is reported by an client.
+ */
+class StaleLocationCacheEntry {
+  // meta replica id where
+  private int metaReplicaId;
+
+  // timestamp in milliseconds
+  private long timestamp;
+
+  private byte[] endKey;
+
+  StaleLocationCacheEntry(final int metaReplicaId, final byte[] endKey) {
+    this.metaReplicaId = metaReplicaId;
+    this.endKey = endKey;
+    timestamp = System.currentTimeMillis();
+  }
+
+  public byte[] getEndKey() {
+    return this.endKey;
+  }
+
+  public int getMetaReplicaId() {
+    return this.metaReplicaId;
+  }
+  public long getTimestamp() {
+    return this.timestamp;
+  }
+
+  @Override
+  public String toString() {
+    return new ToStringBuilder(this)
+      .append("endKey", endKey)
+      .append("metaReplicaId", metaReplicaId)
+      .append("timestamp", timestamp)
+      .toString();
+  }
+}
+
+/**
+ * A simple implementation of MetaReplicaLoadBalanceReplicaChooser.
+ *
+ * It follows a simple algorithm to choose a meta replica to go:
+ *
+ *  1. If there is no stale location entry for rows it looks up, it will randomly
+ *     pick a meta replica region to do lookup.
+ *  2. If the location from meta replica region is stale, client gets RegionNotServedException
+ *     from region server, in this case, it will create StaleLocationCacheEntry in
+ *     MetaReplicaLoadBalanceReplicaSimpleChooser.
+ *  3. When client tries to do meta lookup, it checks StaleLocationCache first for rows it tries to
+ *     lookup, if entry exists, it will go with primary meta region to do lookup; otherwise, it
+ *     will follow step 1.
+ *  4. A chore will periodically run to clean up cache entries in the StaleLocationCache.
+ */
+class MetaReplicaLoadBalanceReplicaSimpleChooser implements MetaReplicaLoadBalanceReplicaChooser {
+  private static final Logger LOG =
+    LoggerFactory.getLogger(MetaReplicaLoadBalanceReplicaSimpleChooser.class);
+  private final long STALE_CACHE_TIMEOUT_IN_MILLISECONDS = 3000; // 3 seconds
+  private final int STALE_CACHE_CLEAN_CHORE_INTERVAL = 1500; // 1.5 seconds
+
+  private final class StaleTableCache {
+    private final ConcurrentNavigableMap<byte[], StaleLocationCacheEntry> cache =
+      new ConcurrentSkipListMap<>(BYTES_COMPARATOR);
+  }
+
+  private final ConcurrentMap<TableName, StaleTableCache> staleCache;
+  private final int numOfMetaReplicas;
+  private final AsyncConnectionImpl conn;
+
+  MetaReplicaLoadBalanceReplicaSimpleChooser(final AsyncConnectionImpl conn) {
+    staleCache = new ConcurrentHashMap<>();
+    this.numOfMetaReplicas = conn.getConfiguration().getInt(
+      META_REPLICAS_NUM, DEFAULT_META_REPLICA_NUM);
+    this.conn = conn;
+    this.conn.getChoreService().scheduleChore(getCacheCleanupChore(this));
+  }
+
+  /**
+   * When a client runs into RegionNotServingException, it will call this method to
+   * update Chooser's internal state.
+   * @param loc the location which causes exception.
+   * @param fromMetaReplicaId the meta replica id where the location comes from.
+   */
+  public void updateCacheOnError(final HRegionLocation loc, final int fromMetaReplicaId) {
+    StaleTableCache tableCache =
+      computeIfAbsent(staleCache, loc.getRegion().getTable(), StaleTableCache::new);
+    byte[] startKey = loc.getRegion().getStartKey();
+    tableCache.cache.putIfAbsent(startKey,
+      new StaleLocationCacheEntry(fromMetaReplicaId, loc.getRegion().getEndKey()));
+    LOG.debug("Add entry to stale cache for table {} with startKey {}, {}",
+      loc.getRegion().getTable(), startKey, loc.getRegion().getEndKey());
+  }
+
+  /**
+   * When it does a meta lookup, it will call this method to find a meta replica to go.
+   * @param tablename  table name it looks up
+   * @param row   key it looks up.
+   * @param locateType locateType, Only BEFORE and CURRENT will be passed in.
+   * @return meta replica id
+   */
+  public int chooseReplicaToGo(final TableName tablename, final byte[] row,
+    final RegionLocateType locateType) {
+    StaleTableCache tableCache = staleCache.get(tablename);
+    int metaReplicaId = 1 + ThreadLocalRandom.current().nextInt(this.numOfMetaReplicas - 1);
+
+    // If there is no entry in StaleCache, pick a random meta replica id.
+    if (tableCache == null) {
+      return metaReplicaId;
+    }
+
+    Map.Entry<byte[], StaleLocationCacheEntry> entry;
+    boolean isEmptyStopRow = isEmptyStopRow(row);
+    // Only BEFORE and CURRENT are passed in.
+    if (locateType == RegionLocateType.BEFORE) {
+      entry = isEmptyStopRow ? tableCache.cache.lastEntry() : tableCache.cache.lowerEntry(row);
+    } else {
+      entry = tableCache.cache.floorEntry(row);
+    }
+
+    if (entry == null) {
+      return metaReplicaId;
+    }
+
+    // Check if the entry times out.
+    if ((System.currentTimeMillis() - entry.getValue().getTimestamp()) >=
+      STALE_CACHE_TIMEOUT_IN_MILLISECONDS) {
+      LOG.debug("Entry for table {} with startKey {}, {} times out", tablename, entry.getKey(),
+        entry);
+      tableCache.cache.remove(entry.getKey());
+      return metaReplicaId;
+    }
+
+    byte[] endKey =  entry.getValue().getEndKey();
+
+    if (isEmptyStopRow(endKey)) {
+      LOG.debug("Lookup {} goes to primary meta", row);
+      return RegionInfo.DEFAULT_REPLICA_ID;
+    }
+
+    if (locateType == RegionLocateType.BEFORE) {
+      if (!isEmptyStopRow && Bytes.compareTo(endKey, row) >= 0) {
+        LOG.debug("Lookup {} goes to primary meta", row);
+        return RegionInfo.DEFAULT_REPLICA_ID;
+      }
+    } else {
+      if (Bytes.compareTo(row, endKey) < 0) {
+        LOG.debug("Lookup {} goes to primary meta", row);
+        return RegionInfo.DEFAULT_REPLICA_ID;
+      }
+    }
+
+    return metaReplicaId;
+  }
+
+  private void cleanupMetaReplicaStaleCache() {
+    long curTimeInMills = System.currentTimeMillis();
+    for (StaleTableCache tableCache : staleCache.values()) {
+      Iterator<Map.Entry<byte[], StaleLocationCacheEntry>> it =
+        tableCache.cache.entrySet().iterator();
+      while (it.hasNext()) {
+        Map.Entry<byte[], StaleLocationCacheEntry> entry = it.next();
+        if (curTimeInMills - entry.getValue().getTimestamp() >=
+          STALE_CACHE_TIMEOUT_IN_MILLISECONDS) {
+          LOG.debug("clean entry {}, {} from stale cache", entry.getKey(), entry.getValue());
+          it.remove();
+        }
+      }
+    }
+  }
+
+  private static Stoppable createDummyStoppable() {
+    return new Stoppable() {
+      private volatile boolean isStopped = false;
+
+      @Override
+      public void stop(String why) {
+        isStopped = true;
+      }
+
+      @Override
+      public boolean isStopped() {
+        return isStopped;
+      }
+    };
+  }
+
+  public ScheduledChore getCacheCleanupChore(
+    final MetaReplicaLoadBalanceReplicaSimpleChooser simpleChooser) {
+    Stoppable stoppable = createDummyStoppable();
+    return new ScheduledChore("CleanupMetaReplicaStaleCache", stoppable,
+      STALE_CACHE_CLEAN_CHORE_INTERVAL) {
+      @Override
+      protected void chore() {
+        simpleChooser.cleanupMetaReplicaStaleCache();
+      }
+    };
+  }
+}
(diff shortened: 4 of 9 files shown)


Apache-HBase:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   1m  8s |  Docker mode activated.  |
||| _ Prechecks _ |
| +1 :green_heart: |  dupname  |   0m  1s |  No case conflicting files found.  |
||| _ HBASE-18070 Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 22s |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   4m  1s |  HBASE-18070 passed  |
| +1 :green_heart: |  checkstyle  |   2m  3s |  HBASE-18070 passed  |
| +1 :green_heart: |  spotbugs  |   3m 47s |  HBASE-18070 passed  |
||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 12s |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   3m 45s |  the patch passed  |
| -0 :warning: |  checkstyle  |   0m 27s |  hbase-client: The patch generated 3 new + 1 unchanged - 0 fixed = 4 total (was 1)  |
| +1 :green_heart: |  hadoopcheck  |  12m 24s |  Patch does not cause any errors with Hadoop 3.1.2 3.2.1.  |
| -1 :x: |  spotbugs  |   1m 12s |  hbase-client generated 1 new + 0 unchanged - 0 fixed = 1 total (was 0)  |
||| _ Other Tests _ |
| +1 :green_heart: |  asflicense  |   0m 33s |  The patch does not generate ASF License warnings.  |
|  |   |  42m 52s |   |


| Reason | Tests |
|-------:|:------|
| FindBugs | module:hbase-client |
|  |  Should org.apache.hadoop.hbase.client.MetaReplicaLoadBalanceReplicaSimpleChooser$StaleTableCache be a _static_ inner class?  At MetaReplicaLoadBalanceReplicaSimpleChooser.java:inner class?  At MetaReplicaLoadBalanceReplicaSimpleChooser.java:[lines 111-112] |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | Client=19.03.13 Server=19.03.13 base: https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-general-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/2570 |
| Optional Tests | dupname asflicense spotbugs hadoopcheck hbaseanti checkstyle |
| uname | Linux b51aff078fd9 4.15.0-112-generic #113-Ubuntu SMP Thu Jul 9 23:41:39 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | HBASE-18070 / 4e47554af8 |
| checkstyle | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-general-check/output/diff-checkstyle-hbase-client.txt |
| spotbugs | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-general-check/output/new-spotbugs-hbase-client.html |
| Max. process+thread count | 84 (vs. ulimit of 30000) |
| modules | C: hbase-common hbase-client hbase-server U: . |
| Console output | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/console |
| versions | git=2.17.1 maven=(cecedd343002696d0abb50b32b541b8a6ba2883f) spotbugs=3.1.12 |


hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+import static org.apache.hadoop.hbase.util.ConcurrentMapUtils.computeIfAbsent;
+
+/**
+ * MetaReplicaLoadBalanceReplicaSimpleChooser implements a simple meta replica load balancing
+ * algorithm. It maintains a stale location cache for each table. Whenever client looks up meta,
+ * it first check if the row is the stale location cache, if yes, this means the the location from
+ * meta replica is stale, it will go to the primary meta to look up update-to-date location;
+ * otherwise, it will randomly pick up a meta replica region for meta lookup. When clients receive
+ * RegionNotServedException from region servers, it will add these region locations to the stale
+ * location cache. The stale cache will be cleaned up periodically by a chore.
```
saintstack, line 48: Is chore heavyweight for a small map like this? Weak references? Or guava cache?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+    private final ConcurrentNavigableMap<byte[], StaleLocationCacheEntry> cache =
+      new ConcurrentSkipListMap<>(BYTES_COMPARATOR);
+  }
+
+  private final ConcurrentMap<TableName, StaleTableCache> staleCache;
+  private final int numOfMetaReplicas;
+  private final AsyncConnectionImpl conn;
+
+  MetaReplicaLoadBalanceReplicaSimpleChooser(final AsyncConnectionImpl conn) {
+    staleCache = new ConcurrentHashMap<>();
```
saintstack, line 121: Declare the data member and assign at same time rather than declare above and assign here? nit.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+    this.numOfMetaReplicas = conn.getConfiguration().getInt(
+      META_REPLICAS_NUM, DEFAULT_META_REPLICA_NUM);
+    this.conn = conn;
+    this.conn.getChoreService().scheduleChore(getCacheCleanupChore(this));
```
saintstack, line 125: What you thinking here? You made the chore on client side generic. You want to make it so different policy implementations can add chore? The chores can do different things dependent on implementation?

Above I wonder if you have to use chore? Is what is needed here too much to do in a guava cache say?

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
```diff
@@ -1128,6 +1128,9 @@
 
   /** Conf key for enabling meta replication */
   public static final String USE_META_REPLICAS = "hbase.meta.replicas.use";
+  public static final String META_REPLICAS_MODE = "hbase.meta.replicas.mode";
```
saintstack, line 1131: Yeah, does it have to be meta exclusive. Can this go somewhere in client package rather than here in the global HConstants? It is a client-only config?

saintstack (commented):
Some high-level feedback @huaxiangsun

Apache-HBase:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   1m  4s |  Docker mode activated.  |
| -0 :warning: |  yetus  |   0m  4s |  Unprocessed flag(s): --brief-report-file --spotbugs-strict-precheck --whitespace-eol-ignore-list --whitespace-tabs-ignore-list --quick-hadoopcheck  |
||| _ Prechecks _ |
||| _ HBASE-18070 Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 23s |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   3m 40s |  HBASE-18070 passed  |
| +1 :green_heart: |  compile  |   1m 44s |  HBASE-18070 passed  |
| +1 :green_heart: |  shadedjars  |   6m 32s |  branch has no errors when building our shaded downstream artifacts.  |
| +1 :green_heart: |  javadoc  |   1m 22s |  HBASE-18070 passed  |
||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 16s |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   3m 28s |  the patch passed  |
| +1 :green_heart: |  compile  |   1m 46s |  the patch passed  |
| +1 :green_heart: |  javac  |   1m 46s |  the patch passed  |
| +1 :green_heart: |  shadedjars  |   6m 39s |  patch has no errors when building our shaded downstream artifacts.  |
| +1 :green_heart: |  javadoc  |   1m 20s |  the patch passed  |
||| _ Other Tests _ |
| +1 :green_heart: |  unit  |   1m 24s |  hbase-common in the patch passed.  |
| +1 :green_heart: |  unit  |   1m  1s |  hbase-client in the patch passed.  |
| -1 :x: |  unit  | 149m 33s |  hbase-server in the patch failed.  |
|  |   | 183m  0s |   |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | Client=19.03.13 Server=19.03.13 base: https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk8-hadoop3-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/2570 |
| uname | Linux 12298d174364 4.15.0-112-generic #113-Ubuntu SMP Thu Jul 9 23:41:39 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | HBASE-18070 / 4e47554af8 |
| Default Java | 1.8.0_232 |
| unit | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk8-hadoop3-check/output/patch-unit-hbase-server.txt |
|  Test Results | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/testReport/ |
| Max. process+thread count | 4557 (vs. ulimit of 30000) |
| modules | C: hbase-common hbase-client hbase-server U: . |
| Console output | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/console |
| versions | git=2.17.1 maven=(cecedd343002696d0abb50b32b541b8a6ba2883f) |


Apache-HBase:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   0m 44s |  Docker mode activated.  |
| -0 :warning: |  yetus  |   0m  4s |  Unprocessed flag(s): --brief-report-file --spotbugs-strict-precheck --whitespace-eol-ignore-list --whitespace-tabs-ignore-list --quick-hadoopcheck  |
||| _ Prechecks _ |
||| _ HBASE-18070 Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 26s |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   5m 12s |  HBASE-18070 passed  |
| +1 :green_heart: |  compile  |   2m 24s |  HBASE-18070 passed  |
| +1 :green_heart: |  shadedjars  |   8m 11s |  branch has no errors when building our shaded downstream artifacts.  |
| -0 :warning: |  javadoc  |   0m 32s |  hbase-client in HBASE-18070 failed.  |
| -0 :warning: |  javadoc  |   0m 19s |  hbase-common in HBASE-18070 failed.  |
| -0 :warning: |  javadoc  |   0m 47s |  hbase-server in HBASE-18070 failed.  |
||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 17s |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   4m 52s |  the patch passed  |
| +1 :green_heart: |  compile  |   2m 25s |  the patch passed  |
| +1 :green_heart: |  javac  |   2m 25s |  the patch passed  |
| +1 :green_heart: |  shadedjars  |   8m 14s |  patch has no errors when building our shaded downstream artifacts.  |
| -0 :warning: |  javadoc  |   0m 18s |  hbase-common in the patch failed.  |
| -0 :warning: |  javadoc  |   0m 30s |  hbase-client in the patch failed.  |
| -0 :warning: |  javadoc  |   0m 47s |  hbase-server in the patch failed.  |
||| _ Other Tests _ |
| +1 :green_heart: |  unit  |   1m 49s |  hbase-common in the patch passed.  |
| +1 :green_heart: |  unit  |   1m 17s |  hbase-client in the patch passed.  |
| -1 :x: |  unit  | 144m 52s |  hbase-server in the patch failed.  |
|  |   | 186m 44s |   |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | Client=19.03.13 Server=19.03.13 base: https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/2570 |
| uname | Linux 2c97bc4d46aa 4.15.0-58-generic #64-Ubuntu SMP Tue Aug 6 11:12:41 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | HBASE-18070 / 4e47554af8 |
| Default Java | 2020-01-14 |
| javadoc | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/branch-javadoc-hbase-client.txt |
| javadoc | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/branch-javadoc-hbase-common.txt |
| javadoc | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/branch-javadoc-hbase-server.txt |
| javadoc | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/patch-javadoc-hbase-common.txt |
| javadoc | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/patch-javadoc-hbase-client.txt |
| javadoc | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/patch-javadoc-hbase-server.txt |
| unit | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/artifact/yetus-jdk11-hadoop3-check/output/patch-unit-hbase-server.txt |
|  Test Results | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/testReport/ |
| Max. process+thread count | 4210 (vs. ulimit of 30000) |
| modules | C: hbase-common hbase-client hbase-server U: . |
| Console output | https://ci-hadoop.apache.org/job/HBase/job/HBase-PreCommit-GitHub-PR/job/PR-2570/1/console |
| versions | git=2.17.1 maven=(cecedd343002696d0abb50b32b541b8a6ba2883f) |


hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
```diff
@@ -0,0 +1,53 @@
...
+  None,
+  HighAvailable,
+  LoadBalance
+}
+
+/**
+ * A Meta replica chooser decides which meta replica to go for scan requests.
+ */
+@InterfaceAudience.Private
+public interface MetaReplicaLoadBalanceReplicaChooser {
```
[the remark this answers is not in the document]
huaxiangsun: Thanks, selector is a better fit, had a hard time to pick up a right name.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
```diff
@@ -0,0 +1,53 @@
...
+ *                      response.
+ *
+ *   LoadBalance      - Client sends requests to meta replica regions in a round-robin mode,
+ *                      if results from replica regions are stale, next time, client sends requests for
+ *                      these stable locations to the primary meta region. In this mode, scan
+ *                      requests are load balanced across all replica regions.
+ */
+enum MetaReplicaMode {
+  None,
+  HighAvailable,
```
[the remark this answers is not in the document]
huaxiangsun: Yeah, META_REPLICAS_MODE_LOADBALANCE_REPILCA_CHOOSER(/SELECTOR) is supposed to be the config change for a different implementation.

If it is changed to ReplicaMode, it sounds like this is supported for user tables as well. A ReplicaMode only applies to some special system tables?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
```diff
@@ -0,0 +1,53 @@
...
+  LoadBalance
+}
+
+/**
+ * A Meta replica chooser decides which meta replica to go for scan requests.
+ */
+@InterfaceAudience.Private
+public interface MetaReplicaLoadBalanceReplicaChooser {
+
+  void updateCacheOnError(final HRegionLocation loc, final int fromMetaReplicaId);
```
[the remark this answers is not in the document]
huaxiangsun: Is the cache here  TableCache or StaleCache maintained in ReplicaSelector? There is no need to pass it in.
Will take care of the rest comments.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: Will take care it.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: I need something to clean up the stale cache periodically. Let me do some research on guava cache.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: That is not what I thought. If guava cache is a better fit, let me make the change.

Apache9:
On a high level design, we used to have a `hbase.meta.replicas.use` config to control whether to make use meta replicas at client side. Do we want to deprecate this config and let our new configs rule here? Just asking.

For me, I think the scope of `hbase.meta.replicas.use` is too wide, as it will impact all the client operations on meta(not sure if we have hacked all the places but at least it is designed to). After reviewing the PR, I think our approach here is only for the locator related logic right? This is also what I expect. In general, I think there are 3 ways we access meta table at client side:

1. Locator related logic. This is the most critical path at client side and also makes the most pressure on meta table.
2. Admin related logic. We have delegated most of the operaations through master but there are still some places we will access meta directly. But admin operation is expected to be low frequent so I do not think we need to deal with it here.
3. Users access meta table directly by their own. This is controlled by user written code so I do not think we need to deal with it either, users should take care of it by their own.

I think only 1 is what we really care here, so I suggest that, we just narrow the scope of the newly introduced configs to be locator only(maybe by adding 'locator' keyword in the config name), and consider it first before `hbase.meta.replicas.use` in the locator related logic. So users do not need to set `hbase.meta.replicas.use` to true to enable this feature, just select the LB mode. If the new config is disabled(set to None maybe), then we honor the old `hbase.meta.replicas.use` config.

In general, I think the abstraction and trick here are both good. Setting the replica id directly on Query is a straight forward way to archive our goal here, and the chooser or selector is also a good abstraction. The only concern is how to implement the 'fallback to primary' logic as we need to pass down from the rpc retrying caller of the actual exception type, anyway, this can be done later.

And I suggest we just make this PR against master branch, it is only client side change and whether we have implement the meta region replication should not impact the code. Why I suggest this is that, the code for master and branch-2 will be different, as on branch-2, the sync client has its own logic to do region locating, which is not well constructed I believe(we expose a bunch of locating methods in ClusterConnection interface and use it everywhere). So if we want to include this feature in 2.4.0, we'd better make this PR against master, and also backport it to branch-2 ASAP.

Thanks.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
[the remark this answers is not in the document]
Apache9: I do not think this is suitable for user tables yet. At least for the LoadBalance mode, where we explicitly mention the 'stale' data will impact our policy. For meta table I think whether the data is 'stale' can be detected by our own, but for user tables, only users know how to determine whether the data is stale, which means we need to make the interface be able to get information from users on whether the previous data is stale and I only want to go to primary replica this time. This requires a good design on interface as it will be IA.Public, so I think keep it as meta only and IA.Private is more suitable for now.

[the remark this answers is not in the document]
Apache9: But better to put this enum to a separated file?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+ * meta replica is stale, it will go to the primary meta to look up update-to-date location;
+ * otherwise, it will randomly pick up a meta replica region for meta lookup. When clients receive
+ * RegionNotServedException from region servers, it will add these region locations to the stale
+ * location cache. The stale cache will be cleaned up periodically by a chore.
+ */
+
+/**
+ * StaleLocationCacheEntry is the entry when a stale location is reported by an client.
+ */
+class StaleLocationCacheEntry {
```
Apache9, line 54: Give it a separated file or make it an inner class. Having two classes in a single file is not a good practise.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+ *     pick a meta replica region to do lookup.
+ *  2. If the location from meta replica region is stale, client gets RegionNotServedException
+ *     from region server, in this case, it will create StaleLocationCacheEntry in
+ *     MetaReplicaLoadBalanceReplicaSimpleChooser.
+ *  3. When client tries to do meta lookup, it checks StaleLocationCache first for rows it tries to
+ *     lookup, if entry exists, it will go with primary meta region to do lookup; otherwise, it
+ *     will follow step 1.
+ *  4. A chore will periodically run to clean up cache entries in the StaleLocationCache.
+ */
+class MetaReplicaLoadBalanceReplicaSimpleChooser implements MetaReplicaLoadBalanceReplicaChooser {
```
Apache9, line 105: Maybe first we just need to cache the replica id which contains the stale data? For example, we get location from replica 1 and it is stale, but it does not mean that the location from replica 2 will be stale? If we use the algorithm here, we will go to replica 0(the primary replica) instead of replica 1.

And consider the whole client locator architecture, where have a meta cache in the upper layer, I would suggest that, we only provide two selection ways here: the first is random(no primay), and if you got a stale data, then use the second way, only go to primary. Maybe we could add a parameter on the choose method to indicate whether we only want to go to the primary replica.

Apache9 (changes requested):

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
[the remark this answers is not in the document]
huaxiangsun: Yeah, will do.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: Our original plan is to come up with a simple algorithm first.
1. Select a meta replica region first for lookup, if the data is stale, then go to primary.

Later, maybe we can experiment some smarter way about how to select the next meta replica in case of error.
Here,  state needed to make decision is contained insider this selector class. If I understand correctly, to have a flag in choose method indicates that top layer needs to keep some state for this purpose. Trying to avoid that so there is no side effect if this mode is not enabled.

huaxiangsun: For a normal case, > 99% of region locations from meta replica will be up to date. In extreme cases such as region server crash, it will depends on how fast replication catches up. Will try to add more tests to find out.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
```diff
@@ -130,8 +133,12 @@ public AsyncConnectionImpl(Configuration conf, ConnectionRegistry registry, Stri
       SocketAddress localAddress, User user) {
     this.conf = conf;
     this.user = user;
-    if (user.isLoginFromKeytab()) {
-      spawnRenewalChore(user.getUGI());
+
+    if (user.isLoginFromKeytab() || isMetaReplicaLBMode()) {
```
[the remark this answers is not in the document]
huaxiangsun: Yeah, trying to convert it to guava cache, which seems light weight.

huaxiangsun:
> On a high level design, we used to have a `hbase.meta.replicas.use` config to control whether to make use meta replicas at client side. Do we want to deprecate this config and let our new configs rule here? Just asking.
> 

For me, original thought is that `hbase.meta.replicas.use` is an existing config, so it still needs to be honored. A new "mode" config is added to support LoadBalance(new) and HighlyAvailable(the existing mode). If we want to deprecate this config, then, it makes sense.

> For me, I think the scope of `hbase.meta.replicas.use` is too wide, as it will impact all the client operations on meta(not sure if we have hacked all the places but at least it is designed to). After reviewing the PR, I think our approach here is only for the locator related logic right? This is also what I expect. In general, I think there are 3 ways we access meta table at client side:
> 
> 1. Locator related logic. This is the most critical path at client side and also makes the most pressure on meta table.
> 2. Admin related logic. We have delegated most of the operaations through master but there are still some places we will access meta directly. But admin operation is expected to be low frequent so I do not think we need to deal with it here.
> 3. Users access meta table directly by their own. This is controlled by user written code so I do not think we need to deal with it either, users should take care of it by their own.

Agreed.

> 
> I think only 1 is what we really care here, so I suggest that, we just narrow the scope of the newly introduced configs to be locator only(maybe by adding 'locator' keyword in the config name), and consider it first before `hbase.meta.replicas.use` in the locator related logic. So users do not need to set `hbase.meta.replicas.use` to true to enable this feature, just select the LB mode. If the new config is disabled(set to None maybe), then we honor the old `hbase.meta.replicas.use` config.

Ok.

> 
> In general, I think the abstraction and trick here are both good. Setting the replica id directly on Query is a straight forward way to archive our goal here, and the chooser or selector is also a good abstraction. The only concern is how to implement the 'fallback to primary' logic as we need to pass down from the rpc retrying caller of the actual exception type, anyway, this can be done later.

Do you think "fallback to primary" logic needs to be passed down from rpc retrying caller? Then it needs to be aware of this feature and needs to maintain some state. Was trying to avoid it. Yeah, this part can be optimized later.

> 
> And I suggest we just make this PR against master branch, it is only client side change and whether we have implement the meta region replication should not impact the code. Why I suggest this is that, the code for master and branch-2 will be different, as on branch-2, the sync client has its own logic to do region locating, which is not well constructed I believe(we expose a bunch of locating methods in ClusterConnection interface and use it everywhere). So if we want to include this feature in 2.4.0, we'd better make this PR against master, and also backport it to branch-2 ASAP.

Yeah, the client side change is quite independent. It will work with today's meta replica, only nit is that there will be more stale data. 
I will put a new patch against master. All implementation will be renamed as Catalog instead of meta, as it can be applied to similar location lookup service. The config will remain same as here it applies to meta only.

Thanks for the feedbacks, Duo.

huaxiangsun:
Forgot to mention that patch against master, the only nit is testing cases as there is no meta replication support in master, will check the tests and see how much test can be ported.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+   * @param loc the location which causes exception.
+   * @param fromMetaReplicaId the meta replica id where the location comes from.
+   */
+  public void updateCacheOnError(final HRegionLocation loc, final int fromMetaReplicaId) {
+    StaleTableCache tableCache =
+      computeIfAbsent(staleCache, loc.getRegion().getTable(), StaleTableCache::new);
+    byte[] startKey = loc.getRegion().getStartKey();
+    tableCache.cache.putIfAbsent(startKey,
+      new StaleLocationCacheEntry(fromMetaReplicaId, loc.getRegion().getEndKey()));
+    LOG.debug("Add entry to stale cache for table {} with startKey {}, {}",
```
ndimiduk, line 140: nit: name the "endKey" in your log message.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
```diff
@@ -0,0 +1,53 @@
...
+  int chooseReplicaToGo(final TableName tablename, final byte[] row,
```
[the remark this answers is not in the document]
ndimiduk: Oh, I understand. this is the TableName of the data table from which we're locating a region. Maybe there's a better parameter name? Likewise, `row` should be `startRow`? Or no, the client doesn't know the region split points or the start row, only the data row they're seeking.

Yeah, I think this method name and it's parameters could be improved for readability. Something like `int selectReplicaIdFor(TableName targetTable, byte[] targetStartRow)`. The correct name gets even more confusing/important when we have split meta in the works...

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+    this.conn.getChoreService().scheduleChore(getCacheCleanupChore(this));
+  }
+
+  /**
+   * When a client runs into RegionNotServingException, it will call this method to
+   * update Chooser's internal state.
+   * @param loc the location which causes exception.
+   * @param fromMetaReplicaId the meta replica id where the location comes from.
+   */
+  public void updateCacheOnError(final HRegionLocation loc, final int fromMetaReplicaId) {
```
ndimiduk, line 134: Please use `@Overrides` annotations.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+  MetaReplicaLoadBalanceReplicaSimpleChooser(final AsyncConnectionImpl conn) {
+    staleCache = new ConcurrentHashMap<>();
+    this.numOfMetaReplicas = conn.getConfiguration().getInt(
+      META_REPLICAS_NUM, DEFAULT_META_REPLICA_NUM);
+    this.conn = conn;
+    this.conn.getChoreService().scheduleChore(getCacheCleanupChore(this));
+  }
+
+  /**
+   * When a client runs into RegionNotServingException, it will call this method to
```
ndimiduk, line 129: Javadoc should be on the `interface`.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
```diff
@@ -0,0 +1,248 @@
...
+    } else {
+      entry = tableCache.cache.floorEntry(row);
+    }
+
+    if (entry == null) {
+      return metaReplicaId;
+    }
+
+    // Check if the entry times out.
+    if ((System.currentTimeMillis() - entry.getValue().getTimestamp()) >=
```
ndimiduk, line 175: EnvironmentEdge please.

Same nit: move the calculation out to a local variable to keep the if-condition concise.

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
ndimiduk: What do you propose @saintstack ? `hbase.catalogue.replicas.mode` ?

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
```diff
@@ -1128,6 +1128,9 @@
...
+  public static final String META_REPLICAS_MODE_LOADBALANCE_REPILCA_CHOOSER =
+    "hbase.meta.replicas.mode.loadbalance.replica.chooser";
```
ndimiduk, line 1133: I see you're forced to work with existing configuration names, but i wish these were all `hbase.meta_replica.`

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestAsyncNonMetaRegionLocatorWithMetaReplicaLoadBalance.java
```diff
@@ -0,0 +1,84 @@
...
+import org.apache.hadoop.hbase.testclassification.ClientTests;
+import org.apache.hadoop.hbase.testclassification.MediumTests;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.hadoop.hbase.util.ServerRegionReplicaUtil;
+import org.junit.BeforeClass;
+import org.junit.ClassRule;
+import org.junit.experimental.categories.Category;
+
+@Category({ MediumTests.class, ClientTests.class })
+public class TestAsyncNonMetaRegionLocatorWithMetaReplicaLoadBalance
```
ndimiduk, line 34: oh no, please not test inheritance :'(

Can this be done as a `Parameterized` test instead?

hbase-server/src/test/java/org/apache/hadoop/hbase/replication/regionserver/TestMetaRegionReplicaReplicationEndpoint.java
```diff
@@ -63,17 +68,20 @@
 /**
  * Tests RegionReplicaReplicationEndpoint class for hbase:meta by setting up region replicas and
  * verifying async wal replication replays the edits to the secondary region in various scenarios.
+ *
  * @see TestRegionReplicaReplicationEndpoint
  */
 @Category({LargeTests.class})
```
ndimiduk, line 74: Not yours, but I think this can be be added to the `MasterTests` category. Or `RegionServerTests`.. I'm not even clear on which one this is at this point. Maybe need a category for `CatalogueTests`.

hbase-server/src/test/java/org/apache/hadoop/hbase/replication/regionserver/TestMetaRegionReplicaReplicationEndpoint.java
```diff
@@ -338,4 +476,134 @@ static boolean contains(List<Result> contains, List<Cell> cells) throws IOExcept
     }
     return !containsScanner.advance() && matches >= 1 && count >= matches && count == cells.size();
   }
+
+  private void doNGets(final Table table, final byte[][] keys) throws Exception {
```
ndimiduk, line 480: nit: for helper methods who's purpose is to make assertions, I like to prefix their names with `assert`. So, `assertDoNGets`, `assertPrimaryNoChangeReplicaIncrease`, &c.

hbase-server/src/test/java/org/apache/hadoop/hbase/replication/regionserver/TestMetaRegionReplicaReplicationEndpoint.java
```diff
@@ -176,18 +184,136 @@ private boolean isMetaRegionReplicaReplicationSource(HRegionServer hrs) {
    */
   @Test
   public void testHBaseMetaReplicates() throws Exception {
-    try (Table table = HTU.createTable(TableName.valueOf(this.name.getMethodName() + "_0"),
-      HConstants.CATALOG_FAMILY,
-        Arrays.copyOfRange(HBaseTestingUtility.KEYS, 1, HBaseTestingUtility.KEYS.length)))  {
-      verifyReplication(TableName.META_TABLE_NAME, NB_SERVERS, getMetaCells(table.getName()));
+    try (Table table = HTU
+      .createTable(TableName.valueOf(this.name.getMethodName() + "_0"), HConstants.CATALOG_FAMILY,
```
ndimiduk, line 188: FYI, there's a `TableNameTestRule` for generating the `TableName` from the running method. I guess it doesn't help where you're adding a suffix though...

hbase-server/src/test/java/org/apache/hadoop/hbase/replication/regionserver/TestMetaRegionReplicaReplicationEndpoint.java
```diff
@@ -338,4 +476,134 @@ static boolean contains(List<Result> contains, List<Cell> cells) throws IOExcept
...
+      // There are read requests increase for primary meta replica.
+      // For rest of meta replicas, there is no change as regionMove will tell the new location
+      primaryIncreaseReplicaNoChange(readReqsForMetaReplicasAfterGet,
+        readReqsForMetaReplicasAfterMove);
+      // Move region again.
+      HTU.moveRegionAndWait(userRegion.getRegionInfo(), srcRs.getServerName());
+
+      // Wait until moveRegion cache timeout.
+      while (destRs.getMovedRegion(userRegion.getRegionInfo().getEncodedName()) != null) {
+        Thread.sleep(1000);
```
ndimiduk, line 588: please use `waitFor` with a maximum time to wait.

ndimiduk (changes requested):
My comments are mostly on style and code structure. Only a couple questions on the implementation details.

Apache9:
> Do you think "fallback to primary" logic needs to be passed down from rpc retrying caller? Then it needs to be aware of this feature and needs to maintain some state. Was trying to avoid it. Yeah, this part can be optimized later.

If you want to make a decision based on whether the previous returned location is stale, then you need to pass something down from the rpc retrying caller to the selector, as only in rpc retrying caller we can get the exception we want. Either by adding a parameter as I proposed above to not go to secondary replicas, or as your current solution, adding a stale cache in the selector and you need to pass down the exception and the related location.
My concern here is that, the current 'simple' algorithm may not work well for some cases as my mentioned above. I suggest that we remove this part of code from the patch here(just do random selecting, without other consideration) to get this patch in quickly, and then open another issue to discuss the suitable algorithm, or maybe multiple algorithms.

Thanks.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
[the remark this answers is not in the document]
huaxiangsun: My bad, Guava cache does not work in this case, we need a tablecache like implementation which can look up locations. Will consider chore service on-demand.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
```diff
@@ -435,6 +475,14 @@ private void locateInMeta(TableName tableName, LocateRequest req) {
       .setReadType(ReadType.PREAD);
     if (useMetaReplicas) {
       scan.setConsistency(Consistency.TIMELINE);
+
+      // If it is meta replica LoadBalance mode, pick up a meta replica to scan. It will try to
+      // pick up meta replica region to scan, in certain error cases, it may pick primary meta
+      // region to scan.
+      if (isMetaReplicaLBMode()) {
+        scan.setReplicaId(this.metaReplicaChooser.chooseReplicaToGo(tableName,
```
huaxiangsun, line 483: Mark here. Just realized that there is a logical bug issue. When chooseReplicaToGo() returns primary region to go, in this case, it should unset TIMELINE. Otherwise, it will fall back to HighlyAvailable mode, which is not what we want here.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
```diff
@@ -577,6 +627,15 @@ private void removeLocationFromCache(HRegionLocation loc) {
       if (!canUpdateOnError(loc, oldLoc)) {
         return;
       }
+      // Tell metaReplicaChooser that the location is stale. It will create a stale entry
+      // with timestamp internally. Next time the client looks up the same location,
+      // it will pick a different meta replica region. For the current implementation,
```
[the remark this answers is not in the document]
huaxiangsun: Will do.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
huaxiangsun: addressed in the next patch.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaChooser.java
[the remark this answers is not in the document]
huaxiangsun: Agreed. I am going to rename it "Catalog" instead of "Meta", to reflect the fact that it can be used to location lookup service. As for user tables, if LoadBalance mode needs to be supported, current interface already supports it. It is up to application to support this mode.

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
huaxiangsun: "hbase.catalog.replicas.mode"? Or we want to include "meta" here so it can only enable meta table. In the future, let's say root table needs to be supported, so another config knob will be added. By this way, configs are separated for both for current "meta" and future system tables.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: Got it, will make it an inner class.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: Will do.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncConnectionImpl.java
```diff
@@ -392,6 +406,16 @@ public Connection toConnection() {
     return c;
   }
 
+  private boolean isMetaReplicaLBMode() {
+    if (conf.getBoolean(USE_META_REPLICAS, DEFAULT_USE_META_REPLICAS)) {
+      String metaReplicaMode = conf.get(META_REPLICAS_MODE, "");
+      if (MetaReplicaMode.LoadBalance.toString().equals(metaReplicaMode)) {
```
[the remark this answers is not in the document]
huaxiangsun: Good catch! I think it should be case-insensitive.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
```diff
@@ -198,6 +208,36 @@ private boolean tryComplete(LocateRequest req, CompletableFuture<RegionLocations
       conn.getConfiguration().getInt(LOCATE_PREFETCH_LIMIT, DEFAULT_LOCATE_PREFETCH_LIMIT);
     this.useMetaReplicas =
       conn.getConfiguration().getBoolean(USE_META_REPLICAS, DEFAULT_USE_META_REPLICAS);
+
+    if (this.useMetaReplicas) {
+      if (conn.getConfiguration()
+        .get(META_REPLICAS_MODE, "").equals(MetaReplicaMode.LoadBalance.toString())) {
```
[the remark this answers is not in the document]
huaxiangsun: Will take care of it.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java
```diff
@@ -198,6 +208,36 @@ private boolean tryComplete(LocateRequest req, CompletableFuture<RegionLocations
...
+        this.metaReplicaMode = MetaReplicaMode.LoadBalance;
+        int numOfMetaReplicas = conn.getConfiguration().getInt(
+          META_REPLICAS_NUM, DEFAULT_META_REPLICA_NUM);
+        if (numOfMetaReplicas <= 1) {
+          LOG.error("Configured to support meta replica load balance mode,"
+            + " but there is no meta replica configured");
+          this.metaReplicaChooser = null;
+        } else {
+          String metaReplicaChooserImpl = conn.getConfiguration().get(
```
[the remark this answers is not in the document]
huaxiangsun: Ok.

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestAsyncNonMetaRegionLocatorWithMetaReplicaLoadBalance.java
huaxiangsun: Let me check it. Trying to make sure that with Parameterized test, minicluster is going to be recreated (once I was there, but forgot the details).

hbase-server/src/test/java/org/apache/hadoop/hbase/replication/regionserver/TestMetaRegionReplicaReplicationEndpoint.java
huaxiangsun: will add "RegionServerTests" tag.

huaxiangsun:
> > Do you think "fallback to primary" logic needs to be passed down from rpc retrying caller? Then it needs to be aware of this feature and needs to maintain some state. Was trying to avoid it. Yeah, this part can be optimized later.
> 
> If you want to make a decision based on whether the previous returned location is stale, then you need to pass something down from the rpc retrying caller to the selector, as only in rpc retrying caller we can get the exception we want. Either by adding a parameter as I proposed above to not go to secondary replicas, or as your current solution, adding a stale cache in the selector and you need to pass down the exception and the related location.
> My concern here is that, the current 'simple' algorithm may not work well for some cases as my mentioned above. I suggest that we remove this part of code from the patch here(just do random selecting, without other consideration) to get this patch in quickly, and then open another issue to discuss the suitable algorithm, or maybe multiple algorithms.
> 

Thank Duo.

I think I am still not 100% clear here. Let me try to explain my understanding of what you thought before moving forward.
1. My current understanding of rpc retry caller and AsyncRegionLocator.

If a location is stale, rpc retry caller will detect it and pass down exception to AsyncRegionLocator(AsyncNonMetaRegionLocator), AsyncNonMetaRegionLocator updates/cleans up table cache accordingly.

2. In the next retry, if there is no table cache entry for the key, it will call locateInMeta() to fetch location from meta.

In the above flow, rpc retry caller layer is the one detecting stale location, it is up to AsyncRegionLocator layer to handle all the logic.

Similarly, meta selector interface tries to do similar and build the state it needs to make decision inside itself.

As for different algorithm, i.e, the case you described above, a concrete example, lets there there are 5 meta replicas, if the location from meta replica 1 stale, it may try other meta replica. We can have a different implementation of selector,
AdvancedSelector, which can implement such algorithm based on the state maintained in selector. I.e, the staleCache entry has info to indicate fromMetaReplicaId. 

Thought that the selector interface is designed for this purpose, maybe I miss something here.

The initial implementation, I kept fromMetaReplica in tableCache entry. When the patch was out for review, I dropped this change, as the current simple algorithm really does not need this info. The select interface keeps fromMetaReplicaId for future enhancement so when it is needed, at least the interface does not need to be changed.

I can drop fromMetaReplicaId in the current simple implementation as it is not needed anyway.

Maybe you were referring to add a flag in locateInMeta() method? If the flag says going with primary only, it will bypass all these logic and goes only to primary region so it will give upper layer more control.

Please share more thoughts, thanks.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: Did some research, and there is not a clean solution as here it needs a TableCache like cache. Will let it as it is for now, and if there is better one found later, we can change it.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/MetaReplicaLoadBalanceReplicaSimpleChooser.java
huaxiangsun: Connection already has a chore for Kerberos keyRelogin, we are going to reuse it here. Mostly, I think a different policy will need to maintain a cache, and chore is needed for different implementation.

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
huaxiangsun: Thought it about again, I think this config knob needs to be specific to meta for now. In the future, if it is going to be turned on for another table, different knob needs to be provided as basically they are for different purpose.

huaxiangsun:
Will post a new patch based on master, the major change in the master branch is that meta replica # is part of meta table, test and selector initialization needs to be changed accordingly.

huaxiangsun:
This request is replaced by a new request against master branch, will do a backport once the master patch is reviewed and merged.

PR #3630, opened 2021-08-26
HBASE-26147: Add a dry run mode to the balancer, where moves are calculated but not actually executed

bbeaudreault:
This is the master port for https://github.com/apache/hbase/pull/3536

I also addressed documentation feedback from https://github.com/apache/hbase/pull/3536. This port was not super straightforward because of rsgroup being moved into `hbase-server`.  The move was not a simply file move, but code moved within various classes. Additionally, HBaseAdmin.java was removed in master in favor of AdminOverAsyncAdmin (and some other minor admin changes).

I've read everything over, but it may be worth giving a quick look for correctness

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceRequest.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceRequest.java
@@ -19,13 +19,11 @@
 package org.apache.hadoop.hbase.client;
 
 import org.apache.yetus.audience.InterfaceAudience;
-import org.apache.yetus.audience.InterfaceStability;
 
 /**
  * Encapsulates options for executing a run of the Balancer.
  */
 @InterfaceAudience.Public
-@InterfaceStability.Evolving
 public final class BalanceRequest {
   private static final BalanceRequest DEFAULT = BalanceRequest.newBuilder().build();
 
@@ -33,7 +31,6 @@ public final class BalanceRequest {
    * Builder for constructing a {@link BalanceRequest}
    */
   @InterfaceAudience.Public
-  @InterfaceStability.Evolving
   public final static class Builder {
     private boolean dryRun = false;
     private boolean ignoreRegionsInTransition = false;
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceResponse.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceResponse.java
@@ -19,20 +19,17 @@
 package org.apache.hadoop.hbase.client;
 
 import org.apache.yetus.audience.InterfaceAudience;
-import org.apache.yetus.audience.InterfaceStability;
 
 /**
  * Response returned from a balancer invocation
  */
 @InterfaceAudience.Public
-@InterfaceStability.Evolving
 public final class BalanceResponse {
 
   /**
    * Builds a {@link BalanceResponse} for returning results of a balance invocation to callers
    */
   @InterfaceAudience.Public
-  @InterfaceStability.Evolving
   public final static class Builder {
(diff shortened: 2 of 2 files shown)


Apache9 (approved):

hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceResponse.java
```diff
@@ -0,0 +1,126 @@
...
+/**
+ * Response returned from a balancer invocation
+ */
+@InterfaceAudience.Public
+public final class BalanceResponse {
+
+  /**
+   * Builds a {@link BalanceResponse} for returning results of a balance invocation to callers
+   */
+  @InterfaceAudience.Public
```
joshelser, line 32: Thinking about it, we probably don't want the Builder class here to be Public. Users shouldn't ever be building this `BalanceResponse` on their own, just using the `BalanceResponse`.

joshelser (approved):
LGTM. Sorry I forgot that Public+Evolving was a no-go (you're right that we should have something in-between -- LimitedPrivate+Evolving might be the way to go but it's hard when we're talking about `Admin` and other long-standing API).

Only one final thought is that `BalanceResponse$Builder` probably shouldn't be Public. Use of the object is certainly Public, but how that object is created does not need to be public.

I think I'd rather see this land rather than another round of reviews. The last precommit round looks good to me (just some flakiness in tests). Let me try to do that tonight.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceResponse.java
bbeaudreault: I created https://issues.apache.org/jira/browse/HBASE-26240 for this. I'll submit a small PR once these existing PRs are merged.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/BalanceRequest.java
```diff
@@ -0,0 +1,117 @@
...
+package org.apache.hadoop.hbase.client;
+
+import org.apache.yetus.audience.InterfaceAudience;
+import org.apache.yetus.audience.InterfaceStability;
+
+/**
+ * Encapsulates options for executing a run of the Balancer.
+ */
+@InterfaceAudience.Public
+@InterfaceStability.Evolving
```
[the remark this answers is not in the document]
ndimiduk: @Apache9 I'm also surprised by your interpretation of the interaction between `IA.Public` and `IS.Evolving`. The javadoc on `InterfaceStability` says
> All classes that are annotated with {@link Public} or {@link LimitedPrivate} must have InterfaceStability annotation.

Indeed, `InterfaceStability` itself is annotated with `IA.Public` and `IS.Evolving`.

ndimiduk:
I left a comment on one of Duo's line comments, but I'll repeat it here.

I'm also surprised by this interpretation of the interaction between `IA.Public` and `IS.Evolving`. The javadoc on `InterfaceStability` says
> All classes that are annotated with {@link Public} or {@link LimitedPrivate} must have InterfaceStability annotation.

Indeed, `InterfaceStability` itself is annotated with `IA.Public` and `IS.Evolving`.

@joshelser @Apache9 do we document the HBase project's interpretation of these annotations somewhere in our developer guide?

joshelser:
> do we document the HBase project's interpretation of these annotations somewhere in our developer guide?

Yeah, we have this: https://hbase.apache.org/book.html#hbase.client.api.surface

I think Bryan definitely found a gap in that we don't have a good way to represent user-facing but not yet stable.

joshelser:
Best as I can tell, the test failures were flakes which passed on re-run.

PR #3633, opened 2021-08-26
HBASE-26228 updateRSGroupConfig operation should be authorized by AccessController

pankaj72981:


pankaj72981:
> And maybe this should be declared as an incompatible issue?

It shouldn't be an incompatible change, ACL check is skipped only in master branch. 

In branch-2 ACL is validated twice (we should correct it)
1. MasterRpcServices#execMasterService
         https://github.com/apache/hbase/blob/3b1482ef6016b1df7dfdf2d87a048242ceec5a12/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterRpcServices.java#L891
2. RSGroupAdminEndpoint (currently updateRSGroupConfig doesn't call checkPermission)
        https://github.com/apache/hbase/blob/3b1482ef6016b1df7dfdf2d87a048242ceec5a12/hbase-rsgroup/src/main/java/org/apache/hadoop/hbase/rsgroup/RSGroupAdminEndpoint.java#L259

PR #3774, opened 2021-10-20
HBASE-26386 Refactor StoreFileTracker implementations to expose the s…

Apache9:
…et method


Apache9:
It turns out that we still need to use StoreFileTrackerBase in MigrationStoreFileTracker so the refactoring is just a small one...

PTAL @wchevreuil 

Thanks.

PR #4175, opened 2022-03-07
HBASE-25783 Close PENDING_OPEN regions make regions FAILED_CLOSE

sunhelly:


Apache9:
Branch-1 will be EOL soon so I tend to not include this sensitive changes.

PR #4232, opened 2022-03-16
HBASE-26849 Fix NPE caused by WAL Compression and Replication

thangTang:
Details in [HBASE-26849](https://issues.apache.org/jira/browse/HBASE-26849)

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/WALEntryStream.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/WALEntryStream.java
@@ -249,7 +249,7 @@ public class WALEntryStream implements Iterator<Entry>, Closeable, Iterable<Entr
             + ", which is too far away from reported file length " + stat.getLen()
             + ". Restarting WAL reading (see HBASE-15983 for details). " + getCurrentPathStat());
         setPosition(0);
-        resetReader();
+        openReader(currentPath, true);
         metrics.incrRestartedWALReading();
         metrics.incrRepeatedFileBytes(currentPosition);
         return false;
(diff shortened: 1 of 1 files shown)


Apache-HBase:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   0m  0s |  Docker mode activated.  |
| -1 :x: |  docker  |   0m  0s |  Docker command '/usr/bin/docker' is too old (1.41 < API v 1.27.0).  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-4232/1/console |
| versions | git=2.17.1 |


hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/WALEntryStream.java
```diff
@@ -249,7 +249,7 @@ private boolean checkAllBytesParsed() throws IOException {
             + ", which is too far away from reported file length " + stat.getLen()
             + ". Restarting WAL reading (see HBASE-15983 for details). " + getCurrentPathStat());
         setPosition(0);
-        resetReader();
+        openReader(currentPath, true);
```
sunhelly, line 252: How about add a resetCompressionContext() method in the implementation of Reader.reset()? That means when the reader is reset, the compressionContext needs also be reset.
And then the resetReader() can be kept everywhere?

thangTang: @sunhelly Pls see [WALEntryStream#resetReader](https://github.com/apache/hbase/blob/branch-1/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/regionserver/WALEntryStream.java#L391)
There will call _seek_ method, which means the resetReader also might reposition the cursor somewhere, so i think we cannot just clear the compressionContext.
As a temporary/staged fix, I think creating the reader directly is the least invasive... What do you think?

PR #4520, opened 2022-06-10
HBASE-27097 SimpleRpcServer is broken

apurtell:
Apply https://issues.apache.org/jira/secure/attachment/13044874/MultiByteBuff.patch from Lijin Bin on JIRA.


apurtell:
Use #4521 instead.

PR #111, opened 2019-04-05
HBASE-22170 Add an import order template under the dev-support directory

Apache9:

PR #5518, opened 2023-11-13
HBASE-28197 The configs for using meta replica can conflict

sunhelly:
We should use "hbase.meta.replicas.use" to control the switch of reading meta replicas. But the codes before this PR reading from meta replicas only when "hbase.locator.meta.replicas.mode" is not NONE, it supassed the switch controller.
The UT can reproduce the problem.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestAsyncTableUseMetaReplicas.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestAsyncTableUseMetaReplicas.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hbase.client;
 
+import static org.apache.hadoop.hbase.client.RegionLocator.LOCATOR_META_REPLICAS_MODE;
 import static org.junit.Assert.assertArrayEquals;
 
 import java.io.IOException;
@@ -24,6 +25,7 @@ import java.util.Optional;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.CatalogReplicaMode;
 import org.apache.hadoop.hbase.HBaseClassTestRule;
 import org.apache.hadoop.hbase.HBaseTestingUtil;
 import org.apache.hadoop.hbase.HConstants;
@@ -65,6 +67,7 @@ public class TestAsyncTableUseMetaReplicas {
   private static byte[] VALUE = Bytes.toBytes("Value");
 
   private static volatile boolean FAIL_PRIMARY_SCAN = false;
+  private static volatile boolean USE_META_REPLICA = false;
 
   public static final class FailPrimaryMetaScanCp implements RegionObserver, RegionCoprocessor {
 
@@ -77,6 +80,10 @@ public class TestAsyncTableUseMetaReplicas {
     public void preScannerOpen(ObserverContext<RegionCoprocessorEnvironment> c, Scan scan)
       throws IOException {
       RegionInfo region = c.getEnvironment().getRegionInfo();
+      if (!USE_META_REPLICA && TableName.isMetaTableName(region.getTable())) {
+        assert region.getReplicaId() == RegionReplicaUtil.DEFAULT_REPLICA_ID
+          : "should read default meta region!";
+      }
       if (
         FAIL_PRIMARY_SCAN && TableName.isMetaTableName(region.getTable())
           && region.getReplicaId() == RegionReplicaUtil.DEFAULT_REPLICA_ID
@@ -117,11 +124,11 @@ public class TestAsyncTableUseMetaReplicas {
     FAIL_PRIMARY_SCAN = false;
   }
 
-  private void testRead(boolean useMetaReplicas)
-    throws IOException, InterruptedException, ExecutionException {
+  private void testRead() throws IOException, InterruptedException, ExecutionException {
     FAIL_PRIMARY_SCAN = true;
     Configuration conf = new Configuration(UTIL.getConfiguration());
-    conf.setBoolean(HConstants.USE_META_REPLICAS, useMetaReplicas);
+    conf.setBoolean(HConstants.USE_META_REPLICAS, USE_META_REPLICA);
+    conf.set(LOCATOR_META_REPLICAS_MODE, CatalogReplicaMode.LOAD_BALANCE.toString());
     conf.setLong(HConstants.HBASE_CLIENT_META_REPLICA_SCAN_TIMEOUT, TimeUnit.SECONDS.toMicros(1));
     try (AsyncConnection conn = ConnectionFactory.createAsyncConnection(conf).get()) {
       Result result = FutureUtils.get(conn.getTableBuilder(TABLE_NAME)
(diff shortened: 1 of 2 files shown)


Apache-HBase:
:broken_heart: **-1 overall**






| +0 :ok: |  reexec  |   1m 59s |  Docker mode activated.  |
| -0 :warning: |  yetus  |   0m  3s |  Unprocessed flag(s): --brief-report-file --spotbugs-strict-precheck --whitespace-eol-ignore-list --whitespace-tabs-ignore-list --quick-hadoopcheck  |
||| _ Prechecks _ |
| +0 :ok: |  mvndep  |   0m 12s |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   2m 29s |  master passed  |
| +1 :green_heart: |  compile  |   0m 52s |  master passed  |
| +1 :green_heart: |  shadedjars  |   5m  9s |  branch has no errors when building our shaded downstream artifacts.  |
| +1 :green_heart: |  javadoc  |   0m 34s |  master passed  |
||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 11s |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   2m 23s |  the patch passed  |
| +1 :green_heart: |  compile  |   0m 52s |  the patch passed  |
| +1 :green_heart: |  javac  |   0m 52s |  the patch passed  |
| +1 :green_heart: |  shadedjars  |   5m  5s |  patch has no errors when building our shaded downstream artifacts.  |
| +1 :green_heart: |  javadoc  |   0m 34s |  the patch passed  |
||| _ Other Tests _ |
| +1 :green_heart: |  unit  |   1m 20s |  hbase-client in the patch passed.  |
| -1 :x: |  unit  | 221m 40s |  hbase-server in the patch failed.  |
|  |   | 247m 40s |   |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.43 ServerAPI=1.43 base: https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-5518/1/artifact/yetus-jdk8-hadoop3-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/5518 |
| uname | Linux 080cf93794fd 5.4.0-1103-aws #111~18.04.1-Ubuntu SMP Tue May 23 20:04:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | master / e806350bd0 |
| Default Java | Temurin-1.8.0_352-b08 |
| unit | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-5518/1/artifact/yetus-jdk8-hadoop3-check/output/patch-unit-hbase-server.txt |
|  Test Results | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-5518/1/testReport/ |
| Max. process+thread count | 4667 (vs. ulimit of 30000) |
| modules | C: hbase-client hbase-server U: . |
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-5518/1/console |
| versions | git=2.34.1 maven=3.8.6 |


hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestAsyncTableUseMetaReplicas.java
```diff
@@ -77,6 +80,10 @@ public Optional<RegionObserver> getRegionObserver() {
     public void preScannerOpen(ObserverContext<RegionCoprocessorEnvironment> c, Scan scan)
       throws IOException {
       RegionInfo region = c.getEnvironment().getRegionInfo();
+      if (!USE_META_REPLICA && TableName.isMetaTableName(region.getTable())) {
+        assert region.getReplicaId() == RegionReplicaUtil.DEFAULT_REPLICA_ID
```
wchevreuil, line 84: Please use assertEquals.

wchevreuil (commented):
So the aim here is to enforce the "LOAD_BALANCE" reading mode only if "USE_META_REPLICAS" is explicitly set?

Apache9:
Have you check the ref guide? For me, I think setting the replica mode to something other than NONE indicates that the user wants to use meta replica?

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestAsyncTableUseMetaReplicas.java
sunhelly: OK. Thanks.

sunhelly:
> So the aim here is to enforce the "LOAD_BALANCE" reading mode only if "USE_META_REPLICAS" is explicitly set?

sunhelly:
> Have you check the ref guide? For me, I think setting the replica mode to something other than NONE indicates that the user wants to use meta replica?

Thanks, @Apache9 . The ref guide doesn't bound the relation ship between the two configs. I think, when we enabling the meta replica, we will set both the two configs under normal circumstances. But when encountering problems caused by  meta replica, disable it throught the configs is a quick recover method. The key problem here is that we should disable meta replica by only one config("hbase.meta.replicas.use") or both the relevant configs. I prefer the first way, but I'm neutral on both the ways.

sunhelly:
> So the aim here is to enforce the "LOAD_BALANCE" reading mode only if "USE_META_REPLICAS" is explicitly set?

Yes, @wchevreuil . My initial thought was when "USE_META_REPLICAS" is false, we should not read the value of "LOAD_BALANCE". But this maybe miss some original design idea. Let's wait for a consensus. Thanks.

Apache9:
> > Have you check the ref guide? For me, I think setting the replica mode to something other than NONE indicates that the user wants to use meta replica?
> 
> Thanks, @Apache9 . The ref guide doesn't bound the relation ship between the two configs. I think, when we enabling the meta replica, we will set both the two configs under normal circumstances. But when encountering problems caused by meta replica, disable it throught the configs is a quick recover method. The key problem here is that we should disable meta replica by only one config("hbase.meta.replicas.use") or both the relevant configs. I prefer the first way, but I'm neutral on both the ways.

Maybe at client side we just do not care about `hbase.meta.replicas.use`? The `hbase.meta.replicas.use` config is only for server side. So maybe a more suitable way here is that, at server side, like when initialize a region server or master, when `hbase.meta.replicas.use` is set to false, we force set the replica mode to NONE as it is useless to use other replica modes?

sunhelly:
> > > Have you check the ref guide? For me, I think setting the replica mode to something other than NONE indicates that the user wants to use meta replica?
> > 
> > 
> > Thanks, @Apache9 . The ref guide doesn't bound the relation ship between the two configs. I think, when we enabling the meta replica, we will set both the two configs under normal circumstances. But when encountering problems caused by meta replica, disable it throught the configs is a quick recover method. The key problem here is that we should disable meta replica by only one config("hbase.meta.replicas.use") or both the relevant configs. I prefer the first way, but I'm neutral on both the ways.
> 
> Maybe at client side we just do not care about `hbase.meta.replicas.use`? The `hbase.meta.replicas.use` config is only for server side. So maybe a more suitable way here is that, at server side, like when initialize a region server or master, when `hbase.meta.replicas.use` is set to false, we force set the replica mode to NONE as it is useless to use other replica modes?

Thanks， @Apache9 . I think the config `hbase.meta.replicas.use` appears before `hbase.locator.meta.replicas.mode`, which is brought by [HBASE-25126](https://issues.apache.org/jira/browse/HBASE-25126) to introduce more meta access mode for clients. And at the server side, in the initialization of regionserver and master, the `hbase.meta.replicas.use` actively force the setting to false, see line https://github.com/apache/hbase/blob/29bfc610d0433f720a34bc47aadca1433bbb1882/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java#L515 and https://github.com/apache/hbase/blob/29bfc610d0433f720a34bc47aadca1433bbb1882/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java#L529. As a result, I think the replica mode should be force set to NONE at both server side and client side. What do you think?

Apache9:
@huaxiangsun WDYT?

Apache9 (approved):

PR #5546, opened 2023-11-29
HBASE-28222 Leak in ExportSnapshot during verifySnapshot on S3A

bbeaudreault:
Encapsulates the setting of `fs.impl.disable.cache` to just the specific FileSystems we plan to create and manage closing. This prevents accidental explosion of uncached FileSystems coming from static utils like CommonFsUtils.getRootDir, which are used in snapshot verification. When the FileSystem is S3A, this results in a leak of S3AImplementation objects that will lead to an OOM.


bbeaudreault:
I may reopen this, but actually I want to investigate a different approach.

PR #5708, opened 2024-02-26
Port HBASE-27230 to branch-2

jhungund:
HBASE-27230 RegionServer should be aborted when WAL.sync throws Timeout


jhungund:
Abandoned

PR #5956, opened 2024-05-31
HBASE-27915: Fixed hadolint check in nightly build.

subrat-mishra:
Tested `hadolint <Dockerfile>` and verified that it doesn't throw any error.


subrat-mishra:
Used a wrong HBASE ticket.

PR #6386, opened 2024-10-23
HBASE-28906 Run nightly tests with multiple Hadoop 3 versions (addend…

stoty:
…um: workaround for MAPREDUCE-7492)


stoty:
Merging without waiting for tests, because the precommit tests don' exercise this.
This was present for my test job, but I forgot to copy it to the final commit.

Apache9 (approved):

PR #6672, opened 2025-02-10
Major compaction is disabled, but the webui shows that major compaction has been triggered

wdzaslzy:
https://issues.apache.org/jira/browse/HBASE-29119

When major compaction is disabled, if all hfiles in a single region participate in compaction, they will be automatically marked as major compaction. I think this is wrong and contrary to the expected major compaction. In addition, according to the context, I found that as long as all hfiles participate in compaction, they are automatically deleted. I think this also needs to be adjusted.


Apache9:
Branch-2.4 is already EOL, please open a PR against master branch first.

Thanks.

wdzaslzy:
Thanks, submitted

PR #6815, opened 2025-03-18
HBASE-29192 Provide installation method that supports running hbase-it

stoty:
alternative PR with profiles


stoty:
wrong local branch

PR #7350, opened 2025-09-30
HBASE-29625 Reduce the search space during file selection in ExploringCompactionPolicy

Huginn-kio:
HBASE-29625 Reduce the search space during file selection in ExploringCompactionPolicy


Huginn-kio:
The ExploringCompactionPolicy prioritizes selecting sets with a larger number of files. Therefore, once the first best candidate is chosen, subsequent iterations only need to consider contiguous sub-intervals with an equal or greater number of files.

PR #7363, opened 2025-10-07
HBASE-29645 Reduce synchronization in AsyncBufferedMutatorImpl

apurtell:
This patch modifies `AsyncBufferedMutatorImpl` class to improve its performance under concurrent usage. 

While `AsyncTable#batch()` is largely asynchronous in nature, it can exhibit blocking behavior during its preparation phase, for instance, while looking up region locations. In the original implementation of `AsyncBufferedMutatorImpl`, calls to `AsyncTable#batch()` occur within a synchronized block, potentially causing severe contention and stalling other threads trying to buffer their mutations. The original implementation relied on coarse grained synchronized blocks for multi-threading safety, so when one thread triggered a buffer flush (either because the buffer was full or a periodic timer fired), all other threads attempting to add mutations via the mutate method would be blocked until the table.batch() call completed, which could take a surprisingly long time.

The new implementation replaces the broad synchronized blocks with a `ReentrantLock`. This lock is acquired only for the brief period needed to safely copy the current batch of mutations and futures into local variables and swap in a new internal buffer. Immediately after this quick operation, the lock is released. The `batch()` call is then executed outside of the locked section. This allows other threads to continue adding new mutations concurrently while the flushing of the previous batch proceeds independently. The client has already opted in to asynchronous and potentially interleaved commit of the mutations submitted to `AsyncBufferedMutator`, by definition. The minimization of critical section scope minimizes thread contention and significantly boosts throughput under load. Other related profiler driven efficiency changes are also included, such as elimination of stream api and array resizing hotspots identified by the profiler.

To validate the performance improvement of these changes, a JMH benchmark, `AsyncBufferedMutatorBenchmark` was created to measure the performance of the mutate method under various conditions. It focuses specifically on the overhead and concurrency management of `AsyncBufferedMutatorImpl` itself, not the underlying network communication. To achieve this, it uses the `Mockito` framework to create a mock `AsyncTable` that instantly returns completed futures, isolating the mutator's buffering logic for measurement. It runs tests with 1, 10, and 100 threads to simulate no, medium, and high levels of concurrency. It uses a low value (100) for `maxMutations` to force frequent flushes based on the number of mutations, and a very high value (100,000) to ensure flushes are rare in that measurement case. The benchmark measures the average time per operation in microseconds, where a lower score indicates better performance and higher throughput.

With a single thread and no contention the performance of both implementations is nearly identical. The minor variations are negligible and show that the new locking mechanism does not introduce any performance regression in the non-concurrent case. For example, with a 10MB buffer and high maxMutations, the NEW implementation scored 0.167 us/op while the OLD scored 0.169 us/op, a statistically insignificant difference. When the test is run with 10 threads, a noticeable gap appears. In the scenario designed to cause frequent flushes (maxMutations = 100), the NEW implementation is approximately 12 times faster than the OLD one (14.250 us/op for NEW vs. 172.463 us/op for OLD). This is because the OLD implementation forces threads to wait while flushes occur, and flushes incur a synthetic thread sleep of 1ms to simulate occasional unexpected blocking behavior in `AsyncTable#batch()`, whereas the NEW implementation allows them to proceed without contention. The most significant results come from the 100-thread tests, which simulate high contention. In the frequent flush scenario (maxMutations = 100) the NEW implementation is 114 times faster in the synthetic benchmark scenario (16.123 us/op for NEW vs. 1847.567 us/op for OLD). Note that blocking IO observed in a real client for e.g. region location lookups can produce a much more significant impact. With the OLD code, 100 threads are constantly competing for a lock that is held for a long duration, leading to a contention storm. The NEW code's reduced locking scope almost entirely eliminates this bottleneck.

OS: Apple Silicon (aarch64) M1 Max / 64 GB

JVM: openjdk version "17.0.11" 2024-04-16 LTS  / OpenJDK 64-Bit Server VM Zulu17.50+19-CA (build 17.0.11+9-LTS, mixed mode, sharing)

| Threads | Max Mutations | OLD Implementation (µs/op) | NEW Implementation (µs/op) | Performance Gain (OLD / NEW) |
| :--- | :--- | :--- | :--- | :--- |
| 1 | 100 | 14.091 | 16.313 | 0.86x (comparable) |
| 1 | 100,000 | 0.169 | 0.167 | 1.01x (comparable) |
| 10 | 100 | 172.463 | 14.250 | 12.10x |
| 10 | 100,000 | 2.465 | 1.072 | 2.30x |
| 100 | 100 | 1847.567 | 16.123 | **114.59x** |
| 100 | 100,000 | 24.125 | 12.796 | 1.89x |

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
@@ -196,7 +196,7 @@ class AsyncBufferedMutatorImpl implements AsyncBufferedMutator {
     try {
       if (this.mutations.isEmpty() && periodicFlushTimeoutNs > 0) {
         periodicFlushTask = periodicalFlushTimer.newTimeout(timeout -> {
-          boolean shouldFlush = false;
+          Batch flushBatch = null;
           lock.lock();
           try {
             // confirm that we are still valid, if there is already an internalFlush call before us,
@@ -205,13 +205,13 @@ class AsyncBufferedMutatorImpl implements AsyncBufferedMutator {
             // are equal.
             if (timeout == periodicFlushTask) {
               periodicFlushTask = null;
-              shouldFlush = true;
+              flushBatch = drainBatch(); // Drains under lock
             }
           } finally {
             lock.unlock();
           }
-          if (shouldFlush) {
-            internalFlush();
+          if (flushBatch != null) {
+            sendBatch(flushBatch); // Sends outside of lock
           }
         }, periodicFlushTimeoutNs, TimeUnit.NANOSECONDS);
       }


Apache-HBase:






| Vote | Subsystem | Runtime |  Logfile | Comment |
|:----:|----------:|--------:|:--------:|:-------:|
| +0 :ok: |  reexec  |   1m  3s |  |  Docker mode activated.  |
| -0 :warning: |  yetus  |   0m  3s |  |  Unprocessed flag(s): --brief-report-file --spotbugs-strict-precheck --author-ignore-list --blanks-eol-ignore-file --blanks-tabs-ignore-file --quick-hadoopcheck  |
|||| _ Prechecks _ |
|||| _ master Compile Tests _ |
| +1 :green_heart: |  mvninstall  |   4m  6s |  |  master passed  |
| +1 :green_heart: |  compile  |   0m 24s |  |  master passed  |
| +1 :green_heart: |  javadoc  |   0m 22s |  |  master passed  |
| +1 :green_heart: |  shadedjars  |   6m 26s |  |  branch has no errors when building our shaded downstream artifacts.  |
|||| _ Patch Compile Tests _ |
| +1 :green_heart: |  mvninstall  |   3m 30s |  |  the patch passed  |
| +1 :green_heart: |  compile  |   0m 24s |  |  the patch passed  |
| +1 :green_heart: |  javac  |   0m 24s |  |  the patch passed  |
| +1 :green_heart: |  javadoc  |   0m 18s |  |  the patch passed  |
| +1 :green_heart: |  shadedjars  |   6m 24s |  |  patch has no errors when building our shaded downstream artifacts.  |
|||| _ Other Tests _ |
| +1 :green_heart: |  unit  |   1m 39s |  |  hbase-client in the patch passed.  |
|  |   |  26m 19s |  |  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.43 ServerAPI=1.43 base: https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/1/artifact/yetus-jdk17-hadoop3-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/7363 |
| Optional Tests | javac javadoc unit compile shadedjars |
| uname | Linux b8b7b4c3d8d8 5.4.0-1103-aws #111~18.04.1-Ubuntu SMP Tue May 23 20:04:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | master / d7aa32532ad43d1f888960f2e85dfa7599957572 |
| Default Java | Eclipse Adoptium-17.0.11+9 |
|  Test Results | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/1/testReport/ |
| Max. process+thread count | 292 (vs. ulimit of 30000) |
| modules | C: hbase-client U: hbase-client |
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/1/console |
| versions | git=2.34.1 maven=3.9.8 |
| Powered by | Apache Yetus 0.15.0 https://yetus.apache.org |


Apache-HBase:
:broken_heart: **-1 overall**






| Vote | Subsystem | Runtime |  Logfile | Comment |
|:----:|----------:|--------:|:--------:|:-------:|
| +0 :ok: |  reexec  |   0m 28s |  |  Docker mode activated.  |
|||| _ Prechecks _ |
| +1 :green_heart: |  dupname  |   0m  0s |  |  No case conflicting files found.  |
| +0 :ok: |  codespell  |   0m  0s |  |  codespell was not available.  |
| +0 :ok: |  detsecrets  |   0m  0s |  |  detect-secrets was not available.  |
| +1 :green_heart: |  @author  |   0m  0s |  |  The patch does not contain any @author tags.  |
| +1 :green_heart: |  hbaseanti  |   0m  0s |  |  Patch does not have any anti-patterns.  |
|||| _ master Compile Tests _ |
| +1 :green_heart: |  mvninstall  |   3m 46s |  |  master passed  |
| +1 :green_heart: |  compile  |   0m 55s |  |  master passed  |
| +1 :green_heart: |  checkstyle  |   0m 18s |  |  master passed  |
| +1 :green_heart: |  spotbugs  |   0m 48s |  |  master passed  |
| +1 :green_heart: |  spotless  |   0m 51s |  |  branch has no errors when running spotless:check.  |
|||| _ Patch Compile Tests _ |
| +1 :green_heart: |  mvninstall  |   3m 15s |  |  the patch passed  |
| +1 :green_heart: |  compile  |   0m 48s |  |  the patch passed  |
| +1 :green_heart: |  javac  |   0m 48s |  |  the patch passed  |
| +1 :green_heart: |  blanks  |   0m  0s |  |  The patch has no blanks issues.  |
| +1 :green_heart: |  checkstyle  |   0m 15s |  |  the patch passed  |
| +1 :green_heart: |  spotbugs  |   0m 49s |  |  the patch passed  |
| +1 :green_heart: |  hadoopcheck  |  12m 17s |  |  Patch does not cause any errors with Hadoop 3.3.6 3.4.1.  |
| -1 :x: |  spotless  |   0m 18s |  |  patch has 28 errors when running spotless:check, run spotless:apply to fix.  |
|||| _ Other Tests _ |
| +1 :green_heart: |  asflicense  |   0m 10s |  |  The patch does not generate ASF License warnings.  |
|  |   |  32m 51s |  |  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.43 ServerAPI=1.43 base: https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/1/artifact/yetus-general-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/7363 |
| Optional Tests | dupname asflicense javac spotbugs checkstyle codespell detsecrets compile hadoopcheck hbaseanti spotless |
| uname | Linux 751ec38fcaba 5.4.0-1103-aws #111~18.04.1-Ubuntu SMP Tue May 23 20:04:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | master / d7aa32532ad43d1f888960f2e85dfa7599957572 |
| Default Java | Eclipse Adoptium-17.0.11+9 |
| spotless | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/1/artifact/yetus-general-check/output/patch-spotless.txt |
| Max. process+thread count | 83 (vs. ulimit of 30000) |
| modules | C: hbase-client U: hbase-client |
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/1/console |
| versions | git=2.34.1 maven=3.9.8 spotbugs=4.7.3 |
| Powered by | Apache Yetus 0.15.0 https://yetus.apache.org |


hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -118,59 +156,94 @@ protected void internalFlush() {
...
-        return futures;
-      }
+    boolean needFlush = false;
+    FlushType flushType = FlushType.SIZE;
+    lock.lock();
+    try {
       if (this.mutations.isEmpty() && periodicFlushTimeoutNs > 0) {
         periodicFlushTask = periodicalFlushTimer.newTimeout(timeout -> {
+          boolean shouldFlush = false;
           synchronized (AsyncBufferedMutatorImpl.this) {
```
Apache9: This should also be changed to lock.lock()?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -118,59 +156,94 @@ protected void internalFlush() {
...
+              shouldFlush = true;
             }
           }
+          if (shouldFlush) {
+            internalFlush(FlushType.PERIODIC);
+          }
         }, periodicFlushTimeoutNs, TimeUnit.NANOSECONDS);
       }
+      // Preallocate to avoid potentially multiple resizes during addAll if we can.
+      if (this.mutations instanceof ArrayList && this.futures instanceof ArrayList) {
```
Apache9: These two fields are private fields, so just declare them as ArrayList?

apurtell:
> In general, all rpc request like locating a region should be an asynchronous call, do you have more details on what makes the table.batch call blocking for a long time?

One pattern we see is the caller threads to mutator.mutate() are blocked by long call chains through netty doing IO for region location lookup.

The theory as to why is CompletableFuture has two modes of operation... there is synchronous chaining (thenApply, thenAccept, whenComplete) where the code executes in the same thread that completes the future, while only asynchronous chaining (thenApplyAsync, thenAcceptAsync, whenCompleteAsync) guarantees execution of the lambdas in a separate thread context. HBase async client code may chain by thenX() not thenXAsync(). If some upstream code like region (re)location completes a future synchronously the downstream pipeline executes immediately in the current thread context and blocks the caller. 

The change to ABM guarantees it won't impact callers to mutate() if this happens, although the fact this happens suggests elsewhere in the async client changes from thenX() to thenXAsync() are warranted, as a follow up issue.

apurtell:
f4d8007 addresses spotbugs findings and implements review feedback.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -118,59 +158,93 @@ protected void internalFlush() {
...
+            lock.unlock();
+          }
+          if (shouldFlush) {
+            internalFlush(FlushType.PERIODIC);
           }
         }, periodicFlushTimeoutNs, TimeUnit.NANOSECONDS);
       }
+      // Preallocate to avoid potentially multiple resizes during addAll
+      this.mutations.ensureCapacity(this.mutations.size() + mutations.size());
+      this.futures.ensureCapacity(this.futures.size() + futures.size());
```
d-c-manning: If they are `ArrayList`, do we need to call `ensureCapacity` or do the internals of `addAll` in the jdk already do that? https://github.com/openjdk/jdk8u/blob/5cffbcb0344f2cf16682a09519894ba705182241/jdk/src/share/classes/java/util/ArrayList.java#L582-L585

apurtell:
> The implementation looks good, but we still need to consideration more about the correctness.

@Apache9 What would you suggest here?

I can add more test coverage. Something like this seems straightforward to implement and will not increase the running time of the test suite too much:

Step 1: Generate a test data set of configurable size, default 100K rows. 
Step 2: Use ABM to commit the test data set.
Step 3: Scan the test table and validate every row of the test data set has been committed. 

This will provide test coverage we don't have now, but it does not validate correctness beyond the happy path.

Apache-HBase:






| Vote | Subsystem | Runtime |  Logfile | Comment |
|:----:|----------:|--------:|:--------:|:-------:|
| +0 :ok: |  reexec  |   0m 31s |  |  Docker mode activated.  |
|||| _ Prechecks _ |
| +1 :green_heart: |  dupname  |   0m  0s |  |  No case conflicting files found.  |
| +0 :ok: |  codespell  |   0m  0s |  |  codespell was not available.  |
| +0 :ok: |  detsecrets  |   0m  0s |  |  detect-secrets was not available.  |
| +1 :green_heart: |  @author  |   0m  0s |  |  The patch does not contain any @author tags.  |
| +1 :green_heart: |  hbaseanti  |   0m  0s |  |  Patch does not have any anti-patterns.  |
|||| _ master Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 23s |  |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   3m 58s |  |  master passed  |
| +1 :green_heart: |  compile  |   4m 16s |  |  master passed  |
| +1 :green_heart: |  checkstyle  |   0m 54s |  |  master passed  |
| +1 :green_heart: |  spotbugs  |   2m 20s |  |  master passed  |
| +1 :green_heart: |  spotless  |   0m 53s |  |  branch has no errors when running spotless:check.  |
|||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 13s |  |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   3m  6s |  |  the patch passed  |
| +1 :green_heart: |  compile  |   4m  9s |  |  the patch passed  |
| +1 :green_heart: |  javac  |   4m  9s |  |  the patch passed  |
| +1 :green_heart: |  blanks  |   0m  0s |  |  The patch has no blanks issues.  |
| +1 :green_heart: |  checkstyle  |   0m 52s |  |  the patch passed  |
| +1 :green_heart: |  spotbugs  |   2m 28s |  |  the patch passed  |
| +1 :green_heart: |  hadoopcheck  |  12m 18s |  |  Patch does not cause any errors with Hadoop 3.3.6 3.4.1.  |
| +1 :green_heart: |  spotless  |   0m 45s |  |  patch has no errors when running spotless:check.  |
|||| _ Other Tests _ |
| +1 :green_heart: |  asflicense  |   0m 18s |  |  The patch does not generate ASF License warnings.  |
|  |   |  45m 21s |  |  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.43 ServerAPI=1.43 base: https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/2/artifact/yetus-general-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/7363 |
| Optional Tests | dupname asflicense javac spotbugs checkstyle codespell detsecrets compile hadoopcheck hbaseanti spotless |
| uname | Linux 2a88c42772a2 5.4.0-1103-aws #111~18.04.1-Ubuntu SMP Tue May 23 20:04:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | master / f4d800767265404a1f88fcfe0fdd55a68708bc6f |
| Default Java | Eclipse Adoptium-17.0.11+9 |
| Max. process+thread count | 85 (vs. ulimit of 30000) |
| modules | C: hbase-client hbase-server U: . |
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/2/console |
| versions | git=2.34.1 maven=3.9.8 spotbugs=4.7.3 |
| Powered by | Apache Yetus 0.15.0 https://yetus.apache.org |


Apache-HBase:






| Vote | Subsystem | Runtime |  Logfile | Comment |
|:----:|----------:|--------:|:--------:|:-------:|
| +0 :ok: |  reexec  |   0m 28s |  |  Docker mode activated.  |
| -0 :warning: |  yetus  |   0m  3s |  |  Unprocessed flag(s): --brief-report-file --spotbugs-strict-precheck --author-ignore-list --blanks-eol-ignore-file --blanks-tabs-ignore-file --quick-hadoopcheck  |
|||| _ Prechecks _ |
|||| _ master Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 23s |  |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   3m 50s |  |  master passed  |
| +1 :green_heart: |  compile  |   1m 17s |  |  master passed  |
| +1 :green_heart: |  javadoc  |   0m 46s |  |  master passed  |
| +1 :green_heart: |  shadedjars  |   6m  9s |  |  branch has no errors when building our shaded downstream artifacts.  |
|||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 14s |  |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   3m  0s |  |  the patch passed  |
| +1 :green_heart: |  compile  |   1m 18s |  |  the patch passed  |
| +1 :green_heart: |  javac  |   1m 18s |  |  the patch passed  |
| +1 :green_heart: |  javadoc  |   0m 44s |  |  the patch passed  |
| +1 :green_heart: |  shadedjars  |   6m  2s |  |  patch has no errors when building our shaded downstream artifacts.  |
|||| _ Other Tests _ |
| +1 :green_heart: |  unit  |   1m 30s |  |  hbase-client in the patch passed.  |
| +1 :green_heart: |  unit  | 209m 40s |  |  hbase-server in the patch passed.  |
|  |   | 240m 29s |  |  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.43 ServerAPI=1.43 base: https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/2/artifact/yetus-jdk17-hadoop3-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/7363 |
| Optional Tests | javac javadoc unit compile shadedjars |
| uname | Linux dbbcb10acb8f 5.4.0-1103-aws #111~18.04.1-Ubuntu SMP Tue May 23 20:04:10 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | master / f4d800767265404a1f88fcfe0fdd55a68708bc6f |
| Default Java | Eclipse Adoptium-17.0.11+9 |
|  Test Results | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/2/testReport/ |
| Max. process+thread count | 3558 (vs. ulimit of 30000) |
| modules | C: hbase-client hbase-server U: . |
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/2/console |
| versions | git=2.34.1 maven=3.9.8 |
| Powered by | Apache Yetus 0.15.0 https://yetus.apache.org |


Apache9:
> > In general, all rpc request like locating a region should be an asynchronous call, do you have more details on what makes the table.batch call blocking for a long time?
> 
> I assume we're not waiting for any RPC responses on the `internalFlush` thread, but with large enough buffers and large enough numbers of mutations and high enough concurrency on incoming mutations, it seems even `AsyncBatchRpcRetryingCaller#groupAndSend` can take long enough (milliseconds?) to decrease overall throughput.
> 
> If the region location is in the cache, then the future completes synchronously in `AsyncNonMetaRegionLocator#getRegionLocationsInternal`, which allows more work under the `synchronized` block, which prevents further mutations from being accepted.
> 
> https://github.com/apache/hbase/blob/d0b94780f509156c66ea9d297a89e73940257eb1/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncNonMetaRegionLocator.java#L504-L513
> 
> When the number of mutations could be >100,000 and number of region locations could be >10,000, and most of those locations are in the cache, `groupAndSend` of those Multi RPCs yields a non-trivial amount of work.

So maybe we should check depth of the stack trace? In netty there are some tricks around this area, if the future is complete synchronously all the time and makes a very deep call stack trace, it will force schedule an asynchronous task to prevent stack overflow and also reduce the blocking execution time.

Apache9:
> > The implementation looks good, but we still need to consideration more about the correctness.
> 
> @Apache9 What would you suggest here?
> 
> I can add more test coverage. Something like this seems straightforward to implement and will not increase the running time of the test suite too much:
> 
> Step 1: Generate a test data set of configurable size, default 100K rows. Step 2: Use ABM to commit the test data set. Step 3: Scan the test table and validate every row of the test data set has been committed.
> 
> This will provide test coverage we don't have now, but it does not validate correctness beyond the happy path.

We need to manually consider the implementation carefully, and after merging this PR, we should run ITBLL several rounds before cutting the next release.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -46,6 +45,8 @@ class AsyncBufferedMutatorImpl implements AsyncBufferedMutator {
 
   private static final Logger LOG = LoggerFactory.getLogger(AsyncBufferedMutatorImpl.class);
 
+  private final int INITIAL_CAPACITY = 100;
```
Apache9: static

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -90,19 +108,41 @@ public Configuration getConfiguration() {
...
-      periodicFlushTask.cancel();
-      periodicFlushTask = null;
-    }
-    List<Mutation> toSend = this.mutations;
-    if (toSend.isEmpty()) {
-      return;
+    internalFlush(FlushType.MANUAL);
+  }
+
+  protected void internalFlush(FlushType trigger) {
```
Apache9: Maybe a better choice is to inline this method to the caller, and make the lock protect the check and replace logic, i.e, if we think we should flush, then swap the mutations and futures to local variables and recreate them, and set a local bool variable may be called `shouldSend` to true, and check this bool outside the lock protection and send the mutate request out.

In this way we do not need the double check and flush type check too, and the logic will be more easy to understand. The only problem is the test which overrides internalFlush method. We could try to find other ways to implement it.

apurtell: > The only problem is the test which overrides internalFlush method. We could try to find other ways to implement it.

Do you have a suggestion? It is your test I think :-) 

Otherwise I would like to keep the test as is, and consider this an improvement for the future.

apurtell: Let me see about simplifying this. I don't think we absolutely need 'FlushType' and the double checking.

apurtell:
> So maybe we should check depth of the stack trace? In netty there are some tricks around this area, if the future is complete synchronously all the time and makes a very deep call stack trace, it will force schedule an asynchronous task to prevent stack overflow and also reduce the blocking execution time.

The potential depth of the stack is out of our control. We are providing a client to an application that is developed by an independent team.
So the move of internalFlush and the potentially synchronous downstream completions it may invoke is definitely the right move here.

apurtell:
Rebased, with additional review feedback implemented.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
apurtell: The calls to ensureCapacity were recommended (I forget exactly which resource I was looking at), but they do seem redundant.

Apache9:
> > So maybe we should check depth of the stack trace? In netty there are some tricks around this area, if the future is complete synchronously all the time and makes a very deep call stack trace, it will force schedule an asynchronous task to prevent stack overflow and also reduce the blocking execution time.
> 
> The potential depth of the stack is out of our control. We are providing a client to an application that is developed by an independent team. So the move of internalFlush and the potentially synchronous downstream completions it may invoke is definitely the right move here.

This can be a separated issue, maybe we could change the implementation of FutureUtils.addListener.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
Apache9: For the test, I think you can just comment it out first, IIRC it was introduced to verify a race issue. If we comletely change the implementation, maybe we can just drop the test.

apurtell: This change was accomplished in 08911e7

apurtell:
@Apache9 All feedback items have been addressed. This PR is waiting for approval. I'd like to merge it and then apply it down to 2.5/2.6 so I can test it with ITBLL in the next 2.6 RC as suggested during review.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -118,33 +175,43 @@ protected void internalFlush() {
...
+          lock.lock();
+          try {
             // confirm that we are still valid, if there is already an internalFlush call before us,
             // then we should not execute anymore. And in internalFlush we will set periodicFlush
             // to null, and since we may schedule a new one, so here we check whether the references
             // are equal.
             if (timeout == periodicFlushTask) {
               periodicFlushTask = null;
-              internalFlush();
+              shouldFlush = true;
```
Apache9: I think here we should also use use drainBatch and then send the batch outside lock pattern, Otherwise I'm afraid there could still be race...

apurtell: Sure, implemented as b0a160c073

Apache-HBase:






| Vote | Subsystem | Runtime |  Logfile | Comment |
|:----:|----------:|--------:|:--------:|:-------:|
| +0 :ok: |  reexec  |   0m 12s |  |  Docker mode activated.  |
|||| _ Prechecks _ |
| +1 :green_heart: |  dupname  |   0m  0s |  |  No case conflicting files found.  |
| +0 :ok: |  codespell  |   0m  0s |  |  codespell was not available.  |
| +0 :ok: |  detsecrets  |   0m  0s |  |  detect-secrets was not available.  |
| +1 :green_heart: |  @author  |   0m  0s |  |  The patch does not contain any @author tags.  |
| +1 :green_heart: |  hbaseanti  |   0m  0s |  |  Patch does not have any anti-patterns.  |
|||| _ master Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 17s |  |  Maven dependency ordering for branch  |
| +1 :green_heart: |  mvninstall  |   5m 12s |  |  master passed  |
| +1 :green_heart: |  compile  |   5m 35s |  |  master passed  |
| +1 :green_heart: |  checkstyle  |   1m 38s |  |  master passed  |
| +1 :green_heart: |  spotbugs  |   3m  9s |  |  master passed  |
| +1 :green_heart: |  spotless  |   1m  6s |  |  branch has no errors when running spotless:check.  |
|||| _ Patch Compile Tests _ |
| +0 :ok: |  mvndep  |   0m 13s |  |  Maven dependency ordering for patch  |
| +1 :green_heart: |  mvninstall  |   4m 39s |  |  the patch passed  |
| +1 :green_heart: |  compile  |   5m 35s |  |  the patch passed  |
| +1 :green_heart: |  javac  |   5m 35s |  |  the patch passed  |
| +1 :green_heart: |  blanks  |   0m  0s |  |  The patch has no blanks issues.  |
| +1 :green_heart: |  checkstyle  |   1m 38s |  |  the patch passed  |
| +1 :green_heart: |  spotbugs  |   3m 31s |  |  the patch passed  |
| +1 :green_heart: |  hadoopcheck  |  14m 18s |  |  Patch does not cause any errors with Hadoop 3.3.6 3.4.1.  |
| +1 :green_heart: |  spotless  |   0m 59s |  |  patch has no errors when running spotless:check.  |
|||| _ Other Tests _ |
| +1 :green_heart: |  asflicense  |   0m 23s |  |  The patch does not generate ASF License warnings.  |
|  |   |  57m 22s |  |  |


| Subsystem | Report/Notes |
|----------:|:-------------|
| Docker | ClientAPI=1.48 ServerAPI=1.48 base: https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/4/artifact/yetus-general-check/output/Dockerfile |
| GITHUB PR | https://github.com/apache/hbase/pull/7363 |
| Optional Tests | dupname asflicense javac spotbugs checkstyle codespell detsecrets compile hadoopcheck hbaseanti spotless |
| uname | Linux 6f851a98550e 6.8.0-1024-aws #26~22.04.1-Ubuntu SMP Wed Feb 19 06:54:57 UTC 2025 x86_64 x86_64 x86_64 GNU/Linux |
| Build tool | maven |
| git revision | master / b0a160c07331a050071859ed957409cccf930937 |
| Default Java | Eclipse Adoptium-17.0.11+9 |
| Max. process+thread count | 70 (vs. ulimit of 30000) |
| modules | C: hbase-client hbase-server U: . |
| Console output | https://ci-hbase.apache.org/job/HBase-PreCommit-GitHub-PR/job/PR-7363/4/console |
| versions | git=2.34.1 maven=3.9.8 spotbugs=4.7.3 |
| Powered by | Apache Yetus 0.15.0 https://yetus.apache.org |


hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -90,21 +105,63 @@ public Configuration getConfiguration() {
 
   // will be overridden in test
   protected void internalFlush() {
```
Apache9, line 107: Do we still need this method? Seems we do not call it anywhere.

TestAsyncBufferedMutator, i think we should override drainBatch method now.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -90,21 +105,63 @@ public Configuration getConfiguration() {
...
+   * {@link #futures} lists into a returned {@link Batch}, replaces them with fresh lists, and
+   * resets {@link #bufferedSize} to zero.
+   * <p>
+   * If there is nothing buffered, returns {@code null} so callers can skip sending work.
+   * @return a {@link Batch} containing drained mutations and futures, or {@code null} if empty
+   */
+  private Batch drainBatch() {
+    ArrayList<Mutation> toSend;
+    ArrayList<CompletableFuture<Void>> toComplete;
+    lock.lock();
```
Apache9, line 130: I think we have already hold lock while calling  this method? So we do not need to lock it again.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -153,24 +220,45 @@ Stream.<CompletableFuture<Void>> generate(CompletableFuture::new).limit(mutation
...
+    if (batch != null) {
+      sendBatch(batch);
     }
     return futures;
   }
 
   @Override
-  public synchronized void flush() {
+  public void flush() {
     internalFlush();
```
Apache9, line 241: Ah, OK, there is still a call here, but I think we'd better implement it by our own here and remove the internalFlush method.
```
    Batch batch = null;
    if (!closed) {
      lock.lock();
      try {
        if (!closed) {
          batch = drainBatch(); // Drains under lock
        }
      } finally {
        lock.unlock();
      }
    }
    // Send the batch
    if (batch != null) {
      sendBatch(batch); // Sends outside of lock
    }
```

The only difference comparing to close method is that we do not need to set closed to true.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncBufferedMutatorImpl.java
```diff
@@ -153,24 +220,45 @@ Stream.<CompletableFuture<Void>> generate(CompletableFuture::new).limit(mutation
       bufferedSize += heapSize;
       if (bufferedSize >= writeBufferSize) {
         LOG.trace("Flushing because write buffer size {} reached", writeBufferSize);
-        internalFlush();
-      } else if (maxMutations > 0 && this.mutations.size() >= maxMutations) {
+        // drain now and send after releasing the lock
+        batch = drainBatch();
+      } else if (this.mutations.size() >= maxMutations) {
```
d-c-manning, line 225: did we lose the positive `maxMutations` check? (Looks like default is `-1` if I'm reading code correctly, which would lead to flush on every mutate.)
```suggestion
      } else if (maxMutations > 0 && this.mutations.size() >= maxMutations) {
```

d-c-manning: .

apurtell: It seems we did. I will fix this oversight.

Apache9:
Any updates here @apurtell ?

I think this is a very important change.

The hbase-operator-tools changes are almost done. I plan to include this change and run ITBLL for branch-3 again, before cutting 3.0.0-beta-2.

Thanks.

Apache9:
Any updates here?

If no objections, I could reuse most of the code here and fix the test code mentioned above to get this in, so I can start running ITBLL for branch-3 to prepare for the 3.0.0-beta-2 release.

Thanks.

apurtell:
Hey @Apache9 sorry for the delay I am on a holiday break. I am back next week and can finish this or if you would like to take it over sooner that is no problem!

Apache9:
#7516

PR #7730, opened 2026-02-10
HBASE-29691: Change TableName.META_TABLE_NAME from being a global static

Kota-SH:
[HBASE-29691: ](https://issues.apache.org/jira/browse/HBASE-29691) Change TableName.META_TABLE_NAME from being a global static 

Change TableName.META_TABLE_NAME from a global static constant to a dynamically discovered value from ConnectionRegistry. 

Please refer to earlier discussion on the PR https://github.com/apache/hbase/pull/7558/ for further details. 

Key changes:
- Add ConnectionRegistry.getMetaTableName() method for dynamic discovery
- Add MetaTableNameStore for persisting meta table name in master region
- Update TableName to support dynamic meta table name
- Update HMaster to integrate with MetaTableNameStore
- Update all client and server code to use dynamic meta table name
- Add protobuf changes for meta table name in Registry.proto

Refactoring for the test classes to be handled separately.

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfoBuilder.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfoBuilder.java
@@ -42,6 +42,8 @@ public class RegionInfoBuilder {
    */
   // TODO: How come Meta regions still do not have encoded region names? Fix.
   // hbase:meta,,1.1588230740 should be the hbase:meta first region name.
+  // TODO: For now, hardcode to default. Future: lazy initialization based on config or make it use
+  // connection
   public static final RegionInfo FIRST_META_REGIONINFO =
     new MutableRegionInfo(1L, TableName.META_TABLE_NAME, RegionInfo.DEFAULT_REPLICA_ID);
 
--- a/hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestAsyncAdminRpcPriority.java
+++ b/hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestAsyncAdminRpcPriority.java
@@ -137,8 +137,8 @@ public class TestAsyncAdminRpcPriority {
     }).when(adminStub).stopServer(any(HBaseRpcController.class), any(StopServerRequest.class),
       any());
     User user = UserProvider.instantiate(CONF).getCurrent();
-    conn = new AsyncConnectionImpl(CONF, new DoNothingConnectionRegistry(CONF, user), "test", null,
-      user) {
+    conn = new AsyncConnectionImpl(CONF, new DoNothingConnectionRegistry(CONF, user), "test",
+      TableName.META_TABLE_NAME, null, user) {
 
       @Override
       CompletableFuture<MasterService.Interface> getMasterStub() {
@@ -187,7 +187,7 @@ public class TestAsyncAdminRpcPriority {
   // that we pass the correct priority
   @Test
   public void testCreateMetaTable() {
-    conn.getAdmin().createTable(TableDescriptorBuilder.newBuilder(TableName.META_TABLE_NAME)
+    conn.getAdmin().createTable(TableDescriptorBuilder.newBuilder(conn.getMetaTableName())
       .setColumnFamily(ColumnFamilyDescriptorBuilder.of("cf")).build()).join();
     verify(masterStub, times(1)).createTable(assertPriority(SYSTEMTABLE_QOS),
       any(CreateTableRequest.class), any());
--- a/hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestAsyncRegionLocatorTracing.java
+++ b/hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestAsyncRegionLocatorTracing.java
@@ -95,7 +95,7 @@ public class TestAsyncRegionLocatorTracing {
       public CompletableFuture<RegionLocations> getMetaRegionLocations() {
         return CompletableFuture.completedFuture(locs);
       }
-    }, "test", null, user);
+    }, "test", TableName.META_TABLE_NAME, null, user);
   }
 
   @AfterEach
@@ -142,38 +142,36 @@ public class TestAsyncRegionLocatorTracing {
 
   @Test
   public void testClearCacheTableName() {
-    conn.getLocator().clearCache(TableName.META_TABLE_NAME);
+    conn.getLocator().clearCache(conn.getMetaTableName());
     SpanData span = waitSpan("AsyncRegionLocator.clearCache");
     assertThat(span,
       allOf(hasStatusWithCode(StatusCode.OK), hasKind(SpanKind.INTERNAL),
         buildConnectionAttributesMatcher(conn),
-        buildTableAttributesMatcher(TableName.META_TABLE_NAME)));
+        buildTableAttributesMatcher(conn.getMetaTableName())));
   }
 
   @Test
   public void testGetRegionLocation() {
-    conn.getLocator().getRegionLocation(TableName.META_TABLE_NAME, HConstants.EMPTY_START_ROW,
+    conn.getLocator().getRegionLocation(conn.getMetaTableName(), HConstants.EMPTY_START_ROW,
       RegionLocateType.CURRENT, TimeUnit.SECONDS.toNanos(1)).join();
     SpanData span = waitSpan("AsyncRegionLocator.getRegionLocation");
-    assertThat(span,
-      allOf(hasStatusWithCode(StatusCode.OK), hasKind(SpanKind.INTERNAL),
-        buildConnectionAttributesMatcher(conn),
-        buildTableAttributesMatcher(TableName.META_TABLE_NAME),
-        hasAttributes(containsEntryWithStringValuesOf("db.hbase.regions",
-          locs.getDefaultRegionLocation().getRegion().getRegionNameAsString()))));
+    assertThat(span, allOf(hasStatusWithCode(StatusCode.OK), hasKind(SpanKind.INTERNAL),
+      buildConnectionAttributesMatcher(conn), buildTableAttributesMatcher(conn.getMetaTableName()),
+      hasAttributes(containsEntryWithStringValuesOf("db.hbase.regions",
+        locs.getDefaultRegionLocation().getRegion().getRegionNameAsString()))));
   }
 
   @Test
   public void testGetRegionLocations() {
-    conn.getLocator().getRegionLocations(TableName.META_TABLE_NAME, HConstants.EMPTY_START_ROW,
+    conn.getLocator().getRegionLocations(conn.getMetaTableName(), HConstants.EMPTY_START_ROW,
       RegionLocateType.CURRENT, false, TimeUnit.SECONDS.toNanos(1)).join();
     SpanData span = waitSpan("AsyncRegionLocator.getRegionLocations");
     String[] expectedRegions =
       Arrays.stream(locs.getRegionLocations()).map(HRegionLocation::getRegion)
         .map(RegionInfo::getRegionNameAsString).toArray(String[]::new);
     assertThat(span, allOf(hasStatusWithCode(StatusCode.OK), hasKind(SpanKind.INTERNAL),
-      buildConnectionAttributesMatcher(conn),
-      buildTableAttributesMatcher(TableName.META_TABLE_NAME), hasAttributes(
+      buildConnectionAttributesMatcher(conn), buildTableAttributesMatcher(conn.getMetaTableName()),
+      hasAttributes(
         containsEntryWithStringValuesOf("db.hbase.regions", containsInAnyOrder(expectedRegions)))));
   }
 }
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/CellComparatorImpl.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/CellComparatorImpl.java
@@ -786,9 +786,21 @@ public class CellComparatorImpl implements CellComparator {
    * @return CellComparator to use going off the {@code tableName} passed.
    */
   public static CellComparator getCellComparator(byte[] tableName) {
-    // FYI, TableName.toBytes does not create an array; just returns existing array pointer.
-    return Bytes.equals(tableName, TableName.META_TABLE_NAME.toBytes())
+    return isMetaTable(tableName)
       ? MetaCellComparator.META_COMPARATOR
       : CellComparatorImpl.COMPARATOR;
+  }
+
+  /**
+   * @see TableName#isMetaTableName(TableName) for the canonical definition of what qualifies as a
+   *      meta table. Kept as a {@code byte[]}-friendly adapter so we can avoid materializing a
+   *      {@link TableName} on hot comparator-selection paths when the input is malformed.
+   */
+  static boolean isMetaTable(byte[] tableName) {
+    try {
+      return TableName.isMetaTableName(TableName.valueOf(tableName));
+    } catch (IllegalArgumentException e) {
+      return false;
+    }
   }
 }
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
@@ -22,9 +22,12 @@ import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
+import java.util.regex.Pattern;
 import org.apache.commons.lang3.ArrayUtils;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.apache.hbase.thirdparty.com.google.common.base.Preconditions;
 
@@ -44,6 +47,8 @@ import org.apache.hbase.thirdparty.com.google.common.base.Preconditions;
  */
 @InterfaceAudience.Public
 public final class TableName implements Comparable<TableName> {
+
+  private static final Logger LOG = LoggerFactory.getLogger(TableName.class);
 
   /** See {@link #createTableNameIfNecessary(ByteBuffer, ByteBuffer)} */
   private static final Set<TableName> tableCache = new CopyOnWriteArraySet<>();
@@ -66,6 +71,7 @@ public final class TableName implements Comparable<TableName> {
     + NAMESPACE_DELIM + ")?)" + "(?:" + VALID_TABLE_QUALIFIER_REGEX + "))";
 
   /** The hbase:meta table's name. */
+  @Deprecated
   public static final TableName META_TABLE_NAME =
     valueOf(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR, "meta");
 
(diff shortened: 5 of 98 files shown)


anmolnar (approved):
lgtm. What are the build failures?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfoBuilder.java
```diff
@@ -42,6 +42,8 @@ public class RegionInfoBuilder {
    */
   // TODO: How come Meta regions still do not have encoded region names? Fix.
   // hbase:meta,,1.1588230740 should be the hbase:meta first region name.
+  // TODO: For now, hardcode to default. Future: lazy initialization based on config or make it use
+  // conenction
```
kgeisz: nit:
```suggestion
  // connection
```

hbase-common/src/main/java/org/apache/hadoop/hbase/CellComparatorImpl.java
```diff
@@ -787,8 +787,13 @@ public static CellComparator getCellComparator(TableName tableName) {
...
     // FYI, TableName.toBytes does not create an array; just returns existing array pointer.
-    return Bytes.equals(tableName, TableName.META_TABLE_NAME.toBytes())
+    // Check if this is a meta table (hbase:meta or hbase:meta_*)
+    return isMetaTable(tableName)
       ? MetaCellComparator.META_COMPARATOR
       : CellComparatorImpl.COMPARATOR;
   }
+
+  static boolean isMetaTable(byte[] tableName) {
+    return Bytes.startsWith(tableName, TableName.META_TABLE_NAME.getName());
```
kgeisz: Why are we using `Bytes.startsWith()` here instead of `Bytes.equals()`?

hbase-common/src/main/java/org/apache/hadoop/hbase/CellComparatorImpl.java
```diff
@@ -787,8 +787,13 @@ public static CellComparator getCellComparator(TableName tableName) {
    */
   public static CellComparator getCellComparator(byte[] tableName) {
     // FYI, TableName.toBytes does not create an array; just returns existing array pointer.
```
kgeisz: nit: Is this comment still applicable?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/janitor/MetaFixer.java
```diff
@@ -203,19 +203,20 @@ private static List<RegionInfo> createMetaEntries(final MasterServices masterSer
...
       .filter(Either::hasRight).map(Either::getRight).collect(Collectors.toList());
-    LOG.debug("Added {}/{} entries to hbase:meta", createMetaEntriesSuccesses.size(),
-      newRegionInfos.size());
+    LOG.debug("Added {}/{} entries to {}", createMetaEntriesSuccesses.size(), newRegionInfos.size(),
+      TableName.META_TABLE_NAME.getNameAsString());
 
     if (!createMetaEntriesFailures.isEmpty()) {
       LOG.warn(
-        "Failed to create entries in hbase:meta for {}/{} RegionInfo descriptors. First"
+        "Failed to create entries in {}} for {}/{} RegionInfo descriptors. First"
```
kgeisz: nit
```suggestion
        "Failed to create entries in {} for {}/{} RegionInfo descriptors. First"
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/ModifyTableProcedure.java
```diff
@@ -110,7 +110,7 @@ protected void preflightChecks(MasterProcedureEnv env, Boolean enabled) throws H
       for (byte[] family : UNDELETABLE_META_COLUMNFAMILIES) {
         if (!cfs.contains(family)) {
           throw new HBaseIOException(
-            "Delete of hbase:meta column family " + Bytes.toString(family));
+            "Delete of " + env.getMetaTableName() + " column family " + Bytes.toString(family));
```
kgeisz, line 113: nit: IMO this could be a little more clear
```suggestion
            "Invalid deletion of " + env.getMetaTableName() + " column family: " + Bytes.toString(family));
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetaTableNameStore.java
```diff
@@ -0,0 +1,93 @@
...
+      get.addColumn(INFO_FAMILY, NAME_QUALIFIER);
+      Result result = masterRegion.get(get);
+
+      if (!result.isEmpty()) {
+        byte[] value = result.getValue(INFO_FAMILY, NAME_QUALIFIER);
+        cachedMetaTableName = TableName.valueOf(Bytes.toString(value));
+        LOG.debug("Loaded meta table name from Master Local Region: {}", cachedMetaTableName);
+        return cachedMetaTableName;
+      }
+      LOG.info("No stored meta table name found in Master Local Region:  {}", cachedMetaTableName);
```
kgeisz: nit
```suggestion
      LOG.info("No stored meta table name found in Master Local Region: {}", cachedMetaTableName);
```

hbase-server/src/main/java/org/apache/hadoop/hbase/tool/BulkLoadHFilesTool.java
```diff
@@ -660,21 +660,21 @@ private int getRegionIndex(List<Pair<byte[], byte[]>> startEndKeys, byte[] key)
   private void checkRegionIndexValid(int idx, List<Pair<byte[], byte[]>> startEndKeys,
     TableName tableName) throws IOException {
     if (idx < 0) {
-      throw new IOException("The first region info for table " + tableName
-        + " can't be found in hbase:meta.Please use hbck tool to fix it first.");
+      throw new IOException("The first region info for table " + tableName + " can't be found in "
+        + "hbase:meta. Please use hbck tool to fix it first.");
```
kgeisz: Just want to confirm - Are we okay with hard-coding `hbase:meta` here?

hbase-server/src/main/java/org/apache/hadoop/hbase/tool/BulkLoadHFilesTool.java
```diff
@@ -660,21 +660,21 @@ private int getRegionIndex(List<Pair<byte[], byte[]>> startEndKeys, byte[] key)
...
     } else if (
       (idx == startEndKeys.size() - 1)
         && !Bytes.equals(startEndKeys.get(idx).getSecond(), HConstants.EMPTY_BYTE_ARRAY)
     ) {
-      throw new IOException("The last region info for table " + tableName
-        + " can't be found in hbase:meta.Please use hbck tool to fix it first.");
+      throw new IOException("The last region info for table " + tableName + " can't be found in "
+        + "hbase:meta. Please use hbck tool to fix it first.");
```
kgeisz: Same with here

hbase-server/src/main/java/org/apache/hadoop/hbase/tool/BulkLoadHFilesTool.java
```diff
@@ -660,21 +660,21 @@ private int getRegionIndex(List<Pair<byte[], byte[]>> startEndKeys, byte[] key)
...
     } else if (
       idx + 1 < startEndKeys.size() && !(Bytes.compareTo(startEndKeys.get(idx).getSecond(),
         startEndKeys.get(idx + 1).getFirst()) == 0)
     ) {
       throw new IOException("The endkey of one region for table " + tableName
         + " is not equal to the startkey of the next region in hbase:meta."
```
kgeisz: And here

hbase-server/src/main/java/org/apache/hadoop/hbase/util/HBaseFsck.java
```diff
@@ -2019,9 +2021,11 @@ void closeRegion(HbckRegionInfo hi) throws IOException, InterruptedException {
...
       RegionInfo hri = h.getRegion();
       if (hri == null) {
-        LOG.warn("Unable to close region " + hi.getRegionNameAsString()
-          + " because hbase:meta had invalid or missing " + HConstants.CATALOG_FAMILY_STR + ":"
-          + Bytes.toString(HConstants.REGIONINFO_QUALIFIER) + " qualifier value.");
+        LOG.warn(
+          "Unable to close region " + hi.getRegionNameAsString()
+            + " because {} had invalid or missing " + HConstants.CATALOG_FAMILY_STR + ":"
+            + Bytes.toString(HConstants.REGIONINFO_QUALIFIER) + " qualifier value.",
+          connection.getMetaTableName());
```
kgeisz: nit: I see a mixture of using string concatenation as well as `{}` in the `LOG.warn()` method.  IMO is should consistently use the `{}` convention.

hbase-server/src/main/java/org/apache/hadoop/hbase/util/HBaseFsck.java
```diff
@@ -2196,7 +2202,8 @@ else if (!inMeta && !inHdfs && !isDeployed) {
             }
           }
         }
-        LOG.info("Patching hbase:meta with .regioninfo: " + hbi.getHdfsHRI());
+        LOG.info("Patching {} with .regioninfo: " + hbi.getHdfsHRI(),
+          connection.getMetaTableName());
```
kgeisz: nit: Mixture of string concatenation and `{}` in log method

hbase-server/src/main/java/org/apache/hadoop/hbase/util/HBaseFsck.java
```diff
@@ -2224,7 +2231,8 @@ else if (!inMeta && !inHdfs && !isDeployed) {
           return;
         }
 
-        LOG.info("Patching hbase:meta with with .regioninfo: " + hbi.getHdfsHRI());
+        LOG.info("Patching {} with with .regioninfo: " + hbi.getHdfsHRI(),
+          connection.getMetaTableName());
```
kgeisz: nit: Mixture of string concatenation and `{}` in log method

hbase-server/src/main/java/org/apache/hadoop/hbase/util/HBaseFsck.java
```diff
@@ -2301,9 +2309,9 @@ else if (!inMeta && !inHdfs && !isDeployed) {
       }
     } else if (inMeta && inHdfs && isMultiplyDeployed) {
       errors.reportError(ERROR_CODE.MULTI_DEPLOYED,
-        "Region " + descriptiveName + " is listed in hbase:meta on region server "
-          + hbi.getMetaEntry().regionServer + " but is multiply assigned to region servers "
-          + Joiner.on(", ").join(hbi.getDeployedOn()));
+        "Region " + descriptiveName + " is listed in " + connection.getMetaTableName()
+          + " on region server " + hbi.getMetaEntry().regionServer + " but is multiply assigned"
```
kgeisz, line 2312: nit: Do you have any idea what they mean by "multiply assigned"?  Like "multi-assigned"?

kgeisz (changes requested):
LGTM overall.  I have some nit comments below.  Also, I see PR https://github.com/apache/hbase/pull/7558 is still open.  Should that PR be closed?

Apache9:
Just came back from vacation...
This is a big change, please give me more time for reviewing...
Thanks.

anmolnar:
Regarding Copilot's concerns about incorrect meta table name prefix matching: is the user able to create tables in the `hbase:` namespace?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionInfoBuilder.java
```diff
@@ -42,6 +42,8 @@ public class RegionInfoBuilder {
    */
   // TODO: How come Meta regions still do not have encoded region names? Fix.
   // hbase:meta,,1.1588230740 should be the hbase:meta first region name.
+  // TODO: For now, hardcode to default. Future: lazy initialization based on config or make it use
+  // connection
```
anmolnar, line 46: Is this going to work for now with the default name?

hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestAsyncAdminRpcPriority.java
```diff
@@ -187,7 +187,7 @@ public void testCreateSystemTable() {
   // that we pass the correct priority
   @Test
   public void testCreateMetaTable() {
-    conn.getAdmin().createTable(TableDescriptorBuilder.newBuilder(TableName.META_TABLE_NAME)
+    conn.getAdmin().createTable(TableDescriptorBuilder.newBuilder(conn.getMetaTableName())
```
anmolnar, line 190: I thought that we don't change the default meta table name in unit tests in this patch for now. Rather we'll do them at once if needed. Why is this change?

hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestAsyncRegionLocatorTracing.java
```diff
@@ -142,38 +142,36 @@ public void testClearCacheServerName() {
 
   @Test
   public void testClearCacheTableName() {
-    conn.getLocator().clearCache(TableName.META_TABLE_NAME);
+    conn.getLocator().clearCache(conn.getMetaTableName());
```
anmolnar, line 145: Same here. Are these tests not passing without the modification?

hbase-common/src/main/java/org/apache/hadoop/hbase/CellComparatorImpl.java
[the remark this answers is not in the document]
anmolnar: You might want to be more specific here.
- Table name equals to `hbase:meta`, or
- Table name matches pattern `hbase:meta_[a-z0-9]+`

As long as we agree on that meta table name will be generated from a hash value.

hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
```diff
@@ -85,9 +90,18 @@ public final class TableName implements Comparable<TableName> {
...
   public static boolean isMetaTableName(final TableName tn) {
-    return tn.equals(TableName.META_TABLE_NAME);
+    if (
+      tn == null || !tn.getNamespaceAsString().equals(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR)
+    ) {
+      return false;
+    }
+    String qualifier = tn.getQualifierAsString();
+    return qualifier.equals("meta") || qualifier.startsWith("meta_");
   }
```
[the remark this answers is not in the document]
anmolnar: Can we do this validation at a single place potentially here in this class?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionStateStore.java
```diff
@@ -283,7 +283,8 @@ private CompletableFuture<Void> updateRegionLocation(RegionInfo regionInfo, Stat
         future = FutureUtils.failedFuture(e);
       }
     } else {
-      AsyncTable<?> table = master.getAsyncConnection().getTable(TableName.META_TABLE_NAME);
+      AsyncTable<?> table =
+        master.getAsyncConnection().getTable(master.getConnection().getMetaTableName());
```
anmolnar:
```java
master.getConnection().getMetaTableName()
```
Do we need `Connection` here instead of `AsyncConnection`?
We need `AsyncConnection` anyways due to `getTable()`, so we could use the same for `getMetaTableName()`.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionStateStore.java
```diff
@@ -329,16 +330,16 @@ private void multiMutate(RegionInfo ri, List<Mutation> mutations) throws IOExcep
       }
     }
     MutateRowsRequest request = builder.build();
-    AsyncTable<?> table =
-      master.getConnection().toAsyncConnection().getTable(TableName.META_TABLE_NAME);
+    AsyncTable<?> table = master.getConnection().toAsyncConnection()
+      .getTable(master.getConnection().getMetaTableName());
```
anmolnar: Uhh, this is getting weird. Why not doing the same as above?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionStateStore.java
```diff
@@ -504,7 +504,7 @@ public void deleteMergeQualifiers(RegionInfo mergeRegion) throws IOException {
         + " in meta table, they are cleaned up already, Skip.");
       return;
     }
-    try (Table table = master.getConnection().getTable(TableName.META_TABLE_NAME)) {
+    try (Table table = master.getConnection().getTable(master.getConnection().getMetaTableName())) {
```
anmolnar: Does it have any advantage if we cache the Connection like:
```java
Connection c = master.getConnection();
Table table = c.getTable(c.getMetaTableName());
```
Does it make any difference?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/janitor/MetaFixer.java
```diff
@@ -203,19 +203,20 @@ private static List<RegionInfo> createMetaEntries(final MasterServices masterSer
         .flatMap(List::stream).collect(Collectors.toList());
     final List<IOException> createMetaEntriesFailures = addMetaEntriesResults.stream()
       .filter(Either::hasRight).map(Either::getRight).collect(Collectors.toList());
-    LOG.debug("Added {}/{} entries to hbase:meta", createMetaEntriesSuccesses.size(),
-      newRegionInfos.size());
+    LOG.debug("Added {}/{} entries to {}", createMetaEntriesSuccesses.size(), newRegionInfos.size(),
+      TableName.META_TABLE_NAME.getNameAsString());
```
anmolnar: You're hardcoding the deprecated TableName anyways, so I think we could just skip modifying this file at all.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/InitMetaProcedure.java
```diff
@@ -67,6 +67,8 @@ public class InitMetaProcedure extends AbstractStateMachineTableProcedure<InitMe
 
   @Override
   public TableName getTableName() {
+    // Always returns default "hbase:meta" for now
+    // Future: will support replica-specific names (e.g., "hbase:meta_replica1")
```
anmolnar, line 71: So, this will be the point where we'll populate the meta table name from config.

anmolnar (commented):
First round of review.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/ModifyTableProcedure.java
```diff
@@ -109,8 +109,8 @@ protected void preflightChecks(MasterProcedureEnv env, Boolean enabled) throws H
       Set<byte[]> cfs = this.modifiedTableDescriptor.getColumnFamilyNames();
       for (byte[] family : UNDELETABLE_META_COLUMNFAMILIES) {
         if (!cfs.contains(family)) {
-          throw new HBaseIOException(
-            "Delete of hbase:meta column family " + Bytes.toString(family));
+          throw new HBaseIOException("Invalid deletion of " + env.getMetaTableName()
```
anmolnar: You've modified the text of exception message, hence the unit test `testEditMeta` failure.
At the end of the test it validates the error message. Right fix would be, I think, is to keep the original message prefix intact.

anmolnar:
@kgeisz @Apache9 Please take another look at this patch. It's in a good shape. The test error is a flaky. Thanks!

hbase-server/src/main/java/org/apache/hadoop/hbase/HBaseServerBase.java
```diff
@@ -745,4 +745,16 @@ private String resolveHostName(Configuration conf, String useThisHostnameInstead
...
+  protected TableName getDefaultMetaTableName() {
+    return TableName.META_TABLE_NAME;
+  }
+
+  @Override
+  public TableName getMetaTableName() {
+    // For now, it is hbase:meta because we don't support custom meta table name.
+    // After adding support for custom meta table names, we can calculate this from conf and use it
+    // downstream to persist it in Master Region.
+    return getDefaultMetaTableName();
```
anmolnar, line 758: Is this comment still valid?

anmolnar (approved):
lgtm. A few of my previous questions haven't been answered yet.

PR #7761, opened 2026-02-16
HBASE-29905 BackupLogCleaner: skip tables no longer in the backup set

janvanbesien:
BackupLogCleaner.serverToPreservationBoundaryTs() computes the WAL deletion boundary by iterating over tableSetTimestampMap from persisted BackupInfo sessions. This map can contain entries for tables that were once part of the backup set but have since had all their backups deleted. Their stale, old timestamps drag the minimum WAL boundary back, preventing old WALs from being cleaned up.

Fix: when computing boundaries, load the incrbackupset per backup root from BackupSystemTable and skip tables not in the active set. The incrbackupset is populated on every full backup and pruned when backups are deleted, so it accurately reflects which tables still need WAL retention.

### DIFF
--- a/hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/master/TestBackupLogCleaner.java
+++ b/hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/master/TestBackupLogCleaner.java
@@ -20,10 +20,14 @@ package org.apache.hadoop.hbase.backup.master;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Collection;
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
@@ -36,6 +40,7 @@ import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.backup.BackupInfo;
 import org.apache.hadoop.hbase.backup.BackupType;
 import org.apache.hadoop.hbase.backup.TestBackupBase;
+import org.apache.hadoop.hbase.backup.impl.BackupManager;
 import org.apache.hadoop.hbase.backup.impl.BackupSystemTable;
 import org.apache.hadoop.hbase.backup.util.BackupBoundaries;
 import org.apache.hadoop.hbase.client.Connection;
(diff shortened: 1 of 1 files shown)


DieterDP-ng:
Hi Jan,

2 remarks regarding the format of the PR:

- HBASE accepts PRs for the master branch. If accepted for master, it's backported to the other branches by the person doing the merge.
- HBASE asks that the commit header message matches the title of the ticket it solves.

hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/master/TestBackupLogCleaner.java
```diff
@@ -215,6 +216,59 @@ public void testBackupLogCleaner() throws Exception {
     }
   }
 
+  @Test
+  public void testRemovedTableDoesNotPinWals() throws Exception {
```
DieterDP-ng: The name of this test is misleading. I'd also add a textual description as to the specific case you're checking here.

The current name implies it has to do with a table that was removed. That is also the problem originally encountered (in NGDATA): logs were not being removed for a table that has already been removed. But you're not testing for that here, this test doesn't remove a table.

There's a difference in the logic that's relevant. If you were to extend your testcase to the following: `full backup F1`, `incremental I1`, `full backup F2` (table 1 only), `incremental I2` (table 1 only), `delete F1, I1`, I'm pretty sure the test would fail because `I2` would still refer to table2.

However, if you deleted table 2 after `I1`, I suspect it will work as expected.

DieterDP-ng: I suggest to also modify your commit message a bit to explain the main use case being tables that have since been deleted.

janvanbesien: Thanks for your review dieter.

As I understand it, the problem can be triggered by explicitly removing a table from the backup (i.e. explicitly deciding to no longer backup said table) or by removing the table completely (i.e. implicitly it is no longer in any future backup either). You're right that I mixed the two cases and that the primary trigger was the latter, not the former.

I now have a test for both scenarios. Does that make sense?

Also updated commit message and rebased on master.

DieterDP-ng:
I've force-pushed a rebase on the current master.

hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/master/TestBackupLogCleaner.java
```diff
@@ -205,6 +208,154 @@ public void testBackupLogCleaner() throws Exception {
...
+      TEST_UTIL.truncateTable(BackupSystemTable.getTableName(TEST_UTIL.getConfiguration())).close();
+    }
+  }
+
+  /**
+   * Similar as {@link #testRemovedBackupDoesNotPinWals()} but for the case where a table is
+   * removed, and hence no longer in any backup either.
+   */
+  @Test
+  public void testRemovedTableDoesNotPinWals() throws Exception {
```
DieterDePaepe: You probably added this test due to my previous comment, but seeing it in action now, I now realize it serves no purpose. The fact that the table is removed has no effect (i.e. all assertions are fine even if the table isn't removed).

The mechanism added in this PR is that the logcleaner only cares about the tables still in scope for backups, and that's covered in the first test.

Given that this test takes +-30secs to run, I think its best to remove it.

hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/master/TestBackupLogCleaner.java
```diff
@@ -205,6 +208,154 @@ public void testBackupLogCleaner() throws Exception {
...
+      Map<String, Object> params = new HashMap<>(1);
+      params.put(HMaster.MASTER, TEST_UTIL.getHBaseCluster().getMaster());
+      cleaner.init(params);
+
+      // Create FULL backup B1 with table1 and table2
+      String backupIdB1 =
+        backupTables(BackupType.FULL, Arrays.asList(table1, table2), backupRoot.toString());
+      assertTrue(checkSucceeded(backupIdB1));
+
+      Set<FileStatus> walFilesAfterB1 =
```
DieterDePaepe, line 232: Add a sanity check that this isn't empty (and for `walFilesAfterB2`).

hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/master/BackupLogCleaner.java
DieterDePaepe, line 1: I rebased your changes to the latest master, and there was some refactoring involved. Would appreciate a second pair of eyes to take a look at the new code.

hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/master/TestBackupLogCleaner.java
```diff
@@ -205,6 +208,154 @@ public void testBackupLogCleaner() throws Exception {
...
+        for (int i = 0; i < NB_ROWS_IN_BATCH; i++) {
+          Put p = new Put(Bytes.toBytes("stale-row-t1" + i));
+          p.addColumn(famName, qualName, Bytes.toBytes("val" + i));
+          t1.put(p);
+        }
+      }
+
+      // Create FULL backup B2 with only table1.
+      // B2's tableSetTimestampMap carries forward the old timestamp from B1 for table2,
+      // while table1 gets a fresh timestamp: { table1: ts(B2), table2: ts(B1) }
```
DieterDePaepe, line 257: Would be useful to verify this in the test as well.

hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/master/BackupLogCleaner.java
DieterDePaepe, line 1: Two observations that I had while reviewing, but not something that needs fixing in this PR. Mainly thinking out loud, and might log issues for these.

- If a table is deleted, it will result in WAL files sticking around until all backups containing that deleted table are gone (because that is when the incremental backup set gets updated). If long backup retention times are used, this can still be problematic. There's room for further improvement here.
- If a table is deleted and recreated with the same name, the backup info will still have the timestamps of the deleted table, and WALs that are included in incremental backups will also contain data of the deleted table. This would lead to a corrupted restore.

DieterDePaepe (changes requested):
A few minor changes to do, looks good overall.

DieterDePaepe:
@hgromer Could you also review this PR?

hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/master/BackupLogCleaner.java
janvanbesien: I fixed all your remarks. However, when trying to rebase on master, I noticed that the test seems to work now without any code changes. So either the problem has been fixed it in the mean time via another change, or my test is no good.

I pushed it like this (test only) but I should find some more time to deep dive.

PR #7762, opened 2026-02-17
HBASE-29774 incremental backup fails on empty WAL

thomassngdata:
Add handling for WALHeaderEOFException in WALInputFormat with retries similar to other caught exceptions, but supress the exception after retries because empty WAL files should be skipped.

IncrementalTableBackupClient.convertWALsToHFiles fails with WALHeaderEOFException "EOF while reading PB WAL magic" when processing an empty WAL file. Because the WALInputFormat reader does not handle the WALHeaderEOFException, unlike the WALEntryStream reader.

### DIFF
--- a/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
+++ b/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
@@ -32,6 +32,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.RemoteIterator;
+import org.apache.hadoop.hbase.regionserver.wal.WALHeaderEOFException;
 import org.apache.hadoop.hbase.util.LeaseNotRecoveredException;
 import org.apache.hadoop.hbase.wal.AbstractFSWALProvider;
 import org.apache.hadoop.hbase.wal.WAL;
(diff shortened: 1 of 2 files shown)


hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
```diff
@@ -154,13 +155,22 @@ private WALStreamReader openReader(Path path, long startPosition) throws IOExcep
       int attempt = 0;
       Exception ee = null;
       WALStreamReader reader = null;
+      boolean supressException = false;
```
DieterDePaepe, line 158: Typo.

hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
```diff
@@ -154,13 +155,22 @@ private WALStreamReader openReader(Path path, long startPosition) throws IOExcep
...
       while (reader == null && attempt++ < maxAttempts) {
         try {
           // Detect if this is a new file, if so get a new reader else
           // reset the current reader so that we see the new data
           reader =
             WALFactory.createStreamReader(path.getFileSystem(conf), path, conf, startPosition);
           return reader;
+        } catch (WALHeaderEOFException wheofe) {
```
DieterDePaepe, line 166: I've looked into the potential causes for this exception. The only legitimate case this can occur in normal operation is when using the legacy WAL writer, where there's a minimal time period between opening and flushing the first entry. Other cases are due to crashing Region Servers.

I prefer skipping the retry because:
- it causes the test to run for over a minute.
- in case the empty WAL file originated from a crash, a retry will not solve anything, and just cause a long delay.
- given the only legitimate case for the empty WAL (the legacy WAL writer) simply caused a crash before, it's not a regression to just skip the WAL file instead.

DieterDePaepe (commented):
Because Thomas (my colleague) is currently unavailable, I've implemented the remarks I made myself in a new PR - see https://github.com/apache/hbase/pull/8327

DieterDePaepe:
Closing this PR as https://github.com/apache/hbase/pull/8327 got accepted and merged.

PR #7818, opened 2026-02-27
HBASE-29809 Unify region reopening methods to use throttled implement…

liuxiaocs7:
…ation

- see: HBASE-29809

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
@@ -4316,17 +4316,21 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
       throw new TableNotFoundException(tableName);
     }
 
+    TableDescriptor tableDescriptor = getTableDescriptors().get(tableName);
+    if (tableDescriptor == null) {
+      throw new TableNotFoundException(tableName);
+    }
+
     return MasterProcedureUtil
       .submitProcedure(new MasterProcedureUtil.NonceProcedureRunnable(this, nonceGroup, nonce) {
         @Override
         protected void run() throws IOException {
           ReopenTableRegionsProcedure proc;
           if (regionNames.isEmpty()) {
-            proc = ReopenTableRegionsProcedure.throttled(getConfiguration(),
-              getTableDescriptors().get(tableName));
+            proc = ReopenTableRegionsProcedure.throttled(getConfiguration(), tableDescriptor);
           } else {
-            proc = ReopenTableRegionsProcedure.throttled(getConfiguration(),
-              getTableDescriptors().get(tableName), regionNames);
+            proc = ReopenTableRegionsProcedure.throttled(getConfiguration(), tableDescriptor,
+              regionNames);
           }
 
           LOG.info("{} throttled reopening {} regions for table {}", getClientIdAuditPrefix(),
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
@@ -4268,7 +4268,12 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
    *                    unique id).
    * @return procedure Id
    * @throws IOException if reopening region fails while running procedure
+   * @deprecated since 3.0.0 and will be removed in 4.0.0. Use
+   *             {@link #reopenRegionsThrottled(TableName, List, long, long)} instead so region
+   *             reopening honors the configured throttling.
+   * @see <a href="https://issues.apache.org/jira/browse/HBASE-29809">HBASE-29809</a>
    */
+  @Deprecated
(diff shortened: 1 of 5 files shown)


liuxiaocs7:
Hi, @ajkh88 and @charlesconnell, could you help take a look when free, thanks!

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/ReopenTableRegionsProcedure.java
```diff
@@ -91,7 +92,7 @@ public class ReopenTableRegionsProcedure
   /**
    * Create a new ReopenTableRegionsProcedure respecting the throttling configuration for the table.
    * First check the table descriptor, then fall back to the global configuration. Only used in
-   * ModifyTableProcedure and in {@link HMaster#reopenRegionsThrottled}.
```
charlesconnell, line 94: Are you sure about this javadoc change? This looks wrong. I think to make the link work property there need to be args, like `{@link HMaster#reopenRegionsThrottled(TableName, ...}`

hbase-server/src/main/java/org/apache/hadoop/hbase/master/RegionsRecoveryChore.java
```diff
@@ -94,11 +93,11 @@ protected void chore() {
           tableToReopenRegionsMap.forEach((tableName, regionNames) -> {
             try {
               LOG.warn("Reopening regions due to high storeFileRefCount. "
-                + "TableName: {} , noOfRegions: {}", tableName, regionNames.size());
-              hMaster.reopenRegions(tableName, regionNames, NONCE_GENERATOR.getNonceGroup(),
```
charlesconnell, line 98: I can't really comment on whether it's a good idea to throttle these reopens since I don't have enough context on this code path.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/ReopenTableRegionsProcedure.java
liuxiaocs7: Thanks for your review, yes, these modifications are unnecessary and can be removed

hbase-server/src/main/java/org/apache/hadoop/hbase/master/RegionsRecoveryChore.java
liuxiaocs7: Hi, @charlesconnell, thanks for your comments, `RegionsRecoveryChore` used to reopen region whose `maxCompactedStoreFileRefCount` metric large than the configured `storeFileRefCountThreshold`, if throttle feature is enabled, we can avoid reopen them at the same time. At the same time, the default behavior will remain consistent with previous

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/ReopenTableRegionsProcedure.java
liuxiaocs7: image

image

charlesconnell (approved):

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
```diff
@@ -4251,6 +4251,7 @@ public void remoteProcedureFailed(long procId, RemoteProcedureException error) {
    * @return procedure Id
    * @throws IOException if reopening region fails while running procedure
    */
+  @Deprecated
```
Apache9, line 4276: Please add a deprecation cycle and explaination about the deprecation? Or if this is not a IA.Public class, let's just remove it?

liuxiaocs7: Done! I’ve added the deprecation cycle and an explanation for the deprecation.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/ReopenTableRegionsProcedure.java
```diff
@@ -31,6 +31,7 @@
 import org.apache.hadoop.hbase.UnknownRegionException;
 import org.apache.hadoop.hbase.client.TableDescriptor;
 import org.apache.hadoop.hbase.conf.ConfigKey;
+import org.apache.hadoop.hbase.master.HMaster;
```
Apache9: Why adding this import?

liuxiaocs7: Strange, Let me check!

Thanks

liuxiaocs7: Oh, I see, there is a java doc warning, let me remove this time.

image

Apache9 (approved):

liuxiaocs7:
Failed UT seems unrelated, let me merge

liuxiaocs7:
Thanks @charlesconnell and @Apache9 for the reviewing!

PR #7830, opened 2026-03-02
HBASE-27380 RitDuration histogram metric is broken

liuxiaocs7:
- see: HBASE-27380

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hbase.master.assignment;
 
+import com.google.errorprone.annotations.RestrictedApi;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
@@ -39,8 +40,6 @@ import org.slf4j.LoggerFactory;
 @InterfaceAudience.Private
 public class RegionInTransitionTracker {
   private static final Logger LOG = LoggerFactory.getLogger(RegionInTransitionTracker.class);
-  private static final LongConsumer NOOP_RIT_DURATION_CONSUMER = ignored -> {
-  };
 
   private static final List<RegionState.State> DISABLE_TABLE_REGION_STATE =
     List.of(RegionState.State.OFFLINE, RegionState.State.CLOSED);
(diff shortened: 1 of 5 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
```diff
@@ -43,8 +46,11 @@ public class RegionInTransitionTracker {
 
   private final ConcurrentSkipListMap<RegionInfo, RegionStateNode> regionInTransition =
     new ConcurrentSkipListMap<>(RegionInfo.COMPARATOR);
+  private final ConcurrentHashMap<RegionInfo, Long> regionEnterTimestamp =
```
Apache9: Is it possible to merge this map to the above regionInTransition map? We can change the value of the above map to contain the timestamp. And do we really need ConcurrentSkipListMap here? Can we just change the above map to ConcurrentHashMap for better performance?

vaijosh: Agree with @Apache9.

@liuxiaocs7 
Please check if we can use ConcurrentHashMap<RegionInfo, Pair<RegionStateNode,Long>>

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestAssignmentManagerRitDurationMetrics.java
```diff
@@ -0,0 +1,144 @@
...
+        MASTER.getAssignmentManager().getRegionStates().getRegionServerOfRegion(regionInfo);
+      ServerName target = MASTER.getServerManager().getOnlineServersList().stream()
+        .filter(sn -> !sn.equals(current)).findFirst()
+        .orElseThrow(() -> new IllegalStateException("Need at least two regionservers"));
+
+      TEST_UTIL.getAdmin().move(regionInfo.getEncodedNameAsBytes(), target);
+      TEST_UTIL.waitFor(10_000, () -> getRitCountFromRegionStates(amSource) > ritDurationNumOps);
+      TEST_UTIL.waitUntilNoRegionTransitScheduled();
+      Thread.sleep(MSG_INTERVAL * 5);
+      assertEquals(ritDurationNumOps + 1, getRitCountFromRegionStates(amSource));
```
vaijosh: @liuxiaocs7 
The test currently asserts only the counter ritDurationNumOps increments: assertEquals(ritDurationNumOps + 1, getRitCountFromRegionStates(amSource)); This verifies an observation was recorded, but it doesn’t validate that the ritDuration histogram itself is being exported/updated correctly (the bug this PR targets).

Can we also assert at least one histogram-derived metric?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestAssignmentManagerRitDurationMetrics.java
```diff
@@ -0,0 +1,144 @@
...
+    }
+  }
+
```
vaijosh, line 101: Can we add a test/assertions to verify that the consumer was injected and ws called exactly once(or expected number of time) with a non-negative duration when a region transitions?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
```diff
@@ -131,15 +137,29 @@ public void handleRegionDelete(RegionInfo regionInfo) {
...
+        EnvironmentEdgeManager.currentTime());
+    }
+    return added;
   }
 
   private boolean removeRegionInTransition(final RegionInfo regionInfo) {
-    return regionInTransition.remove(regionInfo) != null;
+    boolean removed = regionInTransition.remove(regionInfo) != null;
+    if (removed) {
+      Long enter = regionEnterTimestamp.remove(regionInfo);
```
vaijosh: @liuxiaocs7 
Shouldn't we remove entries from regionInTransition and regionEnterTimestamp?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
liuxiaocs7: Thanks @Apache9 and @vaijosh, I will take a closer look.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestAssignmentManagerRitDurationMetrics.java
liuxiaocs7: Thanks @vaijosh, done!

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestAssignmentManagerRitDurationMetrics.java
liuxiaocs7: Added in `TestRegionInTransitionTracker#testInjectedRitDurationConsumerUsesFirstEnterTimestamp`

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
liuxiaocs7: Now we use `ConcurrentHashMap<RegionInfo, Pair<RegionStateNode, Long>> regionInTransition` instead, both the state node and enter timestamp
  are stored in the same map entry, so a single remove clears both together

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
liuxiaocs7: Make sense, done!

Apache9 (approved):

vaijosh (approved):
LGTM

PR #7855, opened 2026-03-05
HBASE-28123 TableRequests metrics broken after altering table

liuxiaocs7:
- see: HBASE-28123


hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/metrics/impl/GlobalMetricRegistriesAdapter.java
```diff
@@ -63,15 +63,17 @@ public final class GlobalMetricRegistriesAdapter {
...
+    MetricsSourceAdapter(MetricRegistryInfo info) {
+      this.info = info;
     }
 
     @Override
     public void getMetrics(MetricsCollector collector, boolean all) {
-      metricsAdapter.snapshotAllMetrics(registry, collector);
+      Optional<MetricRegistry> registryOpt = MetricRegistries.global().get(info);
+      registryOpt
+        .ifPresent(metricRegistry -> metricsAdapter.snapshotAllMetrics(metricRegistry, collector));
```
liuxiaocs7, line 76: this is the fix, we should always use the newest `metricRegistry` here

liuxiaocs7:
Hi, @Apache9, what do you think of this change for this bug, more details see HBASE-28123, Thanks!

liuxiaocs7:
Now the CI pass all!

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/metrics/MetricsTableRequests.java
```diff
@@ -110,14 +111,14 @@ public class MetricsTableRequests {
   private MetricRegistryInfo registryInfo;
 
   private boolean enableTableLatenciesMetrics;
-  private boolean enabTableQueryMeterMetrics;
```
liuxiaocs7, line 113: Typo

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/metrics/TestMetricsTableRequests.java
```diff
@@ -119,15 +113,33 @@ public void testTableQueryMeterSwitch() {
...
+    TableName tn1 = TableName.valueOf(testInfo.getTestMethod().get().getName());
+
+    // create registry twice, should return same instance due to ref-counting
+    MetricsTableRequests requests1 = new MetricsTableRequests(tn1, new Configuration());
+    MetricsTableRequests requests2 = new MetricsTableRequests(tn1, new Configuration());
+
+    MetricRegistry metricRegistry1 = requests1.getMetricRegistry();
+    MetricRegistry metricRegistry2 = requests2.getMetricRegistry();
+    // verify ref-counting returns same instance
+    assertEquals(metricRegistry1, metricRegistry2);
```
Apache9: Should use assertSame?

Apache9 (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/metrics/TestMetricsTableRequests.java
liuxiaocs7: Thanks for the comments, done!

Apache9 (approved):

liuxiaocs7:
Failed UTs seems unrelated, let me merge.
Thanks @Apache9 for the reviewing!

PR #7864, opened 2026-03-05
HBASE-29912 Codel lifoThreshold should be applied on soft queue limit…

Umeshkumar9414:
… instead of hard limit

I think that after [HBASE-16089](https://issues.apache.org/jira/browse/HBASE-16089) codel target delay was changed to 100, that is not standard for CoDel ([patch](https://issues.apache.org/jira/secure/attachment/12813161/HBASE-16089.v3.patch)). I have changed it back. 
To visualize why we should have 5 target delay with 100 interval, I used the [model](https://github.com/Umeshkumar9414/hbase/blob/0056886fb540ca1cfd9035aec489a3106c47bc09/hbase-server/src/main/resources/hbase-webapps/queue-simulator/adaptive-lifo-codel-simulator.html), With target delay 100 and interval 100 we are more prone the queue being full, although CoDel will be still helpful in LIFO mode but a full queue is problem as it can cause OOM.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
@@ -44,7 +44,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   private LinkedBlockingDeque<CallRunner> queue;
 
   // so we can calculate actual threshold to switch to LIFO under load
-  private int maxCapacity;
+  private int currentQueueLimit;
 
   // metrics (shared across all queues)
   private LongAdder numGeneralCallsDropped;
@@ -71,14 +71,15 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   private AtomicBoolean isOverloaded = new AtomicBoolean(false);
 
   public AdaptiveLifoCoDelCallQueue(int capacity, int targetDelay, int interval,
-    double lifoThreshold, LongAdder numGeneralCallsDropped, LongAdder numLifoModeSwitches) {
-    this.maxCapacity = capacity;
+    double lifoThreshold, LongAdder numGeneralCallsDropped, LongAdder numLifoModeSwitches,
+    int currentQueueLimit) {
     this.queue = new LinkedBlockingDeque<>(capacity);
     this.codelTargetDelay = targetDelay;
     this.codelInterval = interval;
     this.lifoThreshold = lifoThreshold;
     this.numGeneralCallsDropped = numGeneralCallsDropped;
     this.numLifoModeSwitches = numLifoModeSwitches;
+    this.currentQueueLimit = currentQueueLimit;
   }
 
   /**
@@ -86,12 +87,14 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
    * @param newCodelTargetDelay new CoDel target delay
    * @param newCodelInterval    new CoDel interval
    * @param newLifoThreshold    new Adaptive Lifo threshold
+   * @param currentQueueLimit   new limit of queue
    */
-  public void updateTunables(int newCodelTargetDelay, int newCodelInterval,
-    double newLifoThreshold) {
+  public void updateTunables(int newCodelTargetDelay, int newCodelInterval, double newLifoThreshold,
+    int currentQueueLimit) {
     this.codelTargetDelay = newCodelTargetDelay;
     this.codelInterval = newCodelInterval;
     this.lifoThreshold = newLifoThreshold;
+    this.currentQueueLimit = currentQueueLimit;
   }
 
   /**
@@ -104,7 +107,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   public CallRunner take() throws InterruptedException {
     CallRunner cr;
     while (true) {
-      if (((double) queue.size() / this.maxCapacity) > lifoThreshold) {
+      if (((double) queue.size() / this.currentQueueLimit) > lifoThreshold) {
         numLifoModeSwitches.increment();
         cr = queue.takeLast();
       } else {
@@ -124,7 +127,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
     CallRunner cr;
     boolean switched = false;
     while (true) {
-      if (((double) queue.size() / this.maxCapacity) > lifoThreshold) {
+      if (((double) queue.size() / this.currentQueueLimit) > lifoThreshold) {
         // Only count once per switch.
         if (!switched) {
           switched = true;
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
@@ -44,7 +44,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   private LinkedBlockingDeque<CallRunner> queue;
 
   // so we can calculate actual threshold to switch to LIFO under load
-  private int currentQueueLimit;
+  private volatile int currentQueueLimit;
 
   // metrics (shared across all queues)
   private LongAdder numGeneralCallsDropped;
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
@@ -44,7 +44,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   private LinkedBlockingDeque<CallRunner> queue;
 
   // so we can calculate actual threshold to switch to LIFO under load
-  private volatile int currentQueueLimit;
+  private volatile int softLimit;
 
   // metrics (shared across all queues)
   private LongAdder numGeneralCallsDropped;
@@ -79,7 +79,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
     this.lifoThreshold = lifoThreshold;
     this.numGeneralCallsDropped = numGeneralCallsDropped;
     this.numLifoModeSwitches = numLifoModeSwitches;
-    this.currentQueueLimit = currentQueueLimit;
+    this.softLimit = currentQueueLimit;
   }
 
   /**
@@ -94,7 +94,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
     this.codelTargetDelay = newCodelTargetDelay;
     this.codelInterval = newCodelInterval;
     this.lifoThreshold = newLifoThreshold;
-    this.currentQueueLimit = currentQueueLimit;
+    this.softLimit = currentQueueLimit;
   }
 
   /**
@@ -107,7 +107,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   public CallRunner take() throws InterruptedException {
     CallRunner cr;
     while (true) {
-      if (((double) queue.size() / this.currentQueueLimit) > lifoThreshold) {
+      if (((double) queue.size() / this.softLimit) > lifoThreshold) {
         numLifoModeSwitches.increment();
         cr = queue.takeLast();
       } else {
@@ -127,7 +127,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
     CallRunner cr;
     boolean switched = false;
     while (true) {
-      if (((double) queue.size() / this.currentQueueLimit) > lifoThreshold) {
+      if (((double) queue.size() / this.softLimit) > lifoThreshold) {
         // Only count once per switch.
         if (!switched) {
           switched = true;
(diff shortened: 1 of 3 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
```diff
@@ -44,7 +44,7 @@ public class AdaptiveLifoCoDelCallQueue implements BlockingQueue<CallRunner> {
   private LinkedBlockingDeque<CallRunner> queue;
 
   // so we can calculate actual threshold to switch to LIFO under load
-  private int maxCapacity;
+  private int currentQueueLimit;
```
apurtell: Because this can be updated with `onConfigurationChange` should this be `volatile` ? Just a nit.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
```diff
@@ -231,20 +234,7 @@ public Map<String, Long> getCallQueueSizeSummary() {
...
   }
 
-  // This method can only be called ONCE per executor instance.
-  // Before calling: queueInitArgs[0] contains the soft limit (desired queue capacity)
-  // After calling: queueInitArgs[0] is set to hard limit and currentQueueLimit stores the original
-  // soft limit.
-  // Multiple calls would incorrectly use the hard limit as the soft limit.
-  // As all the queues has same initArgs and queueClass, there should be no need to call this again.
   protected void initializeQueues(final int numQueues) {
-    if (!queues.isEmpty()) {
```
apurtell, line 241: Do we still need this guard?

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
apurtell: This is also confusing in that it seems to shadow a variable of the same name that *is* declared volatile? 

Please don't shadow. Why do we need this separately?

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
```diff
@@ -814,4 +817,166 @@ public void drop() {
...
+
+      // Track completion order
+      final List<Integer> completedCalls = Collections.synchronizedList(new ArrayList<>());
+
+      // Dispatch many slow calls rapidly to fill the queue beyond 80% threshold
+      // With queue limit of 50, we need > 40 calls to cross 80%
+      int numCalls = 48;
+      for (int i = 0; i < numCalls; i++) {
+        final int callId = i;
+        CallRunner call = createMockTask(HConstants.NORMAL_QOS);
```
apurtell, line 877: Here and in two more cases createMockTask will mock a ServerCall without setting getReceiveTime(). It will default to 0. Doesn't this mess up the call delay calculation that coDel will do?  It will be something like `callDelay = System.currentTimeMillis() - 0` . I guess your tests are passing but CoDel will immediately detect overload. The LIFO switch counter will still increment but other aspects are probably wrong. 
You should look at this.

apurtell (changes requested):

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
Umeshkumar9414: I made currentQueueLimit volatile. 

> Please don't shadow. Why do we need this separately?

Generally a queue should only handle hard limit and layer above it can handle soft limit and that is what was happening but for codel, we needed queue to know the soft limit as well, I didn't saw a way to use the same variable at both place.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
Umeshkumar9414: no we don't need it now.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/AdaptiveLifoCoDelCallQueue.java
apurtell: At least let's not shadow the parent class's variable with the same name. That is a code smell. Pick another one.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
apurtell: Why not? DOesn't it change the early out behavior?

Umeshkumar9414: We added this guard becuase we used to initialize soft limit from queueInitArgs[0] and then change it to hard limit. Now if we call it again. We will set soft limit with queueInitArgs[0] that is hard limit now. 

From this change we are moving the calculation of soft limit and hard limit to RPCExecutor.

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
Umeshkumar9414: Fixed it. Thanks

Umeshkumar9414:
Yetus JDK17 Hadoop3 Unit check failed on TestNamespaceReplication Test case with timout, I checked on my local, it is passing. No related change in this PR as well. Looks like a flakey.

Umeshkumar9414:
cc @saintstack

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
apurtell: Thank you, resolving this.

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
apurtell: Thank you, resolving this.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
```diff
@@ -466,7 +457,15 @@ public void resizeQueues(Configuration conf) {
       }
     }
     final int queueLimit = currentQueueLimit;
-    currentQueueLimit = conf.getInt(configKey, queueLimit);
+    int newQueueLimit = conf.getInt(configKey, queueLimit);
+    if (newQueueLimit > DEFAULT_CALL_QUEUE_SIZE_HARD_LIMIT) {
+      LOG.warn(
+        "Requested soft limit {} exceeds dynamic-resize ceiling {}. "
```
apurtell: This WARN will be misleading. 
The actual limit is `queueHardLimit`, initialized in the constructor as Math.max(maxQueueLength, 250).

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
```diff
@@ -90,7 +90,7 @@ public abstract class RpcExecutor {
   public static final String CALL_QUEUE_CODEL_LIFO_THRESHOLD =
     "hbase.ipc.server.callqueue.codel.lifo.threshold";
 
-  public static final int CALL_QUEUE_CODEL_DEFAULT_TARGET_DELAY = 100;
+  public static final int CALL_QUEUE_CODEL_DEFAULT_TARGET_DELAY = 5;
```
apurtell, line 93: This change is fine for `master` branch, and even `branch-2` but will need to be left to the current default in the releasing branches per our compatibility guidelines, which disallow changes that affect operational behaviors. Existing clusters relying on the looser 100ms target will drop calls maybe more aggressively than expected or benchmarked under the same load.

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
```diff
@@ -808,4 +837,161 @@ public void drop() {
...
+          completedCalls.add(callId);
+          Thread.sleep(100); // Slow processing to allow queue to build up
+          return null;
+        }).when(call).run();
+        scheduler.dispatch(call);
+        // No delay between dispatches - rapidly fill the queue
+      }
+
+      // Wait for some calls to complete
+      await().atMost(2, TimeUnit.SECONDS).until(() -> completedCalls.size() >= 3);
```
apurtell, line 889: This will probably flake the test.
The assertion below for `maxCallIdCompleted` inspect up to 5 completions. With only 3 guaranteed by the `await()`, we might miss a completion that would otherwise land in the 4th/5th slot.

apurtell (changes requested):
Minor changes needed. Otherwise lgtm.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
Umeshkumar9414: Actually I doesn't want to store queueHardLimit as well in the object and without storing it we won't be able to know the actual queueHardLimit. Should I store it anyway ?

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
Umeshkumar9414: In that case let me raise an different PR for this change. I want rest of the change in release branches.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
Umeshkumar9414: added an variable queueHardLimit in the object and changed it

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
Umeshkumar9414: removed the inspection upto 5

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
Umeshkumar9414: We can skip this while backporting.

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
Umeshkumar9414: By the timing we are using we will only need 3 completed run.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
```diff
@@ -466,7 +459,15 @@ public void resizeQueues(Configuration conf) {
...
     }
     final int queueLimit = currentQueueLimit;
-    currentQueueLimit = conf.getInt(configKey, queueLimit);
+    int newQueueLimit = conf.getInt(configKey, queueLimit);
+    if (newQueueLimit > queueHardLimit) {
+      LOG.warn(
+        "Requested soft limit {} exceeds queue hard limit/capacity {}. "
+          + "A region server restart is required to grow the underlying queue.",
+        newQueueLimit, queueHardLimit);
+      newQueueLimit = DEFAULT_CALL_QUEUE_SIZE_HARD_LIMIT;
```
apurtell: Shouldn't this be `newQueueLimit = queueHardLimit;` ? 

Also it would be cleaner to declare this `protected final int queueHardLimit;` because it's never changed outside of the constructor.

apurtell (changes requested):

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/RpcExecutor.java
Umeshkumar9414: my bad. Thanks

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
```diff
@@ -808,4 +838,161 @@ public void drop() {
...
+    } finally {
+      scheduler.stop();
+    }
+  }
+
+  /**
+   * Test that CoDel queue returns to FIFO mode after draining below threshold.
+   */
+  @Test
+  public void testCoDelQueueDrainAndFifoReturn() throws Exception {
```
apurtell, line 917: The new test runs 5 calls (IDs 100–104) on 2 handlers with 50ms processing each, then asserts that completion order has violations <= 1. With two threads pulling concurrently from a single queue FIFO take() can produce an order like 100, 102, 101, 103, 104 (two violations) when handler A picks 100 before handler B picks 101 but B finishes first while A was paused by the OS scheduler. On a busy CI runner this will eventually flake. 

Replace the order-violation count with a check that the first completion has the smallest id (completedCalls.get(0) == 100). This is what FIFO actually guarantees for a single producer issuing sequential dispatch calls, and is robust to handler interleaving.

apurtell (commented):
Everything else looks good, but the new test will probably flake.

hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestSimpleRpcScheduler.java
Umeshkumar9414: yeah, I agree, I missed it. There is also possibility that 101 can finish first as handler A got paused. So lets me just add assert for NumLifoModeSwitches.

apurtell (approved):

Umeshkumar9414:
For branch-2.6 :  https://github.com/apache/hbase/pull/8344
For branch-2 : https://github.com/apache/hbase/pull/8345

Umeshkumar9414:
For branch-2.5 https://github.com/apache/hbase/pull/8346

PR #7882, opened 2026-03-07
HBASE-29974: Persist filter hints across scan circuit breaks

shubham-roy:
### JIRA
[HBASE-29974](https://issues.apache.org/jira/browse/HBASE-29974)

---

## Problem / Motivation

HBase's scan pipeline applies "structural short-circuit checks" in `UserScanQueryMatcher.matchColumn` — including time-range gates, column-set exclusion, and version-limit exhaustion — before `Filter.filterCell` executes. When these gates trigger, the pipeline returns `SKIP` or `SEEK_NEXT_COL` without consulting the filter.

Similarly, `RegionScannerImpl` iterates through rejected rows cell-by-cell after `filterRowKey` returns `true`, even when filters could compute the next valid row boundary deterministically.

The consequence is that filter implementations that could provide meaningful seek hints are never consulted, forcing cell-by-cell advancement instead of issuing a single `requestSeek` / `backwardSeek`, resulting in significant read amplification for range-oriented or sparse-column filters.

## Root Cause

The existing `getNextCellHint` / `SEEK_NEXT_USING_HINT` mechanism requires `filterCell` invocation first. No API contract existed for filters to provide forward hints when cells were structurally excluded before reaching `filterCell`. The three structural gates and `filterRowKey` rejection path lacked hook points.

### New Filter API Methods (`Filter.java`, `FilterBase.java`)

- **`getHintForRejectedRow(Cell firstRowCell)`** — called after `filterRowKey` returns `true`. The filter may use state set during `filterRowKey` (e.g., an updated range pointer) to compute a seek target that bypasses the entire rejected row in a single seek.
- **`getSkipHint(Cell skippedCell)`** — called when cells are rejected by time-range, column-set, or version-limit gates before `filterCell` is ever reached. Must be stateless — only filters with immutable configuration-driven hint computation (e.g., fixed column ranges, fuzzy-row patterns) should override this.

Both methods return `null` by default, preserving full backward compatibility.

### RegionScannerImpl Changes

- Added `reversed` field (from `scan.isReversed()`) to support direction-aware seeking.
- Introduced `nextRowViaHint(ScannerContext, Cell, ExtendedCell)` — a fast-path alternative to cell-by-cell `nextRow()` iteration. Issues a single `requestSeek` (forward) or `backwardSeek` (reversed) to the filter's suggested position.
- Added `getHintForRejectedRow(Cell)` helper that validates the hint is an `ExtendedCell`.
- The hint is only used if it advances past the current row in scan direction (using `compareRows` for row-level comparison); otherwise, falls back safely to `nextRow()`.

### UserScanQueryMatcher Changes

- Added `reversed` field and `pendingSkipHint` field.
- Introduced `resolveSkipHint(ExtendedCell)` — consults the filter's `getSkipHint` at each structural short-circuit point, validates the hint type and direction, and stores it for `getNextKeyHint` to return.
- The direction-aware guard uses full-key compare (intentional: skip hints can advance within the same row to a later column, not just across rows): `(!reversed && cmp > 0) || (reversed && cmp < 0)`.
- `pendingSkipHint` is defensively cleared on row transitions (`setToNewRow`, `clearCurrentRow`) to prevent stale hints from leaking across rows.
- `beforeShipped()` deep-copies `pendingSkipHint` via `KeyValueUtil.toNewKeyCell` to prevent use-after-free on block release.

### FilterWrapper Delegation

Both new methods are delegated to the wrapped filter, consistent with existing `getNextCellHint` delegation.

## Backward Compatibility

- Both new methods return `null` by default — existing filter implementations are completely unaffected.
- All scanner call-sites are null-guarded: the fast-paths are never entered unless a filter explicitly provides a hint.
- `FilterList` inherits the no-op default and does not compose these methods (consistent with the fact that composition semantics require separate design; tracked for follow-up).

### Comparison semantics
- `nextRowViaHint` uses `compareRows` (row-only comparison) because it operates at the row level — the hint must target a different row.
- `resolveSkipHint` uses full-key `compare` (intentionally) because skip hints can advance within the same row (e.g., to a later column or timestamp).

### Stop-row invariant
`nextRowViaHint` does not pre-validate that the hint falls within the scan's stop row. If the hint overshoots, the next loop iteration's `shouldStop(Cell)` check catches it and returns `NO_MORE_VALUES`. One wasted seek may occur, but correctness is maintained.

### Metrics and coprocessor semantics
- The rows-scanned metric is incremented once for the rejected row. Rows physically skipped by the seek are not individually counted (no per-row work was done).
- `postScannerFilterRow` is invoked once with the rejected row's cell, not once per skipped row. Coprocessors counting filtered rows should be aware of this.

### Known limitations
- `FilterList`, `SkipFilter`, and `WhileMatchFilter` do not currently delegate `getHintForRejectedRow` or `getSkipHint` to wrapped sub-filters. Hints from filters used inside these wrappers are silently ignored. This requires AND/OR composition semantics design and is tracked for a follow-up JIRA.

### Integration Tests — `TestFilterHintForRejectedRow` (13 tests)

| Test | Scenario |
|------|----------|
| `testHintAndNoHintReturnSameCellsOnSameData` | Equivalence: hint skips 5 rows in one seek vs cell-by-cell — same results |
| `testHintBeyondLastRowTerminatesScanGracefully` | Hint past EOF → scan returns empty |
| `testPerRowHintCalledOncePerRejectedRow` | Per-row hint called exactly N times for N rejected rows |
| `testReversedScanWithHint` | `getHintForRejectedRow` with reversed scan |
| `testGetSkipHintWithTimeRangeIntegration` | `getSkipHint` on time-range structural skip (forward) |
| `testBackwardHintFallsBackToNextRow` | Backward hint → safe fallback to `nextRow()` |
| `testSameRowHintFallsBackToNextRow` | Same-row hint → safe fallback |
| `testCoprocessorHookCalledOncePerHintedRejection` | `postScannerFilterRow` hook contract maintained |
| `testMultiFamilyScanWithHint` | Multi-family scan returns all families correctly |
| `testDataCorrectnessVsUnfilteredScan` | Golden comparison: filtered = unfiltered with equivalent startRow |
| `testHintOvershootingStopRowTerminatesGracefully` | Hint past stop row → graceful scan termination |
| `testHintWithinStopRowReturnsCorrectResults` | Hint within stop row → correct subset matches baseline |
| `testGetSkipHintCalledForReversedScan` | `getSkipHint` invoked during reversed scan time-range gate |

### Unit Tests — `TestUserScanQueryMatcher` (7 new tests)

| Test | Gate tested |
|------|------------|
| `testSkipHintConsultedForCellNewerThanTimeRange` | `tsCmp > 0` |
| `testSkipHintConsultedForCellOlderThanTimeRange` | `tsCmp < 0` |
| `testNullSkipHintFallsThroughToOriginalCodeOnTimeRangeGate` | Null hint → original MatchCode preserved |
| `testSkipHintConsultedForExcludedColumn` | `checkColumn != INCLUDE` |
| `testSkipHintConsultedOnVersionExhaustion` | `checkVersions` → SEEK_NEXT_COL |
| `testNormalFilterCellHintPathUnaffectedBySkipHintChange` | Existing `filterCell` path unaffected |
| `testSkipHintConsultedForReversedScan` | Reversed scan: smaller-key hint accepted |


virajjasani:
@shubham-roy could you rebase with latest master?

virajjasani:
FYI @tkhurana

virajjasani:
Error Type | Count | Module | Tests
-- | -- | -- | --
Failures | 1 | hbase-server | org.apache.hadoop.hbase.io.hfile.bucket.TestPrefetchWithBucketCache.testPrefetchRunTriggersEvictions
Flakes | 3 | hbase-server | org.apache.hadoop.hbase.master.balancer.TestBalancerDecision.testBalancerDecisions  org.apache.hadoop.hbase.master.region.TestMasterRegionWALCleaner.test  org.apache.hadoop.hbase.security.access.TestZKPermissionWatcher.testPermissionsWatcher

virajjasani:
❌ | spotless | 0m 17s | patch has 63 errors when running spotless:check, run spotless:apply to fix.
-- | -- | -- | --

shubham-roy:
@virajjasani , I have resolved merge conflicts and have also applied spotless command.

For the test failures: 
org.apache.hadoop.hbase.io.hfile.bucket.TestPrefetchWithBucketCache.testPrefetchRunTriggersEvictions

The above passes in my local. Are we sure it is not a flapper?
Screenshot 2026-04-10 at 8 59 49 PM

Also, I believe the 3 count of flakes can be ignored, right?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/querymatcher/UserScanQueryMatcher.java
```diff
@@ -107,18 +117,26 @@ public Filter getFilter() {
...
-      } else {
-        throw new DoNotRetryIOException("Incorrect filter implementation, "
-          + "the Cell returned by getNextKeyHint is not an ExtendedCell. Filter class: "
-          + filter.getClass().getName());
-      }
-
+      throw new DoNotRetryIOException("Incorrect filter implementation, "
+        + "the Cell returned by getNextKeyHint is not an ExtendedCell. Filter class: "
+        + filter.getClass().getName());
     }
```
virajjasani, line 140: Is this part changing at all? I guess this is only structural change.
 
Can we remove this change just to keep it like it was before?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/querymatcher/TestUserScanQueryMatcher.java
```diff
@@ -398,6 +419,232 @@ scanWithFilter, new ScanInfo(this.conf, fam2, 0, 5, ttl, KeepDeletedCells.FALSE,
     assertArrayEquals(nextCell.getQualifierArray(), col4);
   }
 
+  // -----------------------------------------------------------------------
+  // Tests for HBASE-29974: Filter#getSkipHint consulted at matchColumn
+  // structural short-circuits (time-range, column, and version gates).
+  // -----------------------------------------------------------------------
```
virajjasani, line 425: Let's remove these lines

hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFilterHintForRejectedRow.java
```diff
@@ -0,0 +1,361 @@
...
+   * <li>The scanner seeks directly to {@code "row-05"} after the very first rejection, bypassing
+   * cells in rows {@code "row-01"} through {@code "row-04"} entirely.</li>
+   * <li>{@code getHintForRejectedRow} is called exactly once (not five times), confirming no
+   * cell-by-cell iteration occurred for the skipped rows.</li>
+   * <li>Rows {@code "row-05"} through {@code "row-09"} are returned with all their cells
+   * intact.</li>
+   * </ul>
+   */
+  @Test
+  public void testHintJumpsOverAllRejectedRowsInOneSingleSeek() throws IOException {
```
virajjasani, line 148: `testHintJumpsOverAllRejectedRowsInOneSingleSeek()` and `testNullHintFallsThroughToLegacyNextRowBehaviour()` should be combined in single test. When we check for data correctness, we should check it on the same table, same data, one with `hintFilter`, one with `noHintFilter` and compare cells. Separate tests do not serve the same purpose.

hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFilterHintForRejectedRow.java
```diff
@@ -0,0 +1,361 @@
...
+ * <li><b>Correctness</b> — the scan returns exactly the rows that are not rejected by
+ * {@link Filter#filterRowKey(Cell)}, regardless of whether the hint path or the legacy
+ * cell-iteration path is used.</li>
+ * <li><b>Efficiency</b> — when a filter provides a hint that jumps over all N rejected rows in one
+ * seek, {@code getHintForRejectedRow} is called exactly once (not N times), proving that the
+ * scanner did not fall back to cell-by-cell iteration for the skipped rows.</li>
+ * </ol>
+ */
+@Category({ FilterTests.class, MediumTests.class })
+public class TestFilterHintForRejectedRow {
```
virajjasani, line 69: For all tests in this class, it seems we are not overriding `getSkipHint()`. Let's create one test where we some cells are out of scan time-range and `getSkipHint()` is used. Then for the same data, we can use scan without `getSkipHint()` and it should still return same cells, for the data correctness comparison.

virajjasani (commented):
Left few comments, looks good overall

hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFilterHintForRejectedRow.java
virajjasani: Though we need both `getHintForRejectedRow()` and `getSkipHint()` implementations, 
for phoenix cdc, i believe `getSkipHint()` will more likely be common hint than `getHintForRejectedRow()`, is that correct?

shubham-roy: `getHintForRejectedRow()` is also equally applicable for the phoenix cdc based backup solution that we are developing. We identify row keys that were mutated for the backup time range via CDC and then get information for the same row keys. While skipping row keys that were not mutated `getHintForRejectedRow()` comes into play.

virajjasani (approved):
+1

virajjasani:
❌ | spotless | 0m 34s | patch has 23 errors when running spotless:check, run spotless:apply to fix
-- | -- | -- | --

PR #7901, opened 2026-03-11
HBASE-29669 Implement basic row cache

EungsopYoo:


EungsopYoo:
@wchevreuil 
I have completed the implementation of the basic row cache. Please review it.

By the way, it seems like CI is not working, do you know why?

liuxiaocs7:
Hi, @EungsopYoo

CI has been migrated from Jenkins to GitHub Actions, and the pre-commit GitHub Jenkins job should have been disabled, more details: https://lists.apache.org/thread/z6o5hhsd9goh5j7fcl4bnwzsktlwlwl4

Some previously created feature branches also have this issue; perhaps we could give it a try, such as: https://lists.apache.org/thread/hrl7dktkrb8zwwbv62my68lvtbplv7sr

Thanks!

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
```diff
@@ -110,16 +229,67 @@ boolean tryGetFromCache(RowCacheKey key, Get get, List<Cell> results) {
...
+  void increaseRowCacheSeqNum(HRegion region) {
+    region.increaseRowCacheSeqNum();
+  }
+
+  void removeTableLevelBarrier(HRegion region) {
+    regionLevelBarrierMap.computeIfPresent(region, (k, counter) -> {
+      int remaining = counter.decrementAndGet();
+      return (remaining <= 0) ? null : counter;
+    });
+  }
```
liuxiaocs7, line 280: this method should be renamed by `removeRegionLevelBarrier`?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
```diff
@@ -946,14 +949,19 @@ public HRegion(final HRegionFileSystem fs, final WAL wal, final Configuration co
...
     // The setting from TableDescriptor has higher priority than the global configuration
     return fromDescriptor != null
       ? fromDescriptor
       : conf.getBoolean(HConstants.ROW_CACHE_ENABLED_KEY, HConstants.ROW_CACHE_ENABLED_DEFAULT);
   }
 
+  // For testing only
+  void setRowCache(RowCache rowCache) {
+    this.rowCache = rowCache;
+  }
```
liuxiaocs7: For testing only, we could use `@RestrictedApi` annotation?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
```diff
@@ -69,6 +69,7 @@ class MetricsRegionServerWrapperImpl implements MetricsRegionServerWrapper {
   private BlockCache l2Cache = null;
   private MobFileCache mobFileCache;
   private CacheStats cacheStats;
+  private final RowCache rowCache;
```
liuxiaocs7: better move to L71 under `MobFileCache`?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
```diff
@@ -99,6 +100,8 @@ public MetricsRegionServerWrapperImpl(final HRegionServer regionServer) {
     this.regionServer = regionServer;
     initBlockCache();
     initMobFileCache();
+    RSRpcServices rsRpcServices = this.regionServer.getRSRpcServices();
+    this.rowCache = rsRpcServices == null ? null : rsRpcServices.getServer().getRowCache();
```
liuxiaocs7: better to use a method `initRowCache()` here, just like blockcache and mobfilecache does?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RSRpcServices.java
```diff
@@ -2347,8 +2347,21 @@ public BulkLoadHFileResponse bulkLoadHFile(final RpcController controller,
...
+    rowCache.createRegionLevelBarrier(region);
+    try {
+      // We do not invalidate the entire row cache directly, as it contains a large number of
+      // entries and takes a long time. Instead, we increment rowCacheSeqNum, which is used when
+      // constructing a RowCacheKey, thereby making the existing row cache entries stale.
+      rowCache.increaseRowCacheSeqNum(region);
+      return bulkLoadHFileInternal(request);
+    } finally {
+      // The row cache for the region has been enabled again
+      rowCache.removeTableLevelBarrier(region);
```
liuxiaocs7: ditto `region` level

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestRowCache.java
```diff
@@ -0,0 +1,547 @@
...
+import org.junit.Rule;
+import org.junit.Test;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.TestName;
+
+@Category({ RegionServerTests.class, MediumTests.class })
+public class TestRowCache {
+  @ClassRule
+  public static final HBaseClassTestRule CLASS_RULE =
+    HBaseClassTestRule.forClass(TestRowCache.class);
```
liuxiaocs7: For new tests, it is best to use JUnit5 directly, since existing unit tests are currently being migrated to JUnit5.

EungsopYoo:
@liuxiaocs7
Sorry, I just noticed your comments. I’ll take a look and follow up early next week.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
EungsopYoo: Done, renamed to removeRegionLevelBarrier. (95de81b)

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
EungsopYoo: Done, added @RestrictedApi annotation. (267f4ef)

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
EungsopYoo: Done, moved rowCache field next to MobFileCache. (782dd75)

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
EungsopYoo: Done, extracted initRowCache() method. (782dd75)

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RSRpcServices.java
EungsopYoo: Done, renamed to removeRegionLevelBarrier. (95de81b)

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestRowCache.java
EungsopYoo: Done, migrated TestRowCache to JUnit5. (1fc9d5c)

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
```diff
@@ -63,8 +85,8 @@ <R> R execute(RowOperation<R> operation) throws IOException {
   RowCache(Configuration conf) {
     enabledByConf =
       conf.getFloat(HConstants.ROW_CACHE_SIZE_KEY, HConstants.ROW_CACHE_SIZE_DEFAULT) > 0;
-    // TODO: implement row cache
-    rowCacheStrategy = null;
+    // Currently we only support TinyLfu implementation
+    rowCacheStrategy = new TinyLfuRowCacheStrategy(MemorySizeUtil.getRowCacheSize(conf));
```
wchevreuil: Should be made pluggable.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
```diff
@@ -110,16 +229,67 @@ boolean tryGetFromCache(RowCacheKey key, Get get, List<Cell> results) {
...
   }
 
-  void populateCache(List<Cell> results, RowCacheKey key) {
-    // TODO: implement with barrier to avoid cache read during mutation
-    try {
-      rowCacheStrategy.cacheRow(key, new RowCells(results));
-    } catch (CloneNotSupportedException ignored) {
-      // Not able to cache row cells, ignore
-    }
+  void populateCache(HRegion region, List<Cell> results, RowCacheKey key) {
```
wchevreuil: nit: just call this method "cache".

Do we really need the regionLevelBarrierMap key type to be HRegion, or could it be just the String encoded region name? That way, we don't need an extra HRegion parameter in this method.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
```diff
@@ -3435,6 +3473,15 @@ private void updateDeleteLatestVersionTimestamp(Cell cell, Get get, int count, b
...
   public void put(Put put) throws IOException {
     TraceUtil.trace(() -> {
+      // Put with TTL is not allowed on tables with row cache enabled, because cached rows cannot
+      // track TTL expiration
+      if (isRowCacheEnabled) {
+        if (put.getTTL() != Long.MAX_VALUE) {
+          throw new DoNotRetryIOException(
+            "Tables with row cache enabled do not allow setting TTL on Puts");
+        }
+      }
```
wchevreuil: Can't we simply implement TTL expiration within the RowCache.getRow logic?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
EungsopYoo: Done. Introduced `row.cache.strategy.class` config key so a custom `RowCacheStrategy` can be plugged in via configuration; default remains `TinyLfuRowCacheStrategy`. This follows the same convention used by `MemStore` and `RegionSplitPolicy` (`Configuration`-arg constructor + `ReflectionUtils.newInstance`).

Commit: https://github.com/apache/hbase/pull/7901/commits/6739f062accbde26671deaee01b59cfc1fb3d693

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
EungsopYoo: Done.

- Renamed `populateCache` to `cache`.
- Changed `regionLevelBarrierMap` key type from `HRegion` to the encoded region name (`String`). The encoded name is the canonical region identifier already used elsewhere (e.g., `RowCacheKey.isSameRegion`). As a result, `cache` no longer needs an `HRegion` parameter; it derives the encoded region name from the `RowCacheKey`.

I kept the external signatures of `create/remove/getRegionLevelBarrier` taking `HRegion` to keep the caller's intent explicit; only the internal map key type changes. Let me know if you'd prefer those to take `String` as well.

Commit: https://github.com/apache/hbase/pull/7901/commits/e985135e33694b93a8e210a48cdb4836fa95c305

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
EungsopYoo: Good point — turned out to be feasible. Done.

Quick verification: I was initially worried that `Get` results don't carry TTL tags (the design doc assumed this), but that's actually a *client-side* property — the RPC codec strips tags. **Server-side** cells delivered to `RowCache` preserve their TTL tag (carried forward by `TagUtil.carryForwardTTLTag` during mutation, the same tag `ScanQueryMatcher.isCellTTLExpired` reads). Confirmed with a probe test against a live region.

Implementation:

- `RowCells` now precomputes the earliest TTL expiration time across its cells during construction. Cells without a TTL tag yield `Long.MAX_VALUE`, so `RowCells.isExpired(now)` is an `O(1)` `long` comparison on every cache hit — no per-cell tag iteration on the hot path.
- `RowCache.tryGetFromCache`: if the row is expired, evict and fall back to the storage read path.
- `RowCache.cache`: skip caching when results are empty (otherwise an evicted-then-empty row would be re-cached as an empty entry).
- `HRegion.put`: removed the guard that rejected Puts with TTL.

CF-level TTL still disables the row cache via `canCacheRow`'s `isDefaultTtl` check; that policy is unchanged for now.

Commit: https://github.com/apache/hbase/pull/7901/commits/d8fef38bfc02906f0314397a588371be43fa6bf2

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
```diff
@@ -2003,6 +2014,22 @@ public Pair<byte[], Collection<HStoreFile>> call() throws IOException {
...
+    if (!evictOnClose) {
+      return;
+    }
+
+    if (!(rsServices instanceof HRegionServer regionServer)) {
+      return;
+    }
+
+    RowCache rowCache = regionServer.getRSRpcServices().getServer().getRowCache();
+    rowCache.evictRowsByRegion(this);
```
wchevreuil: Can we avoid cyclic dependency by changing evictRowsByRegion signature to accept the encodedRegionName instead?

wchevreuil (commented):
Overall, looks good, just a nit best practice comment. Also, to make this mergeable, can we move the unrelated dev-support and .github/workflows change out this PR?

EungsopYoo:
@wchevreuil Done — dropped the dev-support and .github/workflows commits.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegion.java
EungsopYoo: Done — switched to `String encodedRegionName` to avoid the cyclic dependency on `HRegion`.

Commit: https://github.com/apache/hbase/pull/7901/commits/8ff2fd82a68f2799fff800bb9bd16b6468eeecbc

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RowCache.java
EungsopYoo: Following up — also switched `createRegionLevelBarrier`, `removeRegionLevelBarrier`, and `getRegionLevelBarrier` to take `String encodedRegionName`.

Commit: https://github.com/apache/hbase/pull/7901/commits/361c8e8e934fcb3f773185a25529402fdeeb12e3

wchevreuil (approved):
LGTM, +1. @liuxiaocs7 Do you have any further suggestions?

PR #7933, opened 2026-03-13
HBASE-29878 Add GitHub Action workflow for building and publishing HBase website

PDavid:


Apache9:
Do we have enough permission to push the repo in github actions?

.github/workflows/website.yml
```diff
@@ -0,0 +1,86 @@
...
+# yamllint disable rule:line-length
+---
+name: Build and publish website
+
+"on":
+  # FIXME: Remove this PR here before merge!
+  pull_request:
+    types: [opened, synchronize, reopened]
+  schedule:
+  - cron: '0 14 * * *'
```
Apache9, line 27: Can github actions support something like pollSCM in jenkins? Where we will not run the pipeline if there is no change in git.

Apache9:
@PDavid I think we can implement a new jenkins job to publish the website, where we build the website content on hbase node, stash it, and then unstash it on website node, update the website repo and push it.

I could file an issue and try to implement it.

Thanks.

PDavid:
> Do we have enough permission to push the repo in github actions?

No, unfortunately we don't have enough permissions. Usually the workflow only have permission to push to the same repo it "runs in". 

> To push to a different repository, the standard GITHUB_TOKEN usually won't cut it because its permissions are scoped only to the current repo. You’ll need a Personal Access Token (PAT) or a GitHub App token with "Contents: Write" permissions for the target repository.

.github/workflows/website.yml
PDavid: Very good question. I'm not sure, maybe it can be done somehow. I'd have to check. But makes perfect sense. :+1:

PDavid:
> @PDavid I think we can implement a new jenkins job to publish the website, where we build the website content on hbase node, stash it, and then unstash it on website node, update the website repo and push it.
> 
> I could file an issue and try to implement it.
> 
> Thanks.

Thanks @Apache9, to be completely honest I don't have too much experience with Jenkins but what you suggest makes perfect sense to me. Just wanted to check if I can quickly use GitHub Actions instead of Jenkins.

PDavid:
Closing this as we introduced a new Jenkins job under https://issues.apache.org/jira/browse/HBASE-30004

PR #7934, opened 2026-03-13
HBASE-29995 Improve existing hash implementations by reading 4 bytes at once

jinhyukify:
**Jira** https://issues.apache.org/jira/browse/HBASE-29995

### DIFF
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/util/LittleEndianBytes.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/util/LittleEndianBytes.java
@@ -1,0 +1,169 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.util;
+
+import java.nio.ByteBuffer;
+import org.apache.hadoop.hbase.ByteBufferExtendedCell;
+import org.apache.hadoop.hbase.Cell;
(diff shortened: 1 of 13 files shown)


Apache9:
Please add a dummy change in pom.xml so we can run all the tests in HBase to see if there are any problems with the changes? once we confirm there is no problem, you can do a force push to remove the dummy commit and then merge the PR.

And why not read 8 bytes at once?

jinhyukify:
@Apache9 Thank you for checking!

> And why not read 8 bytes at once?

The existing hashes (Jenkins, Murmur, Murmur3) are all 4-byte based, so they operate on 4-byte chunks. There is currently no place in the algorithms where 8-byte reads are used.

In theory we could read 8 bytes and split them into two ints, but given the current structure it would not provide much benefit. If we add hash functions that operate on 8-byte words in the future, we could revisit this.

hbase-common/src/main/java/org/apache/hadoop/hbase/util/RowColBloomHashKey.java
```diff
@@ -80,9 +84,116 @@ public byte get(int offset) {
...
+    // Compute timestamp section range.
+    final int tsEnd = qualEnd + KeyValue.TIMESTAMP_SIZE;
+    if (offset >= qualEnd && offset + Bytes.SIZEOF_INT <= tsEnd) {
+      return LittleEndianBytes.toInt(LATEST_TS, offset - qualEnd);
+    }
+
+    return (int) assembleCrossingLE(offset, Bytes.SIZEOF_INT);
+  }
+
+  private long assembleCrossingLE(int offset, int wordBytes) {
```
junegunn, line 83: So we're adding a non-trivial amount of code here, and it seems to overlap with the existing `get(int offset)` implementation to some degree. Would it make sense to refactor `get` to call this new method instead, so we can reduce code duplication? We could consider that approach as long as the performance overhead is negligible.

jinhyukify: Thank you for checking.
I changed to use `assembleCrossingLE` method in get method.

hbase-common/src/main/java/org/apache/hadoop/hbase/util/LittleEndianBytes.java
```diff
@@ -0,0 +1,170 @@
...
+
+      @Override
+      int toInt(byte[] bytes, int offset) {
+        int n = 0;
+        for (int i = offset + 3; i >= offset; i--) {
+          n <<= 8;
+          n ^= (bytes[i] & 0xFF);
+        }
+        return n;
+      }
```
junegunn, line 68: How about making this implementation similar to the `toInt(ByteBuffer)` version? It should be a bit easier to understand.

jinhyukify: Sounds great. I updated the implementation.

junegunn (approved):

PR #7995, opened 2026-03-28
HBASE-30038: RefCnt Leak error when caching

dParikesit:
JIRA: [HBASE-30038](https://issues.apache.org/jira/browse/HBASE-30038)

This bug is similar to [HBASE-28890](https://issues.apache.org/jira/browse/HBASE-28890)

HFileBlock.Writer.getBlockForCaching() creates a ref-counted HFileBlock, and the caller must release its own reference after the cache has taken the ownership it needs. 

In HFileWriterImpl.java, the leak happened when shouldCacheBlock() returned before cacheFormatBlock.release() ran.

In HFileBlockIndex.java, the intermediate-index path cached blockForCaching but never released the reference at all. 

Because these blocks can be backed by pooled off-heap ByteBuffs, repeated HFile writes could steadily drain allocator buffers and effectively leak memory, even though the blocks were only meant to live long enough to be considered for cache-on-write.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFile.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFile.java
@@ -38,8 +38,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.Random;
 import java.util.concurrent.ThreadLocalRandom;
+import java.util.concurrent.atomic.AtomicInteger;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FSDataInputStream;
 import org.apache.hadoop.fs.FSDataOutputStream;
@@ -76,6 +78,7 @@ import org.apache.hadoop.hbase.io.hfile.HFile.Writer;
 import org.apache.hadoop.hbase.io.hfile.ReaderContext.ReaderType;
 import org.apache.hadoop.hbase.monitoring.ThreadLocalServerSideScanMetrics;
 import org.apache.hadoop.hbase.nio.ByteBuff;
+import org.apache.hadoop.hbase.nio.RefCnt;
 import org.apache.hadoop.hbase.regionserver.StoreFileWriter;
 import org.apache.hadoop.hbase.testclassification.IOTests;
 import org.apache.hadoop.hbase.testclassification.SmallTests;
@@ -92,6 +95,8 @@ import org.junit.rules.TestName;
 import org.mockito.Mockito;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
+
+import org.apache.hbase.thirdparty.io.netty.util.ResourceLeakDetector;
 
 /**
  * test hfile features.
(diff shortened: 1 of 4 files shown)


wchevreuil (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFile.java
```diff
@@ -202,6 +207,53 @@ public void testReaderWithLRUBlockCache() throws Exception {
...
+          HConstants.LATEST_TIMESTAMP, Bytes.toBytes("value")));
+      } finally {
+        writer.close();
+      }
+
+      Mockito.verify(cache).shouldCacheBlock(Mockito.any(), Mockito.anyLong(), Mockito.any());
+      Mockito.verify(cache, Mockito.never()).cacheBlock(Mockito.any(), Mockito.any(),
+        Mockito.anyBoolean(), Mockito.anyBoolean());
+      System.gc();
+      Thread.sleep(1000);
```
vaijosh: @dParikesit 
I think the 1-second assumption might bite us on slower systems and make the test flaky.
What if we wrap this in a loop instead? We could retry up to 15 times with a small delay—that way it succeeds as soon as it's ready rather than relying on luck.

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFileBlockIndex.java
```diff
@@ -216,6 +216,60 @@ public void onLeak(String s, String s1) {
...
+    CacheConfig cacheConfig = new CacheConfig(localConf, null, cache, allocator);
+    HFileBlockIndex.BlockIndexWriter biw =
+      new HFileBlockIndex.BlockIndexWriter(hbw, cacheConfig, localPath.getName(), null);
+    biw.setMaxChunkSize(512);
+
+    try {
+      writeDataBlocksAndCreateIndex(hbw, outputStream, biw);
+      assertTrue(biw.getNumLevels() >= 3);
+      System.gc();
+      Thread.sleep(1000);
```
vaijosh: @dParikesit 
I think the 1-second assumption might bite us on slower systems and make the test flaky.
What if we wrap this in a loop instead? We could retry up to 15 times with a small delay—that way it succeeds as soon as it's ready rather than relying on luck.

wchevreuil:
Please consider @vaijosh suggestions and also fix spotless issues.

dParikesit:
Thanks for the suggestions. I'll get back to you after I fix it.

pankaj72981 (approved):

wchevreuil:
Any news on this, @dParikesit ? Please get the reviews addressed so that we can move forward here, as this is an important fix you have found.

dParikesit:
@wchevreuil thanks for the reminder. I have modified the tests to use retry-loop instead of sleep (including the test added in [HBASE-28890](https://issues.apache.org/jira/browse/HBASE-28890) ). I've tested it on my machine. Can you help trigger the CI pipeline?

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFile.java
```diff
@@ -244,9 +244,10 @@ public void onLeak(String s, String s1) {
       Mockito.verify(cache).shouldCacheBlock(Mockito.any(), Mockito.anyLong(), Mockito.any());
       Mockito.verify(cache, Mockito.never()).cacheBlock(Mockito.any(), Mockito.any(),
         Mockito.anyBoolean(), Mockito.anyBoolean());
-      System.gc();
-      Thread.sleep(1000);
-      alloc.allocate(128 * 1024).release();
+      for (int i = 0; i < 15 && counter.get() == 0; i++) {
+        System.gc();
+        alloc.allocate(128 * 1024).release();
+      }
```
vaijosh, line 256: @dParikesit 
You have not added sleep inside the loop so it will finish quicky and deafeat the purpose.

Please keep it something like below 

for (int i = 0; i < 30 && counter.get() == 0; i++) {
    System.gc();
    alloc.allocate(128 * 1024).release();
    try {
        Thread.sleep(1000); 
    } catch (InterruptedException e) {
        Thread.currentThread().interrupt();
        break;
    }
}

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFileBlockIndex.java
```diff
@@ -258,9 +256,10 @@ public void onLeak(String s, String s1) {
     try {
       writeDataBlocksAndCreateIndex(hbw, outputStream, biw);
       assertTrue(biw.getNumLevels() >= 3);
-      System.gc();
-      Thread.sleep(1000);
-      allocator.allocate(128 * 1024).release();
+      for (int i = 0; i < 15 && counter.get() == 0; i++) {
+        System.gc();
+        allocator.allocate(128 * 1024).release();
+      }
```
vaijosh, line 274: @dParikesit 
Please add Put slip inside the loop and make 30 retries instead of 15

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestHFile.java
dParikesit: Thanks for the feedback! I believe the sleep should be done after we run allocate.release() right? I have updated the commit to change it.

vaijosh (approved):
Thanks @dParikesit.
Changes LTGM

PR #8013, opened 2026-03-31
HBASE-30049 RestoreSnapshotHelper creates StoreFileTracker with wrong…

bjomobo:
## Description

RestoreSnapshotHelper.restoreRegion() creates a StoreFileTracker using the raw Master conf which lacks table-level settings like `hbase.store.file-tracker.impl=FILE`. This causes DefaultStoreFileTracker to be used, whose `doSetStoreFiles()` is a no-op. The `.filelist` is never updated after restore, leading to `FileNotFoundException` when regions try to open files that were archived.

Regression introduced by HBASE-28564.

## Changes

- Merge table descriptor config via `StoreUtils.createStoreConfiguration()` before creating the tracker in `restoreRegion()`
- Move tracker creation inside the `snapshotFamilyFiles != null` check to avoid `NullPointerException` on families being removed
- Add `withColumnFamilyDescriptor()` to the "Add families not present in the table" code path

## New Tests

- `TestRestoreSnapshotProcedureFileBasedSFT` — end-to-end restore with FILE tracker
- `TestRestoreSnapshotHelperWithFileBasedSFT` — unit-level `.filelist` verification
- `TestRestoreSnapshotFileTrackerTableLevel` — table-level FILE tracker with compaction and multi-family restore

Jira: https://issues.apache.org/jira/browse/HBASE-30049


hbase-server/src/test/java/org/apache/hadoop/hbase/master/procedure/TestRestoreSnapshotProcedureFileBasedSFT.java
```diff
@@ -0,0 +1,268 @@
...
+  }
+
+  /**
+   * Test that reproduces the scenario where restore fails with FileNotFoundException:
+   * 1. Create table with FILE tracker, load data, snapshot
+   * 2. Load more data + flush (creates new HFiles not in snapshot)
+   * 3. Restore from snapshot
+   * 4. Verify all regions open and data matches the snapshot
+   *
+   * Before the fix, step 4 would fail with FileNotFoundException because
```
gvprathyusha6: this test testRestoreSnapshotWithFileTrackerAfterDataChange is still working without the fix as global SFT config itself is FILE, but this issue happens when global is Default and Table config is FILE like here - https://github.com/apache/hbase/pull/8013/changes#diff-35d1d555129ce54d8d45094be2f76300b7cd9d29c9a109ff95d81c626b428939R77

hbase-server/src/test/java/org/apache/hadoop/hbase/master/procedure/TestRestoreSnapshotProcedureFileBasedSFT.java
```diff
@@ -0,0 +1,268 @@
...
+    }
+  }
+
+  /**
+   * Test restore after compaction, compaction creates new HFiles that replace
+   * the originals. After restore, the .filelist must point to the snapshot's
+   * HFiles, not the compaction output.
+   */
+  @Test
+  public void testRestoreSnapshotAfterCompaction() throws Exception {
```
gvprathyusha6, line 171: this as well, since the global SFT config itself is FILE, this should work, but looks like an issue with the test, it fails at this assert https://github.com/apache/hbase/pull/8013/changes#diff-66c1760bfcfc64f561b37dd18df6a41595e93d1c516a326de27ad6e9c631ebb2R266 
it should be failing [here](https://github.com/apache/hbase/pull/8013/changes#diff-66c1760bfcfc64f561b37dd18df6a41595e93d1c516a326de27ad6e9c631ebb2R253) instead?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/procedure/TestRestoreSnapshotProcedureFileBasedSFT.java
bjomobo: thanks for pointing this out. removed the global TRACKER_IMPL=FILE setting from setupCluster() and now setting it at the table descriptor level only.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/procedure/TestRestoreSnapshotProcedureFileBasedSFT.java
```diff
@@ -0,0 +1,271 @@
...
+    UTIL.startMiniCluster(3);
+  }
+
+  @AfterClass
+  public static void tearDownCluster() throws Exception {
+    UTIL.shutdownMiniCluster();
+  }
+
+  /**
+   * Test that reproduces the scenario where restore fails with FileNotFoundException: 1. Create
```
gvprathyusha6, line 90: Looks like this test and TestRestoreSnapshotFileTrackerTableLevel.java have some overlapping test scenarios, can you check once, or am I missing something here?

bjomobo: `TestRestoreSnapshotProcedureFileBasedSFT` tests the end-to-end procedure path (single region, simple data change + compaction scenarios), while `TestRestoreSnapshotFileTrackerTableLevel` uses pre-split regions (4 regions), proper compaction wait via `UTIL.waitFor()`, and covers the multi-family restore path (add/remove column families). Happy to consolidate if you feel strongly, but I kept them separate since they exercise slightly different things

gvprathyusha6: admin.restoreSnapshot(snapshotName); in TestRestoreSnapshotFileTrackerTableLevel also creates procedure know? but its a nit only

gvprathyusha6 (approved):
LGTM

hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/RestoreSnapshotHelper.java
```diff
@@ -504,17 +504,19 @@ private void restoreRegion(final RegionInfo regionInfo,
...
-      List<String> familyFiles = storeFileInfos.stream()
-        .map(storeFileInfo -> storeFileInfo.getPath().getName()).collect(Collectors.toList());
+
+      Set<String> familyFiles = getTableRegionFamilyFiles(familyDir);
       List<SnapshotRegionManifest.StoreFile> snapshotFamilyFiles =
         snapshotFiles.remove(familyDir.getName());
-      List<StoreFileInfo> filesToTrack = new ArrayList<>();
       if (snapshotFamilyFiles != null) {
+        // Family exists in the snapshot, create tracker with merged table descriptor config
+        // so that table-level settings are picked up.
```
wchevreuil: nit: Let's make clear that tableDesc here means the one recovered from the snapshot manifest. This means if current SFT config in the table is different from the snapshot one, it will be overwritten.

wchevreuil (approved):

wchevreuil:
@bjomobo , looks like this needs some code style polishing:


❌ | mvninstall | 1m 28s | root in the patch failed.
-- | -- | -- | --
❌ | compile | 0m 36s | hbase-server in the patch failed.
⚠️ | javac | 0m 36s | hbase-server in the patch failed.
✅ | blanks | 0m 0s | The patch has no blanks issues.
⚠️ | checkstyle | 0m 47s | hbase-server: The patch generated 2 new + 7 unchanged - 0 fixed = 9 total (was 7)
❌ | spotbugs | 0m 36s | hbase-server in the patch failed.
❌ | hadoopcheck | 1m 40s | The patch causes 20 errors with Hadoop v3.3.6.
❌ | hadoopcheck | 3m 14s | The patch causes 20 errors with Hadoop v3.4.2.
❌ | spotless | 0m 32s | patch has 29 errors when running spotless:check, run spotless:apply to fix.

Can you check on the spotless, checkstyles and spotbugs errors?

bjomobo:
@wchevreuil addressed the style issues and merge conflict, can you take a look

hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/RestoreSnapshotHelper.java
bjomobo: done

PR #6931, opened 2025-04-23
HBASE-29236: Add Support for Dynamic Configuration at the Coprocessor Level

kgeisz:
https://issues.apache.org/jira/browse/HBASE-29236

This pull request adds support for Dynamic Configuration at the coprocessor level and makes the `hbase.global.readonly.enabled` config variable dynamically configurable.  This variable is used with the ReadOnlyController coprocessor.  In order to make this class dynamically configurable, we have it implement the ConfigurationObserver interface.  We also need to register this observer with the ConfigurationManager, so the ReadOnlyController is registered to HMaster's, HRegionServer's, and HRegion's ConfigurationManager.  Also, since the ReadOnlyController is a coprocessor, we need to use the CoprocessorHost in order to get the coprocessor and register it with the ConfigurationManager.

`hbase.global.readonly.enabled` can be configured dynamically in the following way:

1. Add `hbase.global.readonly.enabled` to hbase-site.xml:

```
<property>
  <name>hbase.global.readonly.enabled</name>
  <value>false</value>
</property>
```

2. Add the ReadOnlyController coprocessor to the master and region server by adding the following to hbase-site.xml:

```
<property>
  <name>hbase.coprocessor.master.classes</name>
  <value>org.apache.hadoop.hbase.security.access.ReadOnlyController</value>
</property>
<property>
  <name>hbase.coprocessor.regionserver.classes</name>
  <value>org.apache.hadoop.hbase.security.access.ReadOnlyController</value>
</property>
<property>
  <name>hbase.coprocessor.region.classes</name>
  <value>org.apache.hadoop.hbase.security.access.ReadOnlyController</value>
</property>
```

3. Start HBase
4. Change `hbase.global.readonly.enabled` from `false` to `true` in hbase-site.xml and save the file.
5. In the HBase shell, run `update_all_config`.
6. You should see the following log message for the master, region servers(s), and region(s):

```
2025-04-23T15:23:59,008 INFO  [RpcServer.priority.RWQ.Fifo.read.handler=3,queue=1,port=16000] access.ReadOnlyController: Config hbase.global.readonly.enabled has been dynamically changed to true
```

PR #8044, opened 2026-04-08
HBASE-29081: Add HBase Read Replica Cluster feature

anmolnar:
Hi all,

We would like to propose merging the feature “Read Replica Cluster” into 
the main branch.

**Background**

We’d like to implement the open source version of Amazon’s [Read Replica 
Cluster on S3](https://aws.amazon.com/blogs/big-data/setting-up-read-replica-clusters-with-hbase-on-amazon-s3/) feature for Apache HBase. It adds the ability of running 
another HBase cluster on the same cloud storage location in read-only mode, 
allowing users to share the read workload between multiple clusters. Due 
to the characteristics of the implementation and the lack of automated 
synchronization between the active and read-replica clusters, read replicas 
are eventually consistent, hence they’re not suitable for reading most 
recent data. However we still believe that users of open source Apache HBase 
could take advantage of this feature and there are use cases out there which 
read replicas could help with. Please find more information about the 
feature in the linked blog post.

**Pros**

- Running multiple clusters in different Availability Zones adds HA to the 
entire workload,
- No need for data movement or duplication (active-active replication setup) 
which is cost and time efficient,
- No limit for the number of read replica clusters

**Cons**

- Read Replica clusters are eventually consistent: in memory data is not 
visible from read replicas,
- Read Replica clusters must be manually refreshed: flush on active cluster, 
refresh hfiles/meta on read replicas

A detailed description of the design and implementation can be found in the
following document.

[Apache HBase Read Replica Cluster Feature](https://docs.google.com/document/d/1EI0lsURX1BZhv3DYgMvZCl4EUy-ADJRkHUc1PjzZtj0/edit?usp=sharing)

Please review and share your feedback or comments.

Best regards,
Andor

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/ActiveClusterSuffix.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/ActiveClusterSuffix.java
@@ -1,0 +1,142 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase;
+
+import java.io.IOException;
+import java.util.Objects;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.exceptions.DeserializationException;
+import org.apache.hadoop.hbase.util.Bytes;
+import org.apache.yetus.audience.InterfaceAudience;
+
+import org.apache.hbase.thirdparty.com.google.common.base.Strings;
+
+import org.apache.hadoop.hbase.shaded.protobuf.ProtobufUtil;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.ActiveClusterSuffixProtos;
+
+/**
+ * The read-replica cluster id for this cluster. It is serialized to the filesystem and up into
+ * zookeeper. This is a container for the id. Also knows how to serialize and deserialize the
+ * cluster id.
+ */
+@InterfaceAudience.Private
+public class ActiveClusterSuffix implements ClusterIdFile {
+  private final String clusterId;
+  private final String suffix;
+
+  public static class Parser implements ClusterIdFileParser<ActiveClusterSuffix> {
+
+    @Override
+    public String getFileName() {
+      return HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME;
+    }
+
+    /**
+     * Parse the serialized representation of the {@link ActiveClusterSuffix}
+     * @param bytes A pb serialized {@link ActiveClusterSuffix} instance with pb magic prefix
+     * @return An instance of {@link ActiveClusterSuffix} made from <code>bytes</code>
+     * @see #toByteArray()
+     */
+    @Override
+    public ActiveClusterSuffix parseFrom(byte[] bytes) throws DeserializationException {
+      if (ProtobufUtil.isPBMagicPrefix(bytes)) {
+        int pblen = ProtobufUtil.lengthOfPBMagic();
+        ActiveClusterSuffixProtos.ActiveClusterSuffix.Builder builder =
+          ActiveClusterSuffixProtos.ActiveClusterSuffix.newBuilder();
+        ActiveClusterSuffixProtos.ActiveClusterSuffix cs = null;
+        try {
+          ProtobufUtil.mergeFrom(builder, bytes, pblen, bytes.length - pblen);
+          cs = builder.build();
+        } catch (IOException e) {
+          throw new DeserializationException(e);
+        }
+        return convert(cs);
+      } else {
+        // Presume it was written out this way, the old way.
+        return new ActiveClusterSuffix(Bytes.toString(bytes));
+      }
+    }
+
+    @Override
+    public ActiveClusterSuffix readString(String input) {
+      return new ActiveClusterSuffix(input);
+    }
+  }
+
+  public ActiveClusterSuffix(final String clusterId, final String suffix) {
+    this.clusterId = clusterId;
+    this.suffix = suffix;
+  }
+
+  public ActiveClusterSuffix(final String input) {
+    String[] parts = input.split(":", 2);
+    this.clusterId = parts[0];
+    if (parts.length > 1) {
+      this.suffix = parts[1];
+    } else {
+      this.suffix = "";
+    }
+  }
+
+  public static ActiveClusterSuffix parseFrom(byte[] bytes) throws DeserializationException {
+    return new Parser().parseFrom(bytes);
+  }
+
+  public static ActiveClusterSuffix fromConfig(Configuration conf, ClusterId clusterId) {
+    return new ActiveClusterSuffix(clusterId.toString(), conf
+      .get(HConstants.HBASE_META_TABLE_SUFFIX, HConstants.HBASE_META_TABLE_SUFFIX_DEFAULT_VALUE));
+  }
+
+  /** Returns The active cluster suffix serialized using pb w/ pb magic prefix */
+  public byte[] toByteArray() {
+    return ProtobufUtil.prependPBMagic(convert().toByteArray());
+  }
+
+  /** Returns A pb instance to represent this instance. */
+  public ActiveClusterSuffixProtos.ActiveClusterSuffix convert() {
+    return ActiveClusterSuffixProtos.ActiveClusterSuffix.newBuilder().setClusterId(clusterId)
+      .setSuffix(suffix).build();
+  }
+
+  /** Returns A {@link ActiveClusterSuffix} made from the passed in <code>cs</code> */
+  public static ActiveClusterSuffix
+    convert(final ActiveClusterSuffixProtos.ActiveClusterSuffix cs) {
+    return new ActiveClusterSuffix(cs.getClusterId(), cs.getSuffix());
+  }
+
+  /**
+   * @see java.lang.Object#toString()
+   */
+  @Override
+  public String toString() {
+    return String.format("%s:%s", this.clusterId,
+      Strings.isNullOrEmpty(this.suffix) ? "<blank>" : this.suffix);
+  }
+
+  @Override
+  public boolean equals(Object o) {
+    if (o == null || getClass() != o.getClass()) return false;
+    ActiveClusterSuffix that = (ActiveClusterSuffix) o;
+    return Objects.equals(clusterId, that.clusterId) && Objects.equals(suffix, that.suffix);
+  }
+
+  @Override
+  public int hashCode() {
+    return Objects.hash(clusterId, suffix);
+  }
+}
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AdminOverAsyncAdmin.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/client/AdminOverAsyncAdmin.java
@@ -1154,6 +1154,26 @@ class AdminOverAsyncAdmin implements Admin {
   }
 
   @Override
+  public long refreshMeta() throws IOException {
+    return get(admin.refreshMeta());
+  }
+
+  @Override
+  public long refreshHFiles(final TableName tableName) throws IOException {
+    return get(admin.refreshHFiles(tableName));
+  }
+
+  @Override
+  public long refreshHFiles(final String namespace) throws IOException {
+    return get(admin.refreshHFiles(namespace));
+  }
+
+  @Override
+  public long refreshHFiles() throws IOException {
+    return get(admin.refreshHFiles());
+  }
+
+  @Override
   public void restoreBackupSystemTable(String snapshotName) throws IOException {
     get(admin.restoreBackupSystemTable(snapshotName));
   }
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
@@ -1665,6 +1665,29 @@ public final class HConstants {
    */
   public final static boolean REJECT_DECOMMISSIONED_HOSTS_DEFAULT = false;
 
+  /**
+   * Adds a suffix to the meta table name: value='test' -> ‘hbase:meta_test’
+   */
+  public final static String HBASE_META_TABLE_SUFFIX = "hbase.meta.table.suffix";
+
+  /**
+   * Default value of {@link #HBASE_META_TABLE_SUFFIX}
+   */
+  public final static String HBASE_META_TABLE_SUFFIX_DEFAULT_VALUE = "";
+
+  /**
+   * Should HBase only serve Read Requests
+   */
+  public final static String HBASE_GLOBAL_READONLY_ENABLED_KEY = "hbase.global.readonly.enabled";
+
+  /**
+   * Default value of {@link #HBASE_GLOBAL_READONLY_ENABLED_KEY}
+   */
+  public final static boolean HBASE_GLOBAL_READONLY_ENABLED_DEFAULT = false;
+
+  /** name of the file having active cluster suffix */
+  public static final String ACTIVE_CLUSTER_SUFFIX_FILE_NAME = "active.cluster.suffix.id";
+
   private HConstants() {
     // Can't be instantiated with this ctor.
   }
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
@@ -17,16 +17,21 @@
  */
 package org.apache.hadoop.hbase;
 
+import com.google.errorprone.annotations.RestrictedApi;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Arrays;
 import java.util.Set;
 import java.util.concurrent.CopyOnWriteArraySet;
 import org.apache.commons.lang3.ArrayUtils;
+import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.apache.hbase.thirdparty.com.google.common.base.Preconditions;
+import org.apache.hbase.thirdparty.com.google.common.base.Strings;
 
 /**
  * Immutable POJO class for representing a table name. Which is of the form: &lt;table
@@ -44,6 +49,7 @@ import org.apache.hbase.thirdparty.com.google.common.base.Preconditions;
  */
 @InterfaceAudience.Public
 public final class TableName implements Comparable<TableName> {
+  private static final Logger LOG = LoggerFactory.getLogger(TableName.class);
 
   /** See {@link #createTableNameIfNecessary(ByteBuffer, ByteBuffer)} */
   private static final Set<TableName> tableCache = new CopyOnWriteArraySet<>();
@@ -64,10 +70,42 @@ public final class TableName implements Comparable<TableName> {
   // with NAMESPACE_DELIM as delimiter
   public static final String VALID_USER_TABLE_REGEX = "(?:(?:(?:" + VALID_NAMESPACE_REGEX + "\\"
     + NAMESPACE_DELIM + ")?)" + "(?:" + VALID_TABLE_QUALIFIER_REGEX + "))";
+  public static final String VALID_META_TABLE_SUFFIX_REGEX = "[a-zA-Z0-9]+";
 
-  /** The hbase:meta table's name. */
-  public static final TableName META_TABLE_NAME =
-    valueOf(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR, "meta");
+  /**
+   * The name of hbase meta table could either be hbase:meta_xxx or 'hbase:meta' otherwise. Config
+   * hbase.meta.table.suffix will govern the decision of adding suffix to the habase:meta
+   */
+  public static final TableName META_TABLE_NAME;
+  static {
+    Configuration conf = HBaseConfiguration.create();
+    META_TABLE_NAME = initializeHbaseMetaTableName(conf);
+    LOG.info("Meta table name: {}", META_TABLE_NAME);
+  }
+
+  /* Visible for testing only */
+  @RestrictedApi(explanation = "Should only be called in tests", link = "",
+      allowedOnPath = ".*/src/test/.*")
+  public static TableName getDefaultNameOfMetaForReplica() {
+    return valueOf(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR, "meta");
+  }
+
+  public static TableName initializeHbaseMetaTableName(Configuration conf) {
+    String suffix_val = conf.get(HConstants.HBASE_META_TABLE_SUFFIX,
+      HConstants.HBASE_META_TABLE_SUFFIX_DEFAULT_VALUE);
+    LOG.debug("[Read-replica feature] suffix value: {}",
+      (suffix_val == null || suffix_val.isEmpty()) ? "<blank>" : suffix_val);
+    if (Strings.isNullOrEmpty(suffix_val)) {
+      return valueOf(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR, "meta");
+    } else {
+      if (!suffix_val.matches(VALID_META_TABLE_SUFFIX_REGEX)) {
+        throw new IllegalArgumentException("Invalid value '" + suffix_val + "' for config '"
+          + HConstants.HBASE_META_TABLE_SUFFIX + "'. Suffix must only contain ASCII letters and "
+          + "digits matching: " + VALID_META_TABLE_SUFFIX_REGEX);
+      }
+      return valueOf(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR, "meta_" + suffix_val);
+    }
+  }
 
   /**
    * The Namespace table's name.
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
@@ -84,6 +84,7 @@ import org.apache.hadoop.hbase.InvalidFamilyOperationException;
 import org.apache.hadoop.hbase.MasterNotRunningException;
 import org.apache.hadoop.hbase.MetaTableAccessor;
 import org.apache.hadoop.hbase.NamespaceDescriptor;
+import org.apache.hadoop.hbase.NamespaceNotFoundException;
 import org.apache.hadoop.hbase.PleaseHoldException;
 import org.apache.hadoop.hbase.PleaseRestartMasterException;
 import org.apache.hadoop.hbase.RegionMetrics;
@@ -170,6 +171,8 @@ import org.apache.hadoop.hbase.master.procedure.ModifyTableProcedure;
 import org.apache.hadoop.hbase.master.procedure.ProcedurePrepareLatch;
 import org.apache.hadoop.hbase.master.procedure.ProcedureSyncWait;
 import org.apache.hadoop.hbase.master.procedure.RSProcedureDispatcher;
+import org.apache.hadoop.hbase.master.procedure.RefreshHFilesTableProcedure;
+import org.apache.hadoop.hbase.master.procedure.RefreshMetaProcedure;
 import org.apache.hadoop.hbase.master.procedure.ReloadQuotasProcedure;
 import org.apache.hadoop.hbase.master.procedure.ReopenTableRegionsProcedure;
 import org.apache.hadoop.hbase.master.procedure.ServerCrashProcedure;
@@ -248,10 +251,12 @@ import org.apache.hadoop.hbase.security.AccessDeniedException;
 import org.apache.hadoop.hbase.security.SecurityConstants;
 import org.apache.hadoop.hbase.security.Superusers;
 import org.apache.hadoop.hbase.security.UserProvider;
+import org.apache.hadoop.hbase.security.access.AbstractReadOnlyController;
 import org.apache.hadoop.hbase.trace.TraceUtil;
 import org.apache.hadoop.hbase.util.Addressing;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.hbase.util.CommonFSUtils;
+import org.apache.hadoop.hbase.util.ConfigurationUtil;
 import org.apache.hadoop.hbase.util.CoprocessorConfigurationUtil;
 import org.apache.hadoop.hbase.util.DNS;
 import org.apache.hadoop.hbase.util.EnvironmentEdgeManager;
@@ -1086,7 +1091,11 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
     if (!maintenanceMode) {
       startupTaskGroup.addTask("Initializing master coprocessors");
       setQuotasObserver(conf);
-      this.cpHost = new MasterCoprocessorHost(this, conf);
+      CoprocessorConfigurationUtil.syncReadOnlyConfigurations(conf,
+        CoprocessorHost.MASTER_COPROCESSOR_CONF_KEY);
+      AbstractReadOnlyController.manageActiveClusterIdFile(
+        ConfigurationUtil.isReadOnlyModeEnabledInConf(conf), this.getMasterFileSystem());
+      initializeCoprocessorHost(conf);
     } else {
       // start an in process region server for carrying system regions
       maintenanceRegionServer =
@@ -1188,8 +1197,9 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
         int existingReplicasCount =
           assignmentManager.getRegionStates().getRegionsOfTable(TableName.META_TABLE_NAME).size();
         if (existingReplicasCount > metaDesc.getRegionReplication()) {
-          LOG.info("Update replica count of hbase:meta from {}(in TableDescriptor)"
-            + " to {}(existing ZNodes)", metaDesc.getRegionReplication(), existingReplicasCount);
+          LOG.info(
+            "Update replica count of {} from {}(in TableDescriptor)" + " to {}(existing ZNodes)",
+            TableName.META_TABLE_NAME, metaDesc.getRegionReplication(), existingReplicasCount);
           metaDesc = TableDescriptorBuilder.newBuilder(metaDesc)
             .setRegionReplication(existingReplicasCount).build();
           tableDescriptors.update(metaDesc);
@@ -1198,8 +1208,9 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
         if (metaDesc.getRegionReplication() != replicasNumInConf) {
           LOG.info(
             "The {} config is {} while the replica count in TableDescriptor is {}"
-              + " for hbase:meta, altering...",
-            HConstants.META_REPLICAS_NUM, replicasNumInConf, metaDesc.getRegionReplication());
+              + " for {}, altering...",
+            HConstants.META_REPLICAS_NUM, replicasNumInConf, metaDesc.getRegionReplication(),
+            TableName.META_TABLE_NAME);
           procedureExecutor.submitProcedure(new ModifyTableProcedure(
             procedureExecutor.getEnvironment(), TableDescriptorBuilder.newBuilder(metaDesc)
               .setRegionReplication(replicasNumInConf).build(),
@@ -3135,8 +3146,8 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
           if (isActiveMaster() && isInitialized() && assignmentManager != null) {
             try {
               Map<TableName, RegionStatesCount> tableRegionStatesCountMap = new HashMap<>();
-              Map<String, TableDescriptor> tableDescriptorMap = getTableDescriptors().getAll();
-              for (TableDescriptor tableDescriptor : tableDescriptorMap.values()) {
+              List<TableDescriptor> tableDescriptors = listTableDescriptors(null, null, null, true);
+              for (TableDescriptor tableDescriptor : tableDescriptors) {
                 TableName tableName = tableDescriptor.getTableName();
                 RegionStatesCount regionStatesCount =
                   assignmentManager.getRegionStatesCount(tableName);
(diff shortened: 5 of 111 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/AbstractReadOnlyController.java
```diff
@@ -0,0 +1,123 @@
...
+    isWritableInReadOnlyMode(final ObserverContext<? extends RegionCoprocessorEnvironment> c) {
+    return writableTables.contains(c.getEnvironment().getRegionInfo().getTable());
+  }
+
+  public static boolean isWritableInReadOnlyMode(final TableName tableName) {
+    return writableTables.contains(tableName);
+  }
+
+  protected void internalReadOnlyGuard() throws DoNotRetryIOException {
+    throw new DoNotRetryIOException("Operation not allowed in Read-Only Mode");
```
charlesconnell: I think it would be nice to subclass DoNotRetryIOException with something more specific to this situation, like `WriteAttemptedOnReadOnlyClusterException`

hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
```diff
@@ -64,10 +70,42 @@ public final class TableName implements Comparable<TableName> {
...
-  /** The hbase:meta table's name. */
-  public static final TableName META_TABLE_NAME =
-    valueOf(NamespaceDescriptor.SYSTEM_NAMESPACE_NAME_STR, "meta");
+  /**
+   * The name of hbase meta table could either be hbase:meta_xxx or 'hbase:meta' otherwise. Config
+   * hbase.meta.table.suffix will govern the decision of adding suffix to the habase:meta
+   */
+  public static final TableName META_TABLE_NAME;
+  static {
+    Configuration conf = HBaseConfiguration.create();
```
charlesconnell, line 81: When possible I try to avoid using `HBaseConfiguration.create()` because it's so inflexible about what config files/values it will read. How necessary is it to load the meta table name at static initialization time?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RefreshHFilesCallable.java
```diff
@@ -0,0 +1,70 @@
...
+    HRegion region = rs.getRegion(regionInfo.getEncodedName());
+    LOG.debug("Starting refreshHfiles operation on region {}", region);
+
+    try {
+      for (Store store : region.getStores()) {
+        store.refreshStoreFiles();
+      }
+    } catch (IOException ioe) {
+      LOG.warn("Exception while trying to refresh store files: ", ioe);
+    }
```
charlesconnell: Could you talk about the decision to swallow the error here? I'm on the fence if that is the right choice.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/RefreshMetaProcedure.java
```diff
@@ -0,0 +1,480 @@
...
+    LOG.info("Scanned backing storage and found {} regions", regions.size());
+    return regions;
+  }
+
+  private List<RegionInfo> scanTablesInNamespace(FileSystem fs, Path namespacePath)
+    throws IOException {
+    LOG.debug("Scanning namespace {}", namespacePath.getName());
+    List<Path> tableDirs = FSUtils.getLocalTableDirs(fs, namespacePath);
+
+    return tableDirs.parallelStream().flatMap(tableDir -> {
```
charlesconnell: Delegating into the common ForkJoinPool feels dicey here. I would feel safer if this was a regular stream().

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
```diff
@@ -4483,13 +4505,17 @@ public void onConfigurationChange(Configuration newConf) {
     }
     // append the quotas observer back to the master coprocessor key
     setQuotasObserver(newConf);
-    // update region server coprocessor if the configuration has changed.
-    if (
-      CoprocessorConfigurationUtil.checkConfigurationChange(this.cpHost, newConf,
-        CoprocessorHost.MASTER_COPROCESSOR_CONF_KEY) && !maintenanceMode
-    ) {
-      LOG.info("Update the master coprocessor(s) because the configuration has changed");
```
charlesconnell, line 4497: Might be nice to keep this logging?

hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
```diff
@@ -102,4 +116,129 @@ private static boolean hasCoprocessorsConfigured(Configuration conf, String... c
...
+   * @param reloadTask                lambda function that reloads coprocessors on the master,
+   *                                  region server, or region
+   */
+  public static void maybeUpdateCoprocessors(Configuration newConf,
+    boolean originalIsReadOnlyEnabled, CoprocessorHost<?, ?> coprocessorHost,
+    String coprocessorConfKey, boolean isMaintenanceMode, String instance,
+    Consumer<Boolean> stateSetter, CoprocessorReloadTask reloadTask) {
+
+    boolean maybeUpdatedReadOnlyMode = ConfigurationUtil.isReadOnlyModeEnabled(newConf);
+    boolean hasReadOnlyModeChanged = originalIsReadOnlyEnabled != maybeUpdatedReadOnlyMode;
```
charlesconnell, line 273: I think that if this method/class must know whether the read-only mode has changed, it should track that itself, and not depend on the caller to help track it. Maybe you could look at what coprocessors are currently loaded to find out whether read-only mode is enabled without explicitly tracking it.

hbase-server/src/main/java/org/apache/hadoop/hbase/coprocessor/CoprocessorHost.java
```diff
@@ -130,6 +132,39 @@ public Set<String> getCoprocessorClassNames() {
...
+   */
+  public void deregisterConfigurationObservers(ConfigurationManager configurationManager) {
+    Coprocessor foundCp;
+    Set<String> coprocessors = this.getCoprocessors();
+    for (String cp : coprocessors) {
+      foundCp = this.findCoprocessor(cp);
+      if (foundCp instanceof ConfigurationObserver) {
+        configurationManager.deregisterObserver((ConfigurationObserver) foundCp);
+      }
+    }
```
charlesconnell: It looks like none of your coprocessors implement ConfigurationObserver. Was this meant as speculative infrastructure, or left in by accident?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
```diff
@@ -4483,13 +4505,17 @@ public void onConfigurationChange(Configuration newConf) {
...
-      this.cpHost = new MasterCoprocessorHost(this, newConf);
+
+    boolean originalIsReadOnlyEnabled = this.isGlobalReadOnlyEnabled;
+
+    CoprocessorConfigurationUtil.maybeUpdateCoprocessors(newConf, this.isGlobalReadOnlyEnabled,
+      this.cpHost, CoprocessorHost.MASTER_COPROCESSOR_CONF_KEY, this.maintenanceMode,
+      this.toString(), val -> this.isGlobalReadOnlyEnabled = val,
+      conf -> initializeCoprocessorHost(newConf));
```
charlesconnell: Use the captured `conf`

hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/AbstractReadOnlyController.java
anmolnar: Makes sense to me. cc @sharmaar12

hbase-common/src/main/java/org/apache/hadoop/hbase/TableName.java
anmolnar: We already have a **[refactoring patch](https://github.com/apache/hbase/pull/7730)** open to make the meta table name non-static. Until that I think static initialization is a must.

hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/AbstractReadOnlyController.java
sharmaar12: Noted. Will update code accordingly.

anmolnar:
I have no idea what's wrong with CI builds. Code builds completely fine on my machine with the same JDK version and tests are also passing.

charlesconnell:
That will be fixed in the next release of Yetus by https://github.com/apache/yetus/pull/365

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RefreshHFilesCallable.java
anmolnar: I don't think it's needed either. @sharmaar12 could you please comment?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/RefreshMetaProcedure.java
anmolnar: Is that because of filesystem operations?
We already use parallelStream() in `ProtobufUtil.java`.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
anmolnar: Yeah, I think I agree.

hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
anmolnar: It's just a static method in an utility class, it's not stateful. Though we could make it stateful and instantiate in each caller which would allow us to encapsulate tracking of read-only mode. 

Checking the loaded coprocessors might also work, since we already have a `getCoprocessorsFromConfig()` method here.

hbase-server/src/main/java/org/apache/hadoop/hbase/coprocessor/CoprocessorHost.java
anmolnar: The above mentioned callers (HMaster, HRegion, etc.) are the configuration observers. Since the coprocessors don't have to track the read-only mode flag - they're either loaded or not loaded - they don't have to observe the configuration change.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/RefreshHFilesCallable.java
sharmaar12: I agree we should send the exception to the user in case of failure. Will be resolved under [HBASE-30080](https://issues.apache.org/jira/browse/HBASE-30080).

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestCanStartHBaseInReadOnlyMode.java
```diff
@@ -0,0 +1,67 @@
...
+import org.apache.hadoop.hbase.*;
+import org.apache.hadoop.hbase.client.*;
+import org.apache.hadoop.hbase.testclassification.LargeTests;
+import org.apache.hadoop.hbase.testclassification.SecurityTests;
+import org.junit.*;
+import org.junit.experimental.categories.Category;
+import org.junit.rules.TestName;
+
+@Category({ SecurityTests.class, LargeTests.class })
+@SuppressWarnings("deprecation")
```
taklwu, line 31: nit: should we use tag of junit 5 before merging ? you may need to fix it for other classes as well. but if we decide to fix it later, please file JIRA and link it with https://issues.apache.org/jira/browse/HBASE-23671 ? 
```suggestion
@Tag(SecurityTests.TAG)
@Tag(LargeTests.TAG)
```

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestCanStartHBaseInReadOnlyMode.java
```diff
@@ -0,0 +1,67 @@
...
+public class TestCanStartHBaseInReadOnlyMode {
+
+  @ClassRule
```
taklwu, line 34: nit: remove if we decide to use junit 5

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
```diff
@@ -1665,6 +1665,31 @@ public enum OperationStatusCode {
    */
   public final static boolean REJECT_DECOMMISSIONED_HOSTS_DEFAULT = false;
 
+  /**
+   * Adds a suffix to the meta table name: value=’test’ -> ‘hbase:meta_test’ Added in HBASE-XXXXX to
```
taklwu, line 1669: what should `HBASE-XXXXX` refer ? should we update this comment before merging ?

hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/AbstractReadOnlyController.java
```diff
@@ -0,0 +1,123 @@
...
+  }
+
+  public static void manageActiveClusterIdFile(boolean readOnlyEnabled, MasterFileSystem mfs) {
+    FileSystem fs = mfs.getFileSystem();
+    Path rootDir = mfs.getRootDir();
+    Path activeClusterFile = new Path(rootDir, HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME);
+
+    try {
+      if (readOnlyEnabled) {
+        // ENABLING READ-ONLY (false -> true), delete the active cluster file.
```
taklwu, line 78: nit: Would the below fast return be better ? 

```suggestion
    if (!readOnlyEnabled) {
      return;
    }
    FileSystem fs = mfs.getFileSystem();
    Path rootDir = mfs.getRootDir();
    Path activeClusterFile = new Path(rootDir, HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME);

    try {
        // ENABLING READ-ONLY (false -> true), delete the active cluster file.
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
```diff
@@ -4483,13 +4505,17 @@ public void onConfigurationChange(Configuration newConf) {
...
+
+    if (this.isGlobalReadOnlyEnabled != originalIsReadOnlyEnabled) {
```
taklwu, line 4516: nit: if we don't need to manage the active cluster Id for the non-readonly cluster, maybe we should skip early.

```suggestion
    if (this.isGlobalReadOnlyEnabled != originalIsReadOnlyEnabled && this.isGlobalReadOnlyEnabled) {
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/procedure/RefreshMetaProcedure.java
Kota-SH: Thanks for pointing it out @charlesconnell.
The parallelStream() here runs FS work (unlike ProtobufUtil) on the common ForkJoinPool, which can put pressure on other worker threads that use same pool. This can be done using regular stream() for now and optimize it for Cloud Storage if needed after testing perf.
This is being handled in [HBASE-30081](https://issues.apache.org/jira/browse/HBASE-30081)

hbase-client/src/main/java/org/apache/hadoop/hbase/ActiveClusterSuffix.java
```diff
@@ -0,0 +1,142 @@
...
+import org.apache.hadoop.hbase.shaded.protobuf.generated.ActiveClusterSuffixProtos;
+
+/**
+ * The read-replica cluster id for this cluster. It is serialized to the filesystem and up into
+ * zookeeper. This is a container for the id. Also knows how to serialize and deserialize the
+ * cluster id.
+ */
+@InterfaceAudience.Private
+public class ActiveClusterSuffix implements ClusterIdFile {
+  private final String cluster_id;
```
taklwu: nit: please follow the camel case in java class, update respectively in this file and other file.

```suggestion
  private final String clusterId;
```

hbase-client/src/main/java/org/apache/hadoop/hbase/ActiveClusterSuffix.java
```diff
@@ -0,0 +1,142 @@
...
+    }
+
+    @Override
+    public ActiveClusterSuffix readString(String input) {
+      return new ActiveClusterSuffix(input);
+    }
+  }
+
+  public ActiveClusterSuffix(final String ci, final String suffix) {
+    this.cluster_id = ci;
```
taklwu:
```suggestion
  public ActiveClusterSuffix(final String clusterId, final String suffix) {
    this.clusterId = clusterId;
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterFileSystem.java
```diff
@@ -382,4 +392,55 @@ public void stop() {
...
+            + new Path(rootdir, HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME) + " not found");
+        }
+        LOG.debug(
+          "Negotiating active cluster suffix file. File {} : File Suffix {} : Configured suffix {} :  Cluster ID : {}",
+          new Path(rootdir, HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME), acs, activeClusterSuffix,
+          getClusterId());
+        // Suffix file exists and we're in read/write mode. Content should match.
+        if (!this.activeClusterSuffix.equals(acs)) {
+          // throw error
+          LOG.info(
```
taklwu: nit: should the mismatch message be error level? or should it be at least a warning?

```suggestion
          LOG.error(
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterFileSystem.java
```diff
@@ -382,4 +392,55 @@ public void stop() {
   public void logFileSystemState(Logger log) throws IOException {
     CommonFSUtils.logFileSystemState(fs, rootdir, log);
   }
+
+  private void negotiateActiveClusterSuffixFile(long wait) throws IOException {
```
taklwu, line 397: nit: do we have any JIRA for documenting the switch to an active cluster from a read-replica by using the `hbase.meta.table.suffix`?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/CompactSplit.java
```diff
@@ -344,6 +345,12 @@ protected void requestCompactionInternal(HRegion region, HStore store, String wh
       return;
     }
 
+    // Should not allow compaction if cluster is in read-only mode
+    if (isReadOnlyEnabled()) {
+      LOG.info("Ignoring compaction request for " + region + ",because read-only mode is on.");
```
taklwu, line 350: nit about log message

```suggestion
      LOG.info("Ignoring compaction request for {} because read-only mode is on.", region);
```

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/CompactSplit.java
```diff
@@ -442,6 +449,13 @@ private Optional<CompactionContext> selectCompaction(HRegion region, HStore stor
       LOG.info(String.format("User has disabled compactions"));
       return Optional.empty();
     }
+
+    // Should not allow compaction if cluster is in read-only mode
+    if (isReadOnlyEnabled()) {
+      LOG.info(String.format("Compaction request skipped as read-only mode is on"));
```
taklwu:
```suggestion
      LOG.info("Compaction request skipped as read-only mode is on");
```

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/CompactSplit.java
```diff
@@ -857,6 +871,11 @@ public boolean isCompactionsEnabled() {
     return compactionsEnabled;
   }
 
+  private boolean isReadOnlyEnabled() {
+    return conf.getBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY,
+      HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
+  }
+
```
taklwu:
```suggestion
```

hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
```diff
@@ -102,4 +116,129 @@ private static boolean hasCoprocessorsConfigured(Configuration conf, String... c
...
+  /**
+   * This method adds or removes relevant ReadOnlyController coprocessors to the provided
+   * configuration based on whether read-only mode is enabled.
+   * @param conf               The up-to-date configuration used to determine how to handle
+   *                           coprocessors
+   * @param coprocessorConfKey The configuration key name
+   */
+  public static void syncReadOnlyConfigurations(Configuration conf, String coprocessorConfKey) {
+    boolean isReadOnlyModeEnabled = conf.getBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY,
+      HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
```
taklwu:
```suggestion
    boolean isReadOnlyModeEnabled = ConfigurationUtil.isReadOnlyModeEnabled(conf);
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterFileSystem.java
```diff
@@ -382,4 +392,55 @@ public void stop() {
...
+    this.activeClusterSuffix = ActiveClusterSuffix.fromConfig(conf, getClusterId());
+    if (!isReadOnlyModeEnabled(conf)) {
```
taklwu:
```suggestion
    if (!ConfigurationUtil.isReadOnlyModeEnabled(conf)) {
```

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterFileSystem.java
```diff
@@ -382,4 +392,55 @@ public void stop() {
...
+  }
+
+  public ActiveClusterSuffix getActiveClusterSuffix() {
+    return activeClusterSuffix;
+  }
+
+  private boolean isReadOnlyModeEnabled(Configuration conf) {
+    return conf.getBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY,
+      HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
+  }
```
taklwu:
```suggestion
```

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestReadOnlyController.java
```diff
@@ -0,0 +1,217 @@
...
+  public void afterClass() throws Exception {
+    if (connection != null) {
+      connection.close();
+    }
+    TEST_UTIL.shutdownMiniCluster();
+  }
+
+  private static void enableReadOnlyMode() {
+    // Dynamically enable Read-Only mode if it is not active
+    if (!isReadOnlyModeEnabled()) {
```
taklwu:
```suggestion
    if (!ConfigurationUtil.isReadOnlyModeEnabled(conf)) {
```

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestReadOnlyController.java
```diff
@@ -0,0 +1,217 @@
...
+      LOG.info("Dynamically enabling Read-Only mode by setting {} to true",
+        HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
+      conf.setBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY, true);
+      notifyObservers();
+    }
+  }
+
+  private static void disableReadOnlyMode() {
+    // Dynamically disable Read-Only mode if it is active
+    if (isReadOnlyModeEnabled()) {
```
taklwu:
```suggestion
    if (ConfigurationUtil.isReadOnlyModeEnabled(conf)) {
```

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestReadOnlyController.java
```diff
@@ -0,0 +1,217 @@
...
+        HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
+      conf.setBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY, false);
+      notifyObservers();
+    }
+  }
+
+  private static boolean isReadOnlyModeEnabled() {
+    return conf.getBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY,
+      HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
+  }
```
taklwu:
```suggestion
```

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestRefreshHFilesFromClient.java
```diff
@@ -0,0 +1,139 @@
...
+      Assert.fail("RefreshHFilesForTable Should Not Throw Exception: " + e);
+      throw new RuntimeException(e);
+    } finally {
+      // Delete table name post test execution
+      deleteTable(TEST_TABLE);
+    }
+  }
+
+  // Not creating table hence refresh should throw exception
+  @Test(expected = TableNotFoundException.class)
```
taklwu, line 75: nit: need some work for junit 4 to junit 5, but as mentioned, we at least need to know if this is a must in master branch.

taklwu (changes requested):
[DISCUSS] I'm wondered if we should always use junit 5 when merging to master branch, if so, would it be possible to have it in this PR? or at least I want a follow up task on it.

anmolnar:
> [DISCUSS] I'm wondered if we should always use junit 5 when merging to master branch, if so, would it be possible to have it in this PR? or at least I want a follow up task on it.

I completely agree. I haven't actually noticed that we're still using JUnit 4 somewhere. I tried to migrate every test during the rebasing. 

I created a sub-task for it: https://issues.apache.org/jira/browse/HBASE-30085

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestRefreshHFilesFromClient.java
anmolnar: I'm not sure if it's a must, but I believe we should definitely do it in this patch.

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
anmolnar: I don't think we need that comment. Could be removed.

hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/AbstractReadOnlyController.java
anmolnar: There's an "else" branch as well in this code at line 107. You should not fail fast here.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
anmolnar: Similarly, there's and _else_ branch in the `manageActiveClusterIdFile()` method which tries to remove the suffix file if already exists.

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestReadOnlyController.java
anmolnar: Why do you want to remove this?

anmolnar: Gotcha. For some odd reason I'm reading review comments in reversed order. :)

hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
kgeisz: Thanks @charlesconnell, I am working on using the loaded coproccessors (or lack thereof) to determine whether read-only mode is enabled

hbase-server/src/main/java/org/apache/hadoop/hbase/coprocessor/CoprocessorHost.java
kgeisz: This was introduced in PR https://github.com/apache/hbase/pull/6931 when we first added support for dynamic configuration with the ReadOnlyController.  At that time, we had only one ReadOnlyController class, and it implemented ConfigurationObserver because we needed to have dynamic configuration capabilities at the coprocessor level.  It looks like this code is not needed anymore, so I am working on removing it and making sure everything still works as expected.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
kgeisz: I got rid of this log message because the logic for determining if there was actually a configuration change was moved to `CoprocessorConfigurationUtil.maybeUpdateCoprocessors()`.  You can find the log message [here](https://github.com/apache/hbase/pull/8044/changes#diff-497ef4310988f57855965f564c729f02b1aabb5d3945f4844850d688c622471aR232).

I do think I should be a little more specific in this log message.  Right now, it uses the `toString()` representation of HMaster/HRegionServer/HRegion, which is a little vague.  For example, we see this for master on a local standalone deployment of HBase:

```
2026-04-15 16:52:51,650 INFO util.CoprocessorConfigurationUtil: Updating coprocessors for 10.0.0.227,16000,1776296876526 because the configuration has changed
```

I think it could say `...for Master 10.0.0.227,16000,1776296876526...` instead of just `...for 10.0.0.227,16000,1776296876526...`

I can also leave out the `toString()` representation entirely if you think that would be better.  I can use the coprocessor configuration key to figure out whether a master, region server, or region is what is being configured.

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
taklwu: are we going to fix this ?

taklwu (approved):
another comment but otherwise LGTM, 

I also assumed https://github.com/apache/hbase/pull/7730 will be additions to this PR after we merge this change?

anmolnar:
> I also assumed #7730 will be additions to this PR after we merge this change?

Unfortunately we're not getting anywhere with that PR. Given that it's not a strict requirement to complete the feature, I'd like to go forward with this patch at the first place.

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
anmolnar: @kgeisz Please add this to your upcoming patch.

anmolnar:
@charlesconnell Any news about the Yetus release?

charlesconnell:
No news about the Yetus release

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AdminOverAsyncAdmin.java
```diff
@@ -1153,6 +1153,26 @@ public List<String> getCachedFilesList(ServerName serverName) throws IOException
...
+
+  @Override
+  public Long refreshHFiles(final String namespace) throws IOException {
+    return get(admin.refreshHFiles(namespace));
+  }
+
+  @Override
+  public Long refreshHFiles() throws IOException {
+    return get(admin.refreshHFiles());
+  }
```
charlesconnell, line 1174: Perhaps these should return primitive `long`s? There is no need for them to be null.

kgeisz:
We have come across a bug where tables on the replica cluster are not getting new data post-refresh after the read-only flag has been flipped twice.  This is being tracked in [HBASE-30090](https://issues.apache.org/jira/browse/HBASE-30090), and we are working on a solution.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
kgeisz: This log message is now in `CoprocessorConfigurationUtil.maybeUpdateCoprocessors()`

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AdminOverAsyncAdmin.java
sharmaar12: Submitted PR [#8130 ](https://github.com/apache/hbase/pull/8130)

anmolnar:
hey @Apache9 - You might want to take a look at this patch sooner rather than later. Please share your thoughts when you get a chance.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
kgeisz: This has been addressed in PR https://github.com/apache/hbase/pull/8115

hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
kgeisz: This has been addressed in PR https://github.com/apache/hbase/pull/8115

hbase-server/src/main/java/org/apache/hadoop/hbase/coprocessor/CoprocessorHost.java
kgeisz: This has been addressed in PR https://github.com/apache/hbase/pull/8115

hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
kgeisz: This has been addressed in PR https://github.com/apache/hbase/pull/8115

hbase-common/src/main/java/org/apache/hadoop/hbase/HConstants.java
kgeisz: This has been addressed in PR https://github.com/apache/hbase/pull/8115.  Can we mark this comment as resolved?

kgeisz:
Hi @charlesconnell, PR https://github.com/apache/hbase/pull/8115 addresses some of your comments and has been merged into this feature branch.  Can you please mark the relevant comments as resolved?  Thanks

charlesconnell:
https://github.com/apache/hbase/pull/8182 will help with Yetus failures once it's merged and you rebase

anmolnar:
> #8182 will help with Yetus failures once it's merged and you rebase

Rebased.

anmolnar:
@charlesconnell Looks like Yetus is making progress with the builds this time.
We've finished all the outstanding review items on the patch. Could you please do another quick round of review and approve patch if everything is resolved?

charlesconnell:
The yetus checks are failing for legit reasons, for example:
```
[ERROR] /home/runner/work/hbase/hbase/src/hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/ReadOnlyController.java:[889,56] cannot find symbol
  symbol:   method getSuffixFileDataToWrite()
  location: variable mfs of type org.apache.hadoop.hbase.master.MasterFileSystem
```
and
```
/home/runner/work/hbase/hbase/src/hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/master/BackupLogCleaner.java:[207,70] error: cannot find symbol
  symbol:   variable MASTER_REGION_DIR_NAME
  location: class MasterRegionFactory
```

I am happy to review again when the build is fixed

ndimiduk:
I'm pretty sure what you're seeing is an artifact of how Yetus applies each commit in your PR branch (`git apply --binary -v --stat --apply -p1 input.patch`) against a clean checkout of master. Grepping logs, this looks important:

For ReadOnlyController.java, the apply log shows 9 stanzas:
- Line 544 of stat: 393 +++++++++ (HBASE-29083 adds the file)
- Lines 550, 552, 571, 649, 653, 655, 663, 669: subsequent modifications/re-adds across the feature branch
- Line 677: 913 -------------------- (HBASE-29841 deletion stanza, all -, no +)

Somehow it's not tracking changes to this file correctly across the entire history. Looks like this is a known issue in Yetus ([YETUS-983](https://issues.apache.org/jira/browse/YETUS-983)) and has been around since 2020.

My best recommendation for now is to squish your history a bit and see what comes of it.

ndimiduk:
Trying a fix. https://github.com/apache/yetus/pull/384

ndimiduk:
https://issues.apache.org/jira/browse/YETUS-983 is merged. Let's try it again.

anmolnar:
Thanks @ndimiduk for the quick help. I triggered another round of CI.

ndimiduk:
Glad to help.

FYI, last I looked, this commit history needs some cleanup before you land it on master. We still aim for one commit per JIRA issue. Not sure if you want to just squish it all down so that only the umbrella issue lands in the tree? In that case, be sure to update the fix version on all the subtasks appropriately (i.e., set to the feature branch release version, not the release version).

anmolnar:
> Glad to help.
> 
> FYI, last I looked, this commit history needs some cleanup before you land it on master. We still aim for one commit per JIRA issue. Not sure if you want to just squish it all down so that only the umbrella issue lands in the tree? In that case, be sure to update the fix version on all the subtasks appropriately (i.e., set to the feature branch release version, not the release version).

Thanks @ndimiduk for the heads-up, I didn't think about it yet. I can certainly squash the patch is a single commit that's quite straightforward. However, squashing multiple commits of a single sub-task would be a bit more time-consuming, but I think it's still feasible.

What do you recommend?

ndimiduk:
I think that a single squished commit is easiest.

charlesconnell:
I have one unresolved comment still, but other than that no new comments

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MasterFileSystem.java
anmolnar: We haven't documented anything yet until we finalize the implementation.
To make things more complicated #7730 is stuck at code reviews and if ever gets merged, it'll potentially change the implementation radically. 
I created a sub-task about the documentation: https://issues.apache.org/jira/browse/HBASE-30167

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestCanStartHBaseInReadOnlyMode.java
anmolnar: Done.

anmolnar:
> I have one unresolved comment still, but other than that no new comments

I think we have resolved all the outstanding comments. Please carry on with your review when you have a chance.

charlesconnell (approved):
Yetus failures seem bogus. LGTM.

anmolnar:
Thank you @charlesconnell ! Let me squash the commits.

anmolnar:
@ndimiduk Looks like squashing the commits doesn't make any difference.

ndimiduk:
It's not yetus, the failures are legit -- your merge target is shifting under you. You need to rebase onto master.

ndimiduk:
Click through into a failed check and look at the Summary document, https://github.com/apache/hbase/actions/runs/26121214670 . Click on the House and Summary at the top and you'll see an overview.
image

anmolnar:
Thanks @ndimiduk for looking. Let me do the rebase and fix the problems.

anmolnar:
I've validated the flaky tests and all of them passed on my machine locally.

Kota-SH (approved):
lgtm

kgeisz (approved):
LGTM

sharmaar12 (approved):
LGTM

PR #8052, opened 2026-04-09
HBASE-30070 Replace deprecated JRuby LoadService#findFileForLoad with $LOAD_PATH.resolve_feature_path

jinhyukify:
**Jira** https://issues.apache.org/jira/browse/HBASE-30070

### DIFF
--- a/hbase-shell/src/main/ruby/irb/hirb.rb
+++ b/hbase-shell/src/main/ruby/irb/hirb.rb
@@ -209,22 +209,19 @@ module IRB
     ##
     # Determine the loadable path for a given filename by searching through $LOAD_PATH
     #
-    # This serves a similar purpose to IRB::IrbLoader#search_file_from_ruby_path, but uses JRuby's
-    # loader, which allows us to find special paths like "uri:classloader" inside of a Jar.
+    # Uses $LOAD_PATH.resolve_feature_path, which resolves files on disk as well as
+    # entries inside jars on the classpath. Returns nil on miss.
     #
     # @param [String] filename
     # @return [String] path
     def self.path_for_load(filename)
       return File.absolute_path(filename) if File.exist? filename
 
-      # Get JRuby's LoadService from the global (singleton) instance of the runtime
-      # (org.jruby.Ruby), which allows us to use JRuby's tools for searching the load path.
-      runtime = org.jruby.Ruby.getGlobalRuntime
-      loader = runtime.getLoadService
-      search_state = loader.findFileForLoad filename
-      raise LoadError, "no such file to load -- #{filename}" if search_state.library.nil?
+      resolved = $LOAD_PATH.resolve_feature_path(filename)
+      raise LoadError, "no such file to load -- #{filename}" if resolved.nil?
 
-      search_state.loadName
+      _type, path = resolved
+      path
     end
 
     ##


junegunn:
Interesting, I'm not sure how many users rely on loading a script file this way, but we should definitely update the code.

_Here's my understanding of the issue._

In previous versions, the `SCRIPTFILE` argument of `hbase shell` was not limited to the actual file path. It also resolved relative paths against `$LOAD_PATH`, so users could set the `RUBYLIB` environment variable to add custom directories and load scripts from them.

```sh
mkdir -p scripts
echo "list; exit" > scripts/list.rb

RUBYLIB=scripts bin/hbase shell -n list.rb
```

However, this no longer works on the latest master:

```
ERROR NoMethodError: undefined method `findFileForLoad' for #<Java::OrgJrubyRuntimeLoad::LoadService:0x48788853>
Did you mean?  findFileInClasspath
```

I can confirm the patch fixes the problem.

---

That said, I'm not sure loading script files from arbitrary `$LOAD_PATH` entries is what we intended or is desirable. It seems to pose a security risk similar to why adding `.` to `$PATH` is discouraged.

Script resolution from `$LOAD_PATH` was never documented, so we can consider restricting this to actual file paths.

- https://hbase.apache.org/book.html#scripting

And by default, the shell doesn't add extra directories to `$LOAD_PATH`.

```
bin/hbase shell
HBase Shell
Use "help" to get list of supported commands.
Use "exit" to quit this interactive shell.
For Reference, please visit: https://hbase.apache.org/docs/shell
Version 4.0.0-alpha-1-SNAPSHOT, r859dc18330c15f15ad1b4f791bf0871793eeb228, Sun Apr  5 09:36:47 KST 2026
Took 0.0009 seconds
hbase:001:0> $LOAD_PATH
=>
[#<Pathname:/Users/ad02514045/github/hbase/bin/../hbase-shell/src/main/ruby>,
 "uri:classloader:/META-INF/jruby.home/lib/ruby/3.1/site_ruby",
 "uri:classloader:/META-INF/jruby.home/lib/ruby/stdlib"]
```

---

The patch is fine and is good to merge, but I think this is a good chance to discuss the concern.

junegunn:
@NihalJain Hi, could you take a look at this? We still have some code that was removed in JRuby 9.4.

junegunn:
Okay, now I see the point. While I can't imagine an HBase shell user relying on `RUBYLIB`, it is probable that they bundle some scripts in a jar file and add it to the CLASSPATH like so:

```sh
mkdir -p /tmp/scripts
echo 'puts 12345' > /tmp/scripts/demo.rb
jar -cf scripts.jar -C /tmp scripts/

bin/hbase shell -n scripts/demo.rb
  # ERROR LoadError: no such file to load -- scripts/demo.rb

HBASE_CLASSPATH=scripts.jar bin/hbase shell -n scripts/demo.rb
  # hbase:001:0> puts 12345
  # 12345

# Without fix
HBASE_CLASSPATH=scripts.jar bin/hbase shell -n scripts/demo.rb
  # NoMethodError: undefined method `workspace' for #<Shell::Shell:0x62732be7>
  # Did you mean?  get_workspace
```

The patch fixes the problem and it works as expected. Approved.

junegunn (approved):

PR #8057, opened 2026-04-10
HBASE-30073 Test fixes for some flappers and a reproducible error

haridsv:
Jira: [HBASE-30073](https://issues.apache.org/jira/browse/HBASE-30073)

- Fixed `TestZstdDictionarySplitMerge` by adding a missing test dependency
- Fixes for `TestBlockBytesScannedQuota`, `TestHRegion`, `TestHRegionWithInMemoryFlush`, `TestZooKeeper`, `TestStochasticLoadBalancerRegionReplicaSameHosts`, `TestAsyncTableScan`, `TestConnection` and `TestRollbackSCP` was provided by AI and verified by running 20 times each in a loop.

Here are the AI summaries:

---

# TestStochasticLoadBalancerRegionReplicaSameHosts

## **Root cause**

`StochasticLoadBalancer.balanceTable()` set `cluster.setStopRequestedAt(startTime + maxRunningTime)` **at the very beginning** of the method, **before** build ing costs (`initCosts`), the first `computeCost`, `needsBalance`, and other setup.

So the **entire** `maxRunningTime` budget (e.g. **250 ms** in `StochasticBalancerTestBase`) was counting **initialization + search**, not just the stochastic walk.

On a **slow or busy CI host**, setup alone could take **~250 ms or more**. Then when the walk started, `isStopRequested()` was already true (or became true af ter a single rejected move). The loop could exit with **`step` still 0**, **no accepted moves**, and `balanceCluster` returning **`null`** even though region repl icas were still colocated on the same host—so **`TestStochasticLoadBalancerRegionReplicaSameHosts`** failed. Locally, setup is usually only tens of ms, so the bug rarely shows up.

## **Fix**

**Set `setStopRequestedAt` only after initialization is done**, immediately **before** the stochastic `for` loop, using a new timestamp (`searchStartTime + maxRunningTime`). That way **`maxRunningTime` applies to the search phase only**, which matches the intended meaning of the limit and avoids “no search time left” on loaded agents.

# TestRollbackSCP

## **Root cause**

**Surefire reruns failed tests in the same JVM** without running `@BeforeClass` again. The test used a **static** `INJECTED` flag with `compareAndSet(false, t rue)` so fault injection (and `setKillAndToggleBeforeStoreUpdateInRollback`) only ran **once per JVM**.

After the **first** failure (e.g. `IllegalArgumentException: scheduler queue not empty` at `restartMasterProcedureExecutor`), **`INJECTED` stayed `true`**. On **reruns**, `compareAndSet` did nothing, so the executor was never killed in rollback and **`procExec.isRunning()` stayed true**, which made `waitFor(30000, () -> !procExec.isRunning())` **always time out**.

So the “three different failures” were really: **first attempt** hit the real error at line 180; **attempts 2–3** were **misleading timeouts** at line 177 caused by broken rerun state, not the original bug.

## *Fix*

In **`@Before`** (which **does** run before each attempt, including reruns), reset test-only state:

1. **`INJECTED.set(false)`** so injection can arm again on each run.
2. **`ProcedureTestingUtility.setKillAndToggleBeforeStoreUpdateInRollback(procExec,
false)`** so procedure-executor kill flags from the previous attempt are cleared.

That makes each Surefire attempt behave like a fresh test with respect to
injection, so reruns can pass when the underlying issue is fixed or flaky.

# TestZooKeeper

**Problem Identified:**
- Line 78 had a duplicate call to `TEST_UTIL.startMiniDFSCluster(2)`
- This violated the "can only create one" constraint and caused HDFS namespace conflicts when tests ran in parallel

**Why it was flaky:**
- ✅ **Local/isolated runs**: Worked because no leftover HDFS state
- ❌ **CI/parallel runs**: Failed when other ZooKeeper tests ran first, leaving behind HDFS data with incompatible `namespaceID`

**Fix Applied:**
Removed the duplicate call on line 78. The test now properly:
1. Starts Mini DFS Cluster (line 71)
2. Starts Mini ZK Cluster (line 72)
3. Configures settings (lines 73-77)
4. Done! (no duplicate cluster start)

This matches the pattern used in other similar tests like `TestClusterBootOrder` and aligns with the comment's intent: "Test we can first start the ZK cluster by itself."

The test should now be stable across both local and CI environments. You can verify by running the test suite again to ensure `TestZooKeeper` no longer fails with `InconsistentFSStateException`.

# TestHRegion and TestHRegionWithInMemoryFlush

## Fix 1

**Root Cause Analysis**

The TestHRegionWithInMemoryFlush test is flapping due to zombie background threads that continue running after tests timeout, causing directory race conditions with subsequent tests.

**What's Happening**

1. testParallelIncrementWithMemStoreFlush and testParallelAppendWithMemStoreFlush create background flush threads
2. These threads run in a busy loop: while (!incrementDone.get())
3. When the test times out (after ~4 minutes in CI), an InterruptedException is thrown during thread.join()
4. The code never reaches the lines that set the done flag and stop the flush thread
5. The zombie flush thread keeps running, busy-looping trying to flush
6. Subsequent tests' tearDown() methods clean up directories
7. The zombie thread tries to access deleted directories → race condition errors

**Evidence**

- Bad run shows only 5 tests completed (vs 136 in good run)
- Thread-4940 logs "Skipping flush because closing" repeatedly for over a minute
- 6 flaky errors about missing directories in unrelated tests like testGet_Basic and testScanner_JoinedScanners
- Error messages: "Parent directory doesn't exist", "chmod: cannot access"

**The Fix**

I've fixed both problematic tests in /Users/hdara/src/apache/hbase/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestHRegion.java:

Changes made:
1. Added interrupt check to flush thread loop: while (!done.get() && !Thread.currentThread().isInterrupted())
2. Wrapped test logic in try-finally to ensure cleanup even on timeout/failure
3. In finally block:
- Set done flag to true
- Interrupt the flush thread
- Wait up to 5 seconds for thread to stop
- Log warning if thread doesn't stop

Files modified:
- TestHRegion.java:5322-5368 - testParallelIncrementWithMemStoreFlush
- TestHRegion.java:5407-5459 - testParallelAppendWithMemStoreFlush

**Why This Only Happened in CI**

CI environments are typically slower and more heavily loaded, causing tests to hit the 13-minute timeout threshold. Locally, the tests complete faster, so the threads finish normally before cleanup.

The fix ensures that background threads are always properly stopped, preventing zombie threads from interfering with subsequent tests.

## Fix 2

Fix for another flapper condition diagnosed from TestHRegion failure

Summry of the fix by AI:

Fix directory race condition in parallel test execution

When tests run in parallel, HBaseTestingUtil.createSubDirAndSystemProperty()
would reuse system property values (like hadoop.log.dir and hadoop.tmp.dir)
set by other parallel test instances. This caused a race condition:

1. Test A calls setupDataTestDir(), creates unique directory UUID-A
2. Test A sets System.setProperty("hadoop.log.dir", UUID-A/hadoop-log-dir)
3. Test B calls setupDataTestDir(), creates unique directory UUID-B
4. Test B calls createSubDirAndSystemProperty() for hadoop.log.dir
5. System property already set, so Test B reuses UUID-A/hadoop-log-dir
6. Test A finishes, calls cleanupTestDir(), deletes UUID-A directory
7. Test B fails with "chmod: cannot access '.../WALs': No such file"

The original logic was added in HBASE-5064 (2012) to support parallel tests,
with the comment "we do nothing but hope that there will be no conflicts"
suggesting the authors knew this was imperfect but necessary for mini cluster
compatibility.

Fixed by checking if the directory referenced by the system property still
exists before reusing it:
- If exists: reuse it (preserves original behavior for mini clusters)
- If deleted: create new unique directory for this test instance

This prevents tests from using deleted directories while maintaining
compatibility with components that may require shared system properties.

## Fix 3

Thread Cleanup Fixes Applied

1. testMutateRowInParallel (line 7487)

- Issue: Writer thread had while (true) with no interrupt check
- Fix: Changed to while (!Thread.currentThread().isInterrupted()), added try-finally block with thread cleanup
- Impact: This was failing in both TestHRegion and TestHRegionWithInMemoryFlush

2. testGetWhileRegionClose (line 1323)

- Issue: Thread cleanup code was OUTSIDE the finally block, so threads would keep running if test threw exception
- Fix: Moved done.set(true) and thread joins INTO the finally block
- Impact: Prevents zombie threads if test times out before normal completion

3. GetTillDoneOrException thread class (line 1377)

- Issue: Loop only checked !done flag, no interrupt handling
- Fix: Changed to while (!done.get() && !Thread.currentThread().isInterrupted())
- Impact: Threads can now be interrupted by JUnit timeout

4. FlushThread class (line 4575)

- Issue: Loop only checked !done flag
- Fix: Changed to while (!done && !Thread.currentThread().isInterrupted())
- Impact: More robust termination on test timeout

5. PutThread class (line 4791)

- Issue: Loop only checked !done flag
- Fix: Changed to while (!done && !Thread.currentThread().isInterrupted())
- Impact: More robust termination on test timeout

### Root Cause
The test `testPrefetchRunTriggersEvictions` failed because:
1. **Random file size**: 10,000 KVs created ~668KB of data (334 blocks × 2KB)
2. **Insufficient to fill cache**: The 1MB cache wasn't exceeded, so **no evictions occurred**
3. **Flawed assertion**: Assumed `failedInserts == 0` meant evictions MUST have happened
4. **Result**: `evictedCount = 0` AND `failedInserts = 0` → test failed

### Fix Applied

**File**: `TestPrefetchWithBucketCache.java:322-354`

1. **Increased KVs 10,000 → 15,000** (50% increase)
   - Now reliably creates ~1MB+ of data, exceeding cache capacity
   - More resource-efficient than doubling to 20,000

2. **Fixed assertion logic**:
   ```java
   // OLD (incorrect):
   if (bc.getStats().getFailedInserts() == 0) {
     assertNotEquals(0, bc.getStats().getEvictedCount());  // ❌ Could fail
   }

   // NEW (correct):
   long evictions = bc.getStats().getEvictedCount();
   long failedInserts = bc.getStats().getFailedInserts();
   assertTrue(
     "Expected either evictions or failed inserts...",
     evictions > 0 || failedInserts > 0);  // ✅ Accepts either outcome
   ```

3. **Fixed filename**: "testPrefetchInterruptOnCapacity" → "testPrefetchRunTriggersEvictions"

The test now correctly validates capacity management without requiring a specific mechanism (evictions vs failed inserts), while using reasonable resources (1.5x KVs instead of 2x).

### Summary of Fixes

All three flapping tests had the **identical root cause**: directory sharing race conditions in parallel test execution. Each test was trying to reuse directories from previous test runs that were already deleted, causing tests to fail or not be discovered properly.

#### Files Modified:

1. **AbstractTestAsyncTableScan.java** (fixes TestAsyncTableScan)
   - Added a helper method `createConfiguration()` that sets `hbase.test.disable-directory-sharing=true`
   - Passed this configuration to `MiniClusterRule.newBuilder().setConfiguration()`
   - Location: `/Users/hdara/src/apache/hbase/hbase-server/src/test/java/org/apache/hadoop/hbase/client/AbstractTestAsyncTableScan.java:77-88`

2. **TestReplicationBase.java** (fixes TestEditsBehindDroppedTableTiming and all replication tests)
   - Added `conf.setBoolean("hbase.test.disable-directory-sharing", true)` in the `setupConfig()` method
   - This affects both UTIL1 and UTIL2, ensuring both replication clusters use isolated directories
   - Location: `/Users/hdara/src/apache/hbase/hbase-server/src/test/java/org/apache/hadoop/hbase/replication/TestReplicationBase.java:195-198`

3. **TestConnection.java**
   - Added `TEST_UTIL.getConfiguration().setBoolean("hbase.test.disable-directory-sharing", true)` in `setUpBeforeClass()` before `startMiniCluster()`
   - Location: `/Users/hdara/src/apache/hbase/hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestConnection.java:90-93`

#### How It Works:

The fix leverages the infrastructure you already added to `HBaseTestingUtil` in commit `30816234db`, which checks for the `hbase.test.disable-directory-sharing` configuration. When set to `true`, each test gets its own unique test data directory instead of trying to reuse system properties from previous parallel test runs.

This prevents the race condition where:
- Test A starts and creates directories
- Test B starts in parallel and reuses Test A's directories (via system properties)
- Test A finishes and cleans up its directories
- Test B tries to use the now-deleted directories → **FAILURE**

# TestBlockBytesScannedQuota

The test is flapping due to a **timing/race condition** in the quota system:

1. **5-second timeout too short**: The `testTraffic` method only waited 5 seconds for quotas to take effect
2. **Quota cache not fully propagated**: On slower systems (like CI), the quota cache refresh can be asynchronous and may not fully propagate in time
3. **Quotas bypassed**: When cache isn't refreshed, the logs show `"bypass expected false, actual true"`, meaning all requests succeed instead of being throttled
4. **Insufficient retries**: Each iteration takes ~1.3 seconds, so only 3-4 retries fit in 5 seconds, not enough for the quota system to stabilize

**Bad run pattern:**
- Test expects 1 successful request but gets 5 (all succeed because quotas not enforced)
- Retries every ~1.3 seconds for 4 attempts
- Times out after 5 seconds with "Waiting timed out after [5,000] msec"

**Good run pattern:**
- Quotas enforced immediately
- Tests pass quickly (36.97s total vs 63.14s for failed run)

Increased the timeout in `testTraffic()` from **5,000ms to 30,000ms** (line 263). This gives the quota system sufficient time to:
- Complete cache refresh
- Propagate quota settings across all components
- Handle slower CI environments

This is a conservative fix that maintains the retry logic while allowing adequate time for the distributed quota system to stabilize. The 30-second timeout is still reasonable for a test and should handle the asynchronous nature of quota enforcement.

### DIFF
--- a/hbase-balancer/src/main/java/org/apache/hadoop/hbase/master/balancer/StochasticLoadBalancer.java
+++ b/hbase-balancer/src/main/java/org/apache/hadoop/hbase/master/balancer/StochasticLoadBalancer.java
@@ -581,7 +581,6 @@ public class StochasticLoadBalancer extends BaseLoadBalancer {
       rackManager, regionCacheRatioOnOldServerMap);
 
     long startTime = EnvironmentEdgeManager.currentTime();
-    cluster.setStopRequestedAt(startTime + maxRunningTime);
 
     initCosts(cluster);
     balancerConditionals.loadClusterState(cluster);
@@ -632,6 +631,10 @@ public class StochasticLoadBalancer extends BaseLoadBalancer {
       currentCost / sumMultiplier, functionCost(), computedMaxSteps);
 
     final String initFunctionTotalCosts = totalCostsPerFunc();
+    long searchStartTime = EnvironmentEdgeManager.currentTime();
+    // Budget maxRunningTime for the stochastic walk only; initialization (cluster costs, etc.)
+    // can be substantial on busy hosts and must not consume the search deadline.
+    cluster.setStopRequestedAt(searchStartTime + maxRunningTime);
     // Perform a stochastic walk to see if we can get a good fit.
     long step;
     boolean planImprovedConditionals = false;
--- a/hbase-common/src/test/java/org/apache/hadoop/hbase/HBaseJupiterExtension.java
+++ b/hbase-common/src/test/java/org/apache/hadoop/hbase/HBaseJupiterExtension.java
@@ -84,7 +84,7 @@ public class HBaseJupiterExtension implements InvocationInterceptor, BeforeAllCa
 
   private static final Map<String, Duration> TAG_TO_TIMEOUT =
     ImmutableMap.of(SmallTests.TAG, Duration.ofMinutes(3), MediumTests.TAG, Duration.ofMinutes(6),
-      LargeTests.TAG, Duration.ofMinutes(13), IntegrationTests.TAG, Duration.ZERO);
+      LargeTests.TAG, Duration.ofMinutes(20), IntegrationTests.TAG, Duration.ZERO);
 
   private static final String EXECUTOR = "executor";
 
--- a/hbase-compression/pom.xml
+++ b/hbase-compression/pom.xml
@@ -45,6 +45,11 @@
       <artifactId>hbase-resource-bundle</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-api</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
@@ -435,17 +435,26 @@ public class HBaseTestingUtil extends HBaseZKTestingUtil {
     String sysValue = System.getProperty(propertyName);
 
     if (sysValue != null) {
-      // There is already a value set. So we do nothing but hope
-      // that there will be no conflicts
-      LOG.info("System.getProperty(\"" + propertyName + "\") already set to: " + sysValue
-        + " so I do NOT create it in " + parent);
-      String confValue = conf.get(propertyName);
-      if (confValue != null && !confValue.endsWith(sysValue)) {
-        LOG.warn(propertyName + " property value differs in configuration and system: "
-          + "Configuration=" + confValue + " while System=" + sysValue
-          + " Erasing configuration value by system value.");
+      // There is already a value set by a previous test. Check if the directory still exists.
+      File sysDir = new File(sysValue);
+      if (sysDir.exists()) {
+        // Directory exists, so another test may still be using it. Reuse it to avoid conflicts.
+        LOG.info("System.getProperty(\"" + propertyName + "\") already set to: " + sysValue
+          + " and directory exists, so reusing it for " + parent);
+        String confValue = conf.get(propertyName);
+        if (confValue != null && !confValue.endsWith(sysValue)) {
+          LOG.warn(propertyName + " property value differs in configuration and system: "
+            + "Configuration=" + confValue + " while System=" + sysValue
+            + " Erasing configuration value by system value.");
+        }
+        conf.set(propertyName, sysValue);
+      } else {
+        // Directory was deleted (previous test cleaned up), so create our own.
+        LOG.info("System.getProperty(\"" + propertyName + "\") set to: " + sysValue
+          + " but directory no longer exists. Creating new directory under " + parent);
+        createSubDir(propertyName, parent, subDirName);
+        System.setProperty(propertyName, conf.get(propertyName));
       }
-      conf.set(propertyName, sysValue);
     } else {
       // Ok, it's not set, so we create it as a subdirectory
       createSubDir(propertyName, parent, subDirName);
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
@@ -442,7 +442,7 @@ public class HBaseTestingUtil extends HBaseZKTestingUtil {
 
     if (sysValue != null && !disableSharing) {
       // There is already a value set. So we do nothing but hope
-      //  that there will be no conflicts
+      // that there will be no conflicts
       LOG.info("System.getProperty(\"" + propertyName + "\") already set to: " + sysValue
         + " so I do NOT create it in " + parent);
       String confValue = conf.get(propertyName);
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
@@ -434,29 +434,26 @@ public class HBaseTestingUtil extends HBaseZKTestingUtil {
 
     String sysValue = System.getProperty(propertyName);
 
-    if (sysValue != null) {
-      // There is already a value set by a previous test. Check if the directory still exists.
-      File sysDir = new File(sysValue);
-      if (sysDir.exists()) {
-        // Directory exists, so another test may still be using it. Reuse it to avoid conflicts.
-        LOG.info("System.getProperty(\"" + propertyName + "\") already set to: " + sysValue
-          + " and directory exists, so reusing it for " + parent);
-        String confValue = conf.get(propertyName);
-        if (confValue != null && !confValue.endsWith(sysValue)) {
-          LOG.warn(propertyName + " property value differs in configuration and system: "
-            + "Configuration=" + confValue + " while System=" + sysValue
-            + " Erasing configuration value by system value.");
-        }
-        conf.set(propertyName, sysValue);
-      } else {
-        // Directory was deleted (previous test cleaned up), so create our own.
-        LOG.info("System.getProperty(\"" + propertyName + "\") set to: " + sysValue
-          + " but directory no longer exists. Creating new directory under " + parent);
-        createSubDir(propertyName, parent, subDirName);
-        System.setProperty(propertyName, conf.get(propertyName));
+    // Check if directory sharing should be disabled for this test.
+    // Tests that run with high parallelism and don't need shared directories can set this
+    // to avoid race conditions where one test's tearDown() deletes directories another test
+    // is still using.
+    boolean disableSharing = conf.getBoolean("hbase.test.disable-directory-sharing", false);
+
+    if (sysValue != null && !disableSharing) {
+      // There is already a value set. So we do nothing but hope
+      //  that there will be no conflicts
+      LOG.info("System.getProperty(\"" + propertyName + "\") already set to: " + sysValue
+        + " so I do NOT create it in " + parent);
+      String confValue = conf.get(propertyName);
+      if (confValue != null && !confValue.endsWith(sysValue)) {
+        LOG.warn(propertyName + " property value differs in configuration and system: "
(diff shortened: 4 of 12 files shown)


charlesconnell (approved):
amazing!

liuxiaocs7:
Thanks @haridsv and @charlesconnell, great job!

Apache9:
Please revert this PR because it is not qualified enough.

Wo do not prevent using AI coding tool but at least you should review the code by yourself and be resposible for the code.

For example, increasing the timeout value is not the correct way here, you should make the test smaller instead of increasing the timeout.

And for load balancer, changing the stop time calculation changes the behavior, we should discuss first before changing the behavior.

And on the directory sharing, the method is only called when creating hadoop.log.dir and hadoop.tmp.dir directories, can you confirm that the failures on CI are caused by sharing these two directories? We need evidence, not only the conclusion from AI.

And flaky rerun has been a problem for a long time, I've also seen the reruns all failed because of mini cluster already started. But we should have common solution for all tests, not patching a single test.

And no matter with or without AI, you should split this PR to several small ones since they are different problems.

Apache9:
And if you use AI coding tool to provide a big PR without knowing all the details, at least you should provide the information about the tool and the model you use, and also the prompt you used to generate the PR.

haridsv:
I did review the changes and raised the PR after making some tweaks, but of course I am not an expert in HBase code base so relied on the PR reviews to identify anything that I missed.

> And for load balancer, changing the stop time calculation changes the behavior, we should discuss first before changing the behavior.

Yes, except this load balancer fix, all other fixes are purely in the test itself or in the test framework, so IMHO there is non immediate danger to prod code. How about reverting the load balancer change alone and raising a separate PR so that there can be a discussion around it? Alternatively, if you have any review feedback, I can raise a new PR to address it.

> we should have common solution for all tests, not patching a single test

I purposefully limited the scope of tmp directory fix to only those that are failing, but it is easy enough change the default and make it applicable to all tests, please let me know if you think this is correct and I can raise a new PR.

> We need evidence, not only the conclusion from AI.

The conclusion was based on the actual logs from both failed and successful runs and I can attach the full test logs and AI chat history, if one wants to double check.

> you should split this PR to several small ones since they are different problems

I can definitely ensure of this for future fixes, but considering the work already done (including back porting), IMHO, the right thing to do is to review the PR and see if any of those changes need to be reverted rather than a full revert.

Just to add a little more context, I started with a couple of fixes, but when the fix PR itself started hitting new flapper, I kept diagnosing and adding fixes to the same PR until I got a green build. 

> And if you use AI coding tool to provide a big PR without knowing all the details, at least you should provide the information about the tool and the model you use, and also the prompt you used to generate the PR.

Sure, I used Claude Code with the default Sonnet 4.5 model. I used the following format as prompt:

`You need to diagnose the HBase test XYZ that is flapping. It had multiple failures and errors in the PR validation build, but none when I ran locally. The source code is at @../../src/apache/hbase/ and the logs for good run is at @good-run and bad run is at @bad-run. Can you compare the logs and try to understand why the run failed and look at the code for a potential fix?`

Apache9:
Reverting a PR is very common if we have some mistakes in the PR, we can apply it later after fixing the mistakes.

At least we should not increase the test timeout limit, and also the load balancer change.

And on the tmp directory fix, could you please paste the log sections which shows problems for sharing tmp directory? Or you can paste the AI chat log and the full log file it mentioned so we can review it.

> Just to add a little more context, I started with a couple of fixes, but when the fix PR itself started hitting new flapper, I kept diagnosing and adding fixes to the same PR until I got a green build.

I'd say this is common in HBase, that's why we have a flaky dashboard to show flaky tests. Starting a mini cluster in test is heavy and easy to have races when running on a slow machine...

On the AI tool, Sonnet 4.5 is not powerful enough, Opus 4.6 will be bettter. And for debugging flaky tests, I recommend use Codex 5.4 xhigh since HBase is a project live for more than 15 years, we have lots of strange code in our code base, you need to think more and more before actually doing things...

haridsv:
@Apache9 Apologies for the delayed response. I see you've already taken care of the revert — thank you! I'll go ahead and raise the new PRs now.

PR #8058, opened 2026-04-10
HBASE-29289: Make hbase master ui show cluster client connections info

hingu-8103:
Added ClientConnectionInfo object which tracks client connection info for each client connections and also added ClientConnectionRegistry to track all the active ClientConnectionInfo. 

I have added basic tests to test the change and also manually tested on the cluster.

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/UserMetricsBuilder.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/UserMetricsBuilder.java
@@ -36,7 +36,9 @@ public final class UserMetricsBuilder {
     userLoad.getClientMetricsList().stream()
       .map(clientMetrics -> new ClientMetricsImpl(clientMetrics.getHostName(),
         clientMetrics.getReadRequestsCount(), clientMetrics.getWriteRequestsCount(),
-        clientMetrics.getFilteredRequestsCount()))
+        clientMetrics.getFilteredRequestsCount(), clientMetrics.getHostAddress(),
+        clientMetrics.getUserName(), clientMetrics.getServiceName(),
+        clientMetrics.getClientVersion()))
       .forEach(builder::addClientMetris);
     return builder.build();
   }
@@ -49,7 +51,10 @@ public final class UserMetricsBuilder {
         .setHostName(clientMetrics.getHostName())
         .setWriteRequestsCount(clientMetrics.getWriteRequestsCount())
         .setReadRequestsCount(clientMetrics.getReadRequestsCount())
-        .setFilteredRequestsCount(clientMetrics.getFilteredReadRequestsCount()).build())
+        .setFilteredRequestsCount(clientMetrics.getFilteredReadRequestsCount())
+        .setHostAddress(clientMetrics.getHostAddress()).setUserName(clientMetrics.getUserName())
+        .setServiceName(clientMetrics.getServiceName())
+        .setClientVersion(clientMetrics.getClientVersion()).build())
       .forEach(builder::addClientMetrics);
     return builder.build();
   }
@@ -79,13 +84,22 @@ public final class UserMetricsBuilder {
     private final String hostName;
     private final long readRequestCount;
     private final long writeRequestCount;
+    private final String hostAddress;
+    private final String userName;
+    private final String serviceName;
+    private final String clientVersion;
 
     public ClientMetricsImpl(String hostName, long readRequest, long writeRequest,
-      long filteredReadRequestsCount) {
+      long filteredReadRequestsCount, String hostAddress, String userName, String serviceName,
+      String clientVersion) {
       this.hostName = hostName;
       this.readRequestCount = readRequest;
       this.writeRequestCount = writeRequest;
       this.filteredReadRequestsCount = filteredReadRequestsCount;
+      this.hostAddress = hostAddress != null ? hostAddress : "Unknown";
+      this.userName = userName != null ? userName : "Unknown";
+      this.serviceName = serviceName != null ? serviceName : "Unknown";
+      this.clientVersion = clientVersion != null ? clientVersion : "Unknown";
     }
 
     @Override
@@ -106,6 +120,26 @@ public final class UserMetricsBuilder {
     @Override
     public long getFilteredReadRequestsCount() {
       return filteredReadRequestsCount;
+    }
+
+    @Override
+    public String getHostAddress() {
+      return hostAddress;
+    }
+
+    @Override
+    public String getUserName() {
+      return userName;
+    }
+
+    @Override
+    public String getServiceName() {
+      return serviceName;
+    }
+
+    @Override
+    public String getClientVersion() {
+      return clientVersion;
     }
   }
 
--- a/hbase-protocol-shaded/src/main/protobuf/server/ClusterStatus.proto
+++ b/hbase-protocol-shaded/src/main/protobuf/server/ClusterStatus.proto
@@ -206,6 +206,18 @@ message ClientMetrics {
 
   /** the current total filtered requests made from a client */
   optional uint64 filtered_requests_count = 4;
+
+  /** the IP address of the client */
+  optional string host_address = 5;
+
+  /** the name of the user making requests from this client */
+  optional string user_name = 6;
+
+  /** the version of the client */
+  optional string client_version = 7;
+
+  /** the protobuf RPC service name from the client connection header */
+  optional string service_name = 8;
 }
 
 /* Server-level protobufs */
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsUserAggregateImpl.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsUserAggregateImpl.java
@@ -23,10 +23,13 @@ import java.util.Optional;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.CompatibilitySingletonFactory;
 import org.apache.hadoop.hbase.ipc.RpcServer;
+import org.apache.hadoop.hbase.ipc.ServerRpcConnection;
 import org.apache.hadoop.hbase.security.User;
 import org.apache.hadoop.hbase.security.UserProvider;
 import org.apache.hadoop.hbase.util.LossyCounting;
 import org.apache.yetus.audience.InterfaceAudience;
+
+import org.apache.hadoop.hbase.shaded.protobuf.generated.RPCProtos;
 
 @InterfaceAudience.Private
 public class MetricsUserAggregateImpl implements MetricsUserAggregate {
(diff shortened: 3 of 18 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/ipc/TestNettyServerRpcConnectionClientConnectionInfo.java
```diff
@@ -0,0 +1,195 @@
...
+
+    // Wait for the channel to close
+    channel.closeFuture().await(5000, TimeUnit.MILLISECONDS);
+
+    // Wait for the ClientConnectionInfo to be unregistered
+    Waiter.waitFor(CONF, 5000, () -> registry.getClientConnections().size() <= initialSize);
+
+    // Verify that the connection is unregistered
+    connections = registry.getClientConnections();
+    assertTrue(connections.size() <= initialSize);
```
hingu-8103, line 123: I believe this assertion might introduce flakiness in tests when there are new client connections happens during this test execution by some other test case which is running parallel.

How should i make test robust in this kind of scenarios ?

Apache9: I do not think other tests can connect to the rpc server started in this test?

hingu-8103: Thanks @Apache9 for the clarification. I have verified it won't introduce any flakiness.

hbase-protocol-shaded/src/main/protobuf/server/ClusterStatus.proto
wchevreuil, line 1: There's already a ClientMetrics object defined in ClusterStatus.proto, which is used by the hbtop tool to provide some client related information. I think we should add these there and reuse that object on this UI, while also update hbtop to report the additional information.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/ClientConnectionRegistry.java
wchevreuil, line 1: Another reason to try integrate this into the already existing client metrics: MetricsUserAggregateImpl already provides its own maps and tracking mechanisms, which would dismiss the need for changes in the rpc layer.

wchevreuil (changes requested):
I think we should try to reuse the existing metrics subsystem for tracking and reporting these info. Please look at MetricsUserSourceImpl, MetricsUserAggregateImpl, LossyCounting, ClusterStatus.ClientMetrics. If we reuse the existing metric system, we don't need the extra tracking logic at related RPC layer classes and it may also make it easier to update the hbtop views.

hbase-server/src/main/java/org/apache/hadoop/hbase/ipc/ClientConnectionRegistry.java
hingu-8103: Thanks @wchevreuil for the comments, Let me go through the existing implementation and try to use that for this purpose.

hbase-server/src/main/resources/hbase-webapps/master/regionServerListClientConnectionsStats.jsp
```diff
@@ -0,0 +1,54 @@
...
+          for(Map.Entry<String, UserMetrics.ClientMetrics> clientEntry : clientMetricsMap.entrySet()) {
+            UserMetrics.ClientMetrics clientConnection = clientEntry.getValue();
+
+  %>
+  <tr>
+    <td><%= clientConnection.getHostAddress() %></td>
+    <td><%= clientConnection.getUserName() %></td>
+    <td><%= clientConnection.getClientVersion() %></td>
+    <td><%= clientConnection.getServiceName() %></td>
+    <td><%= serverName.getServerName() %></td>cc
```
PDavid: Here there's a stray `cc` after the closing </td> tag that will render as visible text in the HTML.


```suggestion
    <td><%= serverName.getServerName() %></td>
```

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsUserAggregateImpl.java
```diff
@@ -23,10 +23,15 @@
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.CompatibilitySingletonFactory;
 import org.apache.hadoop.hbase.ipc.RpcServer;
+import org.apache.hadoop.hbase.ipc.ServerRpcConnection;
 import org.apache.hadoop.hbase.security.User;
 import org.apache.hadoop.hbase.security.UserProvider;
 import org.apache.hadoop.hbase.util.LossyCounting;
 import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
```
PDavid: These are unused imports. Can we remove them?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsUserAggregateImpl.java
```diff
@@ -193,4 +204,51 @@ private MetricsUserSource getOrCreateMetricsUser(String user) {
...
+  private ClientConnectionContext getClientConnectionContext() {
+    String hostAddress = null;
+    String userName = "Unknown";
+    String clientVersion = "Unknown";
+    String serviceName = "Unknown";
+
+    Optional<ServerRpcConnection> rpcConnectionOptional = RpcServer.getCurrentServerRpcConnection();
+    if (rpcConnectionOptional.isPresent()) {
+      ServerRpcConnection rpcConnection = rpcConnectionOptional.get();
+      hostAddress = rpcConnection.getLocalHostAddress();
```
PDavid: The field `localHostAddress` in `ServerRpcConnection` represents the server-side IP (the RS's own address), but it's surfaced in the UI under the column "ClientIP" (via `getHostAddress()`).

This is confusing, the value shown as "ClientIP" is actually the region server's local bind address, not the client's IP. Shouldn't we use `hostAddress` from `ServerRpcConnection` here instead?

hbase-server/src/main/resources/hbase-webapps/master/regionServerListClientConnectionsStats.jsp
```diff
@@ -0,0 +1,54 @@
...
+
+          for(Map.Entry<String, UserMetrics.ClientMetrics> clientEntry : clientMetricsMap.entrySet()) {
+            UserMetrics.ClientMetrics clientConnection = clientEntry.getValue();
+
+  %>
+  <tr>
+    <td><%= clientConnection.getHostAddress() %></td>
+    <td><%= clientConnection.getUserName() %></td>
+    <td><%= clientConnection.getClientVersion() %></td>
+    <td><%= clientConnection.getServiceName() %></td>
```
PDavid: Here we directly interpolate values from these client-controlled fields (`userName`, `clientVersion`, `serviceName`) using `<%= ... %>` without HTML escaping. Malicious or buggy clients could inject `<script>` tags. Other HBase JSPs use `StringEscapeUtils.escapeHtml4()` - the same should be applied here.

hbase-protocol-shaded/src/main/protobuf/server/ClusterStatus.proto
```diff
@@ -212,6 +206,11 @@ message ClientMetrics {
 
   /** the current total filtered requests made from a client */
   optional uint64 filtered_requests_count = 4;
+
+  optional string host_address = 5;
+  optional string user_name = 6;
+  optional string client_version = 7;
+  optional string service_name = 8;
```
PDavid, line 220: Other fields are all having comments. Would it make sense to add comments for these new fields as well?

PDavid (commented):
Many thanks, looks nice already.

hbase-client/src/main/java/org/apache/hadoop/hbase/UserMetricsBuilder.java
```diff
@@ -79,13 +84,27 @@ public static class ClientMetricsImpl implements UserMetrics.ClientMetrics {
     private final String hostName;
     private final long readRequestCount;
     private final long writeRequestCount;
+    private final String hostAddress;
```
wchevreuil, line 87: Is hostAddress different from hostName or could we just reuse the already existing hostName?

hbase-server/src/main/resources/hbase-webapps/master/regionServerList.jsp
```diff
@@ -63,6 +63,9 @@
     <li class="nav-item">
       <a class="nav-link" href="#tab_replicationStats" data-bs-toggle="tab" role="tab">Replications</a>
     </li>
+    <li class="nav-item">
+      <a class="nav-link" href="#tab_clientConnectionsStats" data-bs-toggle="tab" role="tab">ClientConnections</a>
+    </li>
```
wchevreuil, line 68: nit: Client Connections

hbase-client/src/main/java/org/apache/hadoop/hbase/UserMetricsBuilder.java
hingu-8103: I'm trying to store IP address of the host in hostAddress whereas hostName can have both Name/IP.

hingu-8103:
I've addressed all the comments and added tests as well for this task. 

For tests, 
1. Modified hbtop related tests to include the newly introduced fields. 
2. Added new test to verify serialization to protobuf from ClientMetrics and again deserialization to ClientMetrics from protobuf.
3. Also added verification for Client Connection related fields in Master UI page.

hingu-8103:
The PR run failed due to flakey tests. I ran flakey tests locally on this branch and they all passed.

PDavid (approved):
Many thanks for the improvements. This looks good to me. 👍

wchevreuil:
How should I test it? I'm trying by running a full scan on a table from hbase shell, however, I don't see any updates for th "client connections" tab in Master UI, not the related view in hbtop. Am I missing something here?

PDavid:
> How should I test it? I'm trying by running a full scan on a table from hbase shell, however, I don't see any updates for th "client connections" tab in Master UI, not the related view in hbtop. Am I missing something here?

@wchevreuil I think when I tested it last time, the key was to add the following to hbase-site.xml:

```xml
  <property>
    <name>hbase.regionserver.user.metrics.enabled</name>
    <value>true</value>
  </property>
```

hingu-8103:
@wchevreuil sorry about that i forgot mention about that property, Do we need to document this anywhere ?
cc : @PDavid

wchevreuil (approved):

wchevreuil:
> @wchevreuil sorry about that i forgot mention about that property, Do we need to document this anywhere ? cc : @PDavid

Just add it on the jira release-notes field.

PR #8061, opened 2026-04-11
HBASE-30074 Update downloads content.md for 2.5.14

apurtell:


apurtell:
This is the first time I'm doing this since the migration. Asking for a double check. Is this the correct way to do this?

liuxiaocs7 (approved):

PDavid (approved):
Many thanks Andrew, looks good to me. 👍

Apache9 (approved):

PR #8062, opened 2026-04-11
HBASE-30075 Update downloads content.md for 2.6.5

apurtell:


apurtell:
This is the first time I'm doing this since the migration. Asking for a double check. Is this the correct way to do this?

liuxiaocs7 (approved):

PDavid (approved):
Many thanks, built the site locally and changes look good to me. 👍

Apache9 (approved):

PR #8072, opened 2026-04-12
HBASE-30079 Upgrade hbase-compression to use junit5

liuxiaocs7:
- see: HBASE-30079

### DIFF
--- a/hbase-compression/pom.xml
+++ b/hbase-compression/pom.xml
@@ -50,6 +50,11 @@
       <artifactId>junit-jupiter-api</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.junit.jupiter</groupId>
+      <artifactId>junit-jupiter-engine</artifactId>
+      <scope>test</scope>
+    </dependency>
   </dependencies>
 
   <build>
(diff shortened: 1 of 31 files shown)


hbase-compression/pom.xml
```diff
@@ -45,6 +45,21 @@
       <artifactId>hbase-resource-bundle</artifactId>
       <optional>true</optional>
     </dependency>
+    <dependency>
```
Apache9, line 48: We do not have junit-vintage dependencies in the past?

Strange, how could they pass compiling?

liuxiaocs7: > We do not have junit-vintage dependencies in the past?
> 
> Strange, how could they pass compiling?

Yes, previously the `hbase-compression` module did not explicitly reference JUnit-related dependencies...

PDavid:
Hmm, it seems the java compiler now generates some new warnings. I guess these are not critical. Can you please check them?

image

See: https://github.com/apache/hbase/actions/runs/24324541700?pr=8072

Apache9:
Pelase try to fix the javac warnings?

liuxiaocs7:
Now the general check looks good.

https://github.com/apache/hbase/actions/runs/25016408688/job/73265250014?pr=8072

image

Apache9 (approved):

PR #8077, opened 2026-04-14
HBASE-30062 Device layer simulator for MiniDFSCluster-based tests

apurtell:
On EBS-backed deployments in AWS, or equivalents in other cloud infrastructure providers, HBase compaction and replication throughput can be constrained by per-volume IOPS limits rather than bandwidth. A faithful device-level simulator within the test harness allows developers to reproduce, analyze, and validate fixes for such performance issues without requiring actual cloud infrastructure.

This proposed change adds a test-only EBS device layer that operates at the DataNode storage level within `MiniDFSCluster` by replacing the `FsDatasetSpi` implementation via Hadoop's pluggable factory mechanism. This allows HBase integration tests to simulate realistic cloud block storage characteristics, such as per-volume bandwidth budgets, IOPS limits, sequential IO coalescing, and per-IO device latency, enabling identification and reproduction of IO bottlenecks.

The simulator wraps the real `FsDatasetImpl` with a `java.lang.reflect.Proxy` that intercepts the three `FsDatasetSpi` methods where DataNode local IO actually engages the underlying block device, without compile-time coupling to internal Hadoop classes.

Each proxy gets its own set of `EBSVolumeDevice` instances with independent budgets. Block-to-volume resolution uses `delegate.getVolume(block)`, providing real HDFS placement decisions. A single configuration applies to all volumes, but each volume maintains its own token buckets, matching production where all attached block devices to a host share the same SKU but have independent throughput budgets, and where the host itself has a cap on maximum aggregate throughput.

EBS merges sequential IOs up to 1 MiB before counting them as a single IOPS token. The simulator tracks read streams and write streams independently.

After each IOPS token consumption, the simulator sleeps for a configurable duration (default 1 ms), modeling physical device service time.

`IntegrationTestCompactionWithDeviceSimulator` provides an end-to-end example of using the simulator

### DIFF
--- a/hbase-it/src/test/java/org/apache/hadoop/hbase/IntegrationTestCompactionWithDeviceSimulator.java
+++ b/hbase-it/src/test/java/org/apache/hadoop/hbase/IntegrationTestCompactionWithDeviceSimulator.java
@@ -1,0 +1,564 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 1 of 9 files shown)


apurtell:
I see from feedback on another PR I need to upgrade the tests here to JUnit 5.

PR #8081, opened 2026-04-14
HBASE-30080: Send exception to user when refreshing store file fails for a region

sharmaar12:


anmolnar (changes requested):
I think we could go much simpler than that. Don't need to catch the exception at all, we don't need to explicitly log it here. Just let it be caught at higher levels, like other Callables do.

Don't need to unit test either.

anmolnar (approved):

anmolnar:
Merged. Thanks @sharmaar12 !

PR #8100, opened 2026-04-17
HBASE-28741 Rpc ConnectionRegistry APIs should have timeout

ndimiduk:


virajjasani (commented):
Finally!
Looks good overall, just one thought - can we use a new config instead of using `hbase.rpc.timeout`? Because this is more of metadata level API than normal rpc API, we can consider a bit lower timeout as default value e.g. `DEFAULT_HBASE_RPC_TIMEOUT / 6` or `DEFAULT_HBASE_RPC_TIMEOUT / 3`

hbase-client/src/main/java/org/apache/hadoop/hbase/client/ConnectionRegistryRpcStubHolder.java
```diff
@@ -106,23 +106,35 @@ private ImmutableMap<ServerName, ClientMetaService.Interface> createStubs(RpcCli
...
-          if (error != null) {
-            addr2StubFuture.completeExceptionally(error);
-          } else {
-            RpcClient c = RpcClientFactory.createClient(conf, clusterId);
-            ImmutableMap<ServerName, ClientMetaService.Interface> m =
-              createStubs(c, bootstrapNodes);
-            rpcClient = c;
-            addr2Stub = m;
-            addr2StubFuture.complete(m);
+    try {
```
Apache9, line 109: Why we need the try catch here? FutureUtils.addListener does not throw any exception out.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/ClusterIdFetcher.java
```diff
@@ -88,10 +91,9 @@ class ClusterIdFetcher {
   private void getClusterId(int index) {
     ServerName server = bootstrapServers.get(index);
     LOG.debug("Going to request {} for getting cluster id", server);
-    // user and rpcTimeout are both not important here, as we will not actually send any rpc calls
-    // out, only a preamble connection header, but if we pass null as user, there will be NPE in
-    // some code paths...
-    RpcChannel channel = rpcClient.createRpcChannel(server, user, 0);
+    // The preamble exchange still requires TCP connect and potentially TLS negotiation, so we use
+    // the configured rpc timeout to bound the connection attempt.
+    RpcChannel channel = rpcClient.createRpcChannel(server, user, rpcTimeoutMs);
```
Apache9, line 96: We only send a preamble header here, I wonder whether the rpc timeout actually work. And I can make sure that we do not have TLS negotiation when sending preamble header.

hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestConnectionRegistryRpcClientFailure.java
```diff
@@ -0,0 +1,203 @@
...
+        "Expected simulated failure in cause chain, got: " + e);
+    }
+  }
+
+  /**
+   * Verify that after a failed stub initialization, subsequent getClusterId() calls also fail
+   * promptly rather than returning a zombie future.
+   */
+  @Test
+  public void testSubsequentCallsAfterFailure() throws Exception {
```
Apache9, line 192: Does this test fail before the patch? We will create a ClusterIdFetcher every time when fetching in RpcConnectionRegistry, so I do not get the point why we could return a zombie future...

hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestConnectionRegistryRpcClientFailure.java
```diff
@@ -0,0 +1,203 @@
...
+   * RPC client that succeeds for the preamble cluster ID fetch (clusterId is null) but throws on
+   * the real client creation (clusterId is non-null). This simulates the production failure where
+   * TLS certificate provisioning fails during RPC client construction.
+   */
+  public static final class FailingRpcClientImpl implements RpcClient {
+
+    public FailingRpcClientImpl(Configuration configuration, String clusterId,
+      SocketAddress localAddress, MetricsConnection metrics, Map<String, byte[]> attributes) {
+      if (clusterId != null) {
+        throw new RuntimeException("Simulated RPC client creation failure");
```
Apache9, line 87: So the actually problem is here, we may throw exception when creating RpcClient?

Looking at the code, neither BlockingRpcClient nor NettyRpcClient will throw exception when constructing.

```
  @SuppressWarnings("FutureReturnValueIgnored")
  public static <T> void addListener(CompletableFuture<T> future,
    BiConsumer<? super T, ? super Throwable> action) {
    future.whenComplete((resp, error) -> {
      try {
        // See this post on stack overflow(shorten since the url is too long),
        // https://s.apache.org/completionexception
        // For a chain of CompletableFuture, only the first child CompletableFuture can get the
        // original exception, others will get a CompletionException, which wraps the original
        // exception. So here we unwrap it before passing it to the callback action.
        action.accept(resp, unwrapCompletionException(error));
      } catch (Throwable t) {
        LOG.error("Unexpected error caught when processing CompletableFuture", t);
      }
    });
  }
```

We will log something out when the callback throws exception, have you seen something like this?

ndimiduk:
I don't personally think a new config is needed here, but if you really feel strongly about it...

We run 2.6 in prod and my investigation into the incident was rooted in that version. I naively implemented a fix on master. Let me do a gap analysis between branch-2.6 and master.

hbase-client/src/test/java/org/apache/hadoop/hbase/client/TestConnectionRegistryRpcClientFailure.java
ndimiduk: Yes, we saw exactly the "Unexpected error caught when processing CompletableFuture" log — that's how we traced this. The exception came from a custom RpcClient loaded via CUSTOM_RPC_CLIENT_IMPL_CONF_KEY. Our impl does TLS cert provisioning in the constructor (HTTP call to a pod-local certificate authority), which threw when the CA sidecar returned 500.

You're right that the built-in clients don't throw today. But CUSTOM_RPC_CLIENT_IMPL_CONF_KEY is a public extension point and constructors can throw. FutureUtils.addListener already acknowledges unexpected callback exceptions at line 72, but logging isn't enough when there's an orphaned future that callers are blocked on.

PR #8109, opened 2026-04-21
HBASE-30100 Fix misleading Javadoc for Table.batch about null in results array

junegunn:
https://issues.apache.org/jira/browse/HBASE-30100


PDavid (approved):
Many thanks for the correction. 👍

PR #8122, opened 2026-04-24
HBASE-30101 Move login() before RpcServer construction

junegunn:
Login before creating the RPC services to prevent a stray TGT Renewer thread in HBase 2.6+.

Description:
- https://issues.apache.org/jira/browse/HBASE-30101?focusedCommentId=18075865&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-18075865

Related:
- https://github.com/apache/hbase/pull/8110


Apache9 (commented):
I think this is the correct way to fix the problem, we should setup the resources before starting rpc server.

junegunn:
I agree. #8110 doesn't fix the problem on 2.6+ anyway. I still think it's the right thing to do, but once this is merged, that one becomes optional.

Apache9 (approved):

PR #8128, opened 2026-04-24
HBASE-30102 Add metric to account for region data classified as cold by the Time Based Priority logic

wchevreuil:


hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionWrapper.java
```diff
@@ -80,6 +80,11 @@ public interface MetricsRegionWrapper {
    */
   float getCurrentRegionCacheRatio();
 
+  /**
+   * Gets the current cold date % ratio for this region.
```
petersomogyi:
```suggestion
   * Gets the current cold data % ratio for this region.
```

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -2490,7 +2490,16 @@ public Optional<Boolean> shouldCacheFile(HFileInfo hFileInfo, Configuration conf
...
     if (dataTieringManager != null && !dataTieringManager.isHotData(hFileInfo, conf)) {
-      LOG.debug("Data tiering is enabled for file: '{}' and it is not hot data", fileName);
+      LOG.debug("Custom tiering is enabled for file: '{}' and it is not hot data", fileName);
+      // If custom tiering has been just enabled for a file that was cached, we now need
+      // to evict it.
+      Set<BlockCacheKey> keySet =
+        getAllCacheKeysForFile(hFileInfo.getHFileContext().getHFileName(), 0, Long.MAX_VALUE);
+      int evictedBlocks = evictBlockSet(keySet);
+      if (evictedBlocks > 0) {
+        LOG.info("Evicted {} blocks for file {} as it is now considered cold by DataTieringManager",
```
taklwu: nit: should we have it as debug level? I'm wondered if we see a lot of these message.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketProtoUtils.java
```diff
@@ -100,9 +100,16 @@ private static void addEntryToBuilder(Map.Entry<BlockCacheKey, BucketEntry> entr
...
-      .setBlockType(toPB(key.getBlockType())).build();
+    BucketCacheProtos.BlockCacheKey.Builder builder = BucketCacheProtos.BlockCacheKey.newBuilder()
+      .setHfilename(key.getHfileName()).setOffset(key.getOffset())
+      .setPrimaryReplicaBlock(key.isPrimary()).setBlockType(toPB(key.getBlockType()));
+    if (key.getCfName() != null) {
+      builder.setFamilyName(key.getCfName());
+    }
+    if (key.getRegionName() != null) {
+      builder.setRegionName(key.getRegionName());
+    }
```
taklwu, line 111: question: why weren't the cf and regionname filled before ? is it because the cold data needs for log message or other compute usage?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java
```diff
@@ -1613,6 +1600,16 @@ RegionLoad createRegionLoad(final HRegion r, RegionLoad.Builder regionLoadBldr,
...
+    final MutableFloat currentRegionColdDataRatio = new MutableFloat(0.0f);
+    if (DataTieringManager.getInstance() != null) {
+      Pair<List<String>, Long> coldEntry =
+        DataTieringManager.getInstance().getRegionColdDataSize().get(regionEncodedName);
+      if (coldEntry != null) {
+        int coldSizeMB = roundSize(coldEntry.getSecond(), unitMB);
+        currentRegionColdDataRatio
+          .setValue(regionSizeMB == 0 ? 0.0f : (float) coldSizeMB / regionSizeMB);
+      }
+    }
```
taklwu, line 1612: nit: the code logic is right, just wondered if we should align the same lambda expression style like the lines above between 1594 and 1602 ?

taklwu (changes requested):
when Time Based Priority is disabled, what would be the `% Cold Data`? is it always showing?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketProtoUtils.java
wchevreuil: Yes. This is required not only by the new "coldDataRatio" metric we are adding, but also the existing "regionCachedRatio" that is critical for the CacheAwareLoadBalancer. Without this change here, we cannot calculate these metrics when recovering the persistent cache. IMO, it's a bug in the current CacheAwareLoadBalancer implementation.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
wchevreuil: Maybe, yeah. Although it would be triggered only upon enabling of the time based priority on the individual store, and once for each affected file, it can still flood the logs. Let me switch it to DEBUG.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java
wchevreuil: Sure, let me do it.

taklwu (approved):
LGTM.

something is wrong with the github action, can you give a try to trigger them ?

wchevreuil:
> when Time Based Priority is disabled, what would be the `% Cold Data`? is it always showing?

It would show as 0%. With the "% Cold Data" and "% Cached", operators can infer if there's indeed a problem with the region cache, as those are mutually exclusive, if both are low, it means the region caching went into some problems, most likely, not enough cache capacity.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/DataTieringManager.java
```diff
@@ -347,4 +387,29 @@ private static boolean isDataTieringFeatureEnabled(Configuration conf) {
...
+      for (HStoreFile file : compactedFiles) {
+        if (v.getFirst().contains(file.getPath().getName())) {
+          v.getFirst().remove(file.getPath().getName());
+          v.setSecond(v.getSecond() - Bytes.toLong(file.getMetadataValue(FILE_SIZE)));
+        }
+      }
+      for (HStoreFile file : newFiles) {
+        // call isHotData to account for the new file size in regionColdDataSize, if the new file is
+        // considered cold data as per data-tiering logic.
+        isHotData(file.getFileInfo().getHFileInfo(), file.getFileInfo().getConf());
```
petersomogyi, line 410: Can this cause a deadlock? This part is inside `regionColdDataSize.computeIfPresent` block and the `isHotData` also runs `regionColdDataSize.compute` on the same ConcurrentMap.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -2490,7 +2490,17 @@ public Optional<Boolean> shouldCacheFile(HFileInfo hFileInfo, Configuration conf
     String fileName = hFileInfo.getHFileContext().getHFileName();
     DataTieringManager dataTieringManager = DataTieringManager.getInstance();
     if (dataTieringManager != null && !dataTieringManager.isHotData(hFileInfo, conf)) {
-      LOG.debug("Data tiering is enabled for file: '{}' and it is not hot data", fileName);
+      LOG.debug("Custom tiering is enabled for file: '{}' and it is not hot data", fileName);
+      // If custom tiering has been just enabled for a file that was cached, we now need
+      // to evict it.
+      Set<BlockCacheKey> keySet =
+        getAllCacheKeysForFile(hFileInfo.getHFileContext().getHFileName(), 0, Long.MAX_VALUE);
+      int evictedBlocks = evictBlockSet(keySet);
```
petersomogyi, line 2498: The method name `shouldCacheFile` suggests this just a check but it actually evicts the blocks.

hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerSource.java
```diff
@@ -654,6 +654,11 @@ public interface MetricsRegionServerSource extends BaseSource, JvmPauseMonitorSo
   String CURRENT_REGION_CACHE_RATIO = "currentRegionCacheRatio";
   String CURRENT_REGION_CACHE_RATIO_DESC = "The percentage of caching completed for this region.";
 
+  String CURRENT_REGION_COLD_DATA_RATIO = "currentRegionColdDataRatio";
+
+  String CURRENT_REGION_COLD_DATA_RATIO_DESC = "The percentage of data in this region that "
+    + "is marked as cold by the configured time based priority logic. ";
```
petersomogyi:
```suggestion
    + "is marked as cold by the configured time based priority logic.";
```

wchevreuil:
> LGTM.
> 
> something is wrong with the github action, can you give a try to trigger them ?

Sure. I've run the failing tests locally, and it all passed. I'm trying another round here.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/DataTieringManager.java
wchevreuil: You mean, if another thread is calling isHotData? I don't think it would, which ever reaches the regionColdDataSize atomic methods first would own the lock and block the other, right?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
wchevreuil: Yeah, not ideal, but I still think it should be BucketCache responsibility to evict blocks from files that became classified as cold, upon changing time based priority configuration. Moving this to DataTieringManager would tighter couple it with BucketCache.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/DataTieringManager.java
petersomogyi: I checked this part with Claude and it gave this answer:
> No — this is a single-thread deadlock. That's what makes it especially insidious.
>
>  One thread does this:
>
>  1. Enters computeIfPresent(regionName, lambda) — acquires the internal bin lock
>  2. Inside the lambda, calls isHotData(...)
>  3. isHotData calls compute(regionName, lambda2) — tries to acquire the same bin lock
>  4. The lock is non-reentrant, so the same thread blocks waiting on itself
>
>  It's not a classic two-thread deadlock — it's a single thread trying to re-acquire a lock it already holds, on a lock that doesn't support reentrancy. The thread hangs forever.
>
>  This will happen every time a compaction produces a new cold file in a region that already has cold data tracked. That's a normal steady-state scenario, not a rare race condition.

wchevreuil: I don't think this is true. I've added a UT that simulates the compaction resulting in new cold file (the scenario claude mentioned) and it doesn't dead lock.

petersomogyi: Thanks for taking a look and adding a unit test for it!

petersomogyi (approved):

PR #6986, opened 2025-05-13
HBASE-29314 TableDescriptorChecker should verify CF configuration set via setConfiguration (branch-2)

junegunn:
HBASE-29137 was backported to branch-2 in #6799, but @guluo2016 pointed out that it doesn't perform verification of the column-familiy level configuration.

* See https://github.com/apache/hbase/pull/6799#discussion_r2084374632

This was due to the change of the alter command of HBase shell in #1959 in master branch which decided to use `setValue` instead of `setConfiguration`. So in branch-2, we need to check the values in the `configuration`, instead of `values`. But in this patch, I updated the code to check both.

The unit test was unable to detect this because it was using `setValue` instead of `setConfiguration`.

@Apache9 Do you think we also need this in master branch? Could you explain the difference between "values" and "configuration"? Are we phasing out the latter?

PR #8144, opened 2026-04-26
HBASE-30117 Fix HBase shell CONFIGURATION for column family using setValue instead of setConfiguration

jinhyukify:
**Jira** https://issues.apache.org/jira/browse/HBASE-30117

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/util/TableDescriptorChecker.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/util/TableDescriptorChecker.java
@@ -233,8 +233,8 @@ public final class TableDescriptorChecker {
     checkDateTieredCompactionForTimeRangeDataTiering(conf);
     for (ColumnFamilyDescriptor cfd : td.getColumnFamilies()) {
       // Column family level configurations
-      Configuration cfdConf =
-        new CompoundConfiguration().add(conf).addStringMap(cfd.getConfiguration());
+      Configuration cfdConf = new CompoundConfiguration().add(conf)
+        .addStringMap(cfd.getConfiguration()).addBytesMap(cfd.getValues());
       checkDateTieredCompactionForTimeRangeDataTiering(cfdConf);
     }
   }
@@ -318,7 +318,8 @@ public final class TableDescriptorChecker {
     throws IOException {
     warnOrThrowExceptionForFailure(conf, () -> {
       for (ColumnFamilyDescriptor cfd : td.getColumnFamilies()) {
-        Configuration cfdConf = new CompoundConfiguration().addStringMap(cfd.getConfiguration());
+        Configuration cfdConf = new CompoundConfiguration().addStringMap(cfd.getConfiguration())
+          .addBytesMap(cfd.getValues());
         try {
           BloomFilterUtil.getBloomFilterParam(cfd.getBloomFilterType(), cfdConf);
         } catch (IllegalArgumentException e) {
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestIllegalTableDescriptor.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestIllegalTableDescriptor.java
@@ -212,24 +212,38 @@ public class TestIllegalTableDescriptor {
       "org.apache.hadoop.hbase.regionserver.DefaultStoreEngine");
     checkTableIsIllegal(builder.build());
 
-    // column family level configuration changes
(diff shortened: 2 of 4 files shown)


junegunn:
It seems no one has a clear answer, but I got the impression that "configuration" was being phased out in favor of "value". So previously in https://github.com/apache/hbase/pull/6986, I updated the checker to check both maps, instead of reverting what we already changed in HBASE-20819. Can we do the same here, instead of reviving the use of the configuration map?

jinhyukify:
@junegunn Thank you for your review.
Good point. agree with keeping the [HBASE-20819](https://issues.apache.org/jira/browse/HBASE-20819) and [HBASE-29314](https://issues.apache.org/jira/browse/HBASE-29314) direction. I've reverted the `admin.rb` change and instead extended the same `CompoundConfiguration` pattern to the two CF-level validators that were still reading only from `cfd.getConfiguration()`:

- `TableDescriptorChecker#checkBloomFilterType`
- `TableDescriptorChecker#checkDateTieredCompactionForTimeRangeDataTiering`

hbase-server/src/main/java/org/apache/hadoop/hbase/util/TableDescriptorChecker.java
```diff
@@ -233,8 +233,8 @@ private static void checkDateTieredCompactionForTimeRangeDataTiering(final Confi
     checkDateTieredCompactionForTimeRangeDataTiering(conf);
     for (ColumnFamilyDescriptor cfd : td.getColumnFamilies()) {
       // Column family level configurations
-      Configuration cfdConf =
-        new CompoundConfiguration().add(conf).addStringMap(cfd.getConfiguration());
+      Configuration cfdConf = new CompoundConfiguration().add(conf)
+        .addStringMap(cfd.getConfiguration()).addBytesMap(cfd.getValues());
```
jinhyukify, line 237: This will fix this bug:

### Before
```
# This should fail but succeed
create 'mytable2', {NAME => 'f', CONFIGURATION => {'hbase.hstore.datatiering.type'  => 'TIME_RANGE', 'hbase.hstore.engine.class' => 'org.apache.hadoop.hbase.regionserver.DefaultStoreEngine'}}
```


### After
```
# This fails
create 'mytable2', {NAME => 'f', CONFIGURATION => {'hbase.hstore.datatiering.type'  => 'TIME_RANGE', 'hbase.hstore.engine.class' => 'org.apache.hadoop.hbase.regionserver.DefaultStoreEngine'}}

ERROR: org.apache.hadoop.hbase.DoNotRetryIOException: Time Range Data Tiering should be enabled with Date Tiered Compaction. Set hbase.table.sanity.checks to false at conf or table descriptor if you want to bypass sanity checks
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.warnOrThrowExceptionForFailure(TableDescriptorChecker.java:365)
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.warnOrThrowExceptionForFailure(TableDescriptorChecker.java:382)
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.checkDateTieredCompactionForTimeRangeDataTiering(TableDescriptorChecker.java:247)
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.checkDateTieredCompactionForTimeRangeDataTiering(TableDescriptorChecker.java:238)
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.sanityCheck(TableDescriptorChecker.java:210)
        at org.apache.hadoop.hbase.master.HMaster.createTable(HMaster.java:2494)
        at org.apache.hadoop.hbase.master.MasterRpcServices.createTable(MasterRpcServices.java:792)
        at org.apache.hadoop.hbase.shaded.protobuf.generated.MasterProtos$MasterService$2.callBlockingMethod(MasterProtos.java)
        at org.apache.hadoop.hbase.ipc.RpcServer.call(RpcServer.java:461)
        at org.apache.hadoop.hbase.ipc.CallRunner.run(CallRunner.java:124)
        at org.apache.hadoop.hbase.ipc.RpcHandler.run(RpcHandler.java:102)
        at org.apache.hadoop.hbase.ipc.RpcHandler.run(RpcHandler.java:82)
Caused by: java.lang.IllegalArgumentException: Time Range Data Tiering should be enabled with Date Tiered Compaction.
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.lambda$checkDateTieredCompactionForTimeRangeDataTiering$4(TableDescriptorChecker.java:253)
        at org.apache.hadoop.hbase.util.TableDescriptorChecker.warnOrThrowExceptionForFailure(TableDescriptorChecker.java:380)
        ... 10 more
```

junegunn (approved):

PR #8151, opened 2026-04-27
HBASE-30124 Upgrade hbase-server to use junit5 Part14

liuxiaocs7:
- see: HBASE-30124

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/compactions/PerfTestCompactionPolicies.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/compactions/PerfTestCompactionPolicies.java
@@ -25,9 +25,10 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
+import java.util.stream.Stream;
 import org.apache.hadoop.conf.Configuration;
-import org.apache.hadoop.hbase.HBaseClassTestRule;
 import org.apache.hadoop.hbase.HBaseConfiguration;
+import org.apache.hadoop.hbase.HBaseParameterizedTestTemplate;
 import org.apache.hadoop.hbase.client.RegionInfoBuilder;
 import org.apache.hadoop.hbase.logging.Log4jUtils;
 import org.apache.hadoop.hbase.regionserver.HStore;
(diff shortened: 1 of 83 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/wal/TestWALCellCodecWithCompression.java
```diff
@@ -17,78 +17,71 @@
...
+  @TestTemplate
   public void testEncodeDecodeKVsWithTagsWithTagsCompression() throws Exception {
     doTest(true, false);
   }
 
-  @Test
+  @TestTemplate
   public void testEncodeDecodeOffKVsWithTagsWithTagsCompression() throws Exception {
     doTest(true, false);
   }
```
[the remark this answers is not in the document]
liuxiaocs7: Will file another jira to resolve this issue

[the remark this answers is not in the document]
liuxiaocs7: see: HBASE-30128 and https://github.com/apache/hbase/pull/8158

Apache9 (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/compactions/PerfTestCompactionPolicies.java
```diff
@@ -36,25 +37,20 @@
...
 /**
  * This is not a unit test. It is not run as part of the general unit test suite. It is for
  * comparing compaction policies. You must run it explicitly; e.g. mvn test
  * -Dtest=PerfTestCompactionPolicies
  */
-@Category({ RegionServerTests.class, MediumTests.class })
-@RunWith(Parameterized.class)
+@Tag(RegionServerTests.TAG)
+@Tag(MediumTests.TAG)
+@HBaseParameterizedTestTemplate
```
Apache9: Better add name field here? And seems the javadoc is incorrect now?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/wal/AbstractTestFSWAL.java
```diff
@@ -391,20 +391,22 @@ public void testFindMemStoresEligibleForFlush() throws Exception {
...
+        HConstants.HREGION_OLDLOGDIR_NAME, CONF, null, true, null, null);
+      long filenum = EnvironmentEdgeManager.currentTime();
+      Path path = wal.computeFilename(filenum);
+      wal.createWriterInstance(FS, path);
+      Path parent = path.getParent();
+      path = wal.computeFilename(filenum + 1);
+      Path newPath = new Path(parent.getParent(), parent.getName() + "-splitting");
+      FS.rename(parent, newPath);
+      wal.createWriterInstance(FS, path);
+      fail("It should fail to create the new WAL");
```
Apache9: We do not need this fail assertion any more? And we'd better only wrap the statement which throws IOException only, not the whole method.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/wal/AbstractTestFSWAL.java
```diff
@@ -533,13 +535,15 @@ public void testWriteEntryCanBeNull() throws IOException {
...
-    AbstractFSWAL<?> wal = newWAL(FS, CommonFSUtils.getWALRootDir(CONF), DIR.toString(), testName,
-      CONF, null, true, null, null);
-    wal.close();
-    wal.rollWriter();
+    assertThrows(WALClosedException.class, () -> {
+      String testName = currentTest;
+      AbstractFSWAL<?> wal = newWAL(FS, CommonFSUtils.getWALRootDir(CONF), DIR.toString(), testName,
+        CONF, null, true, null, null);
+      wal.close();
+      wal.rollWriter();
```
Apache9: Ditto.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/wal/TestAsyncFSWALDurability.java
```diff
@@ -22,37 +22,32 @@
...
+import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Tag;
 
 import org.apache.hbase.thirdparty.io.netty.channel.Channel;
 import org.apache.hbase.thirdparty.io.netty.channel.EventLoopGroup;
 import org.apache.hbase.thirdparty.io.netty.channel.nio.NioEventLoopGroup;
 import org.apache.hbase.thirdparty.io.netty.channel.socket.nio.NioSocketChannel;
 
-@Category({ RegionServerServices.class, SmallTests.class })
```
Apache9, line 41: OK, fixed a typo...

Apache9 (approved):

liuxiaocs7:
I checked, and the number of UTs is the same before and after the modification

PR #8155, opened 2026-04-27
HBASE-30019 Introduce CacheEngine and CacheTopology abstractions

VladRodionov:
This PR targets the HBASE-30018 feature branch.

Introduces foundational internal abstractions for the pluggable block cache architecture:

- CacheEngine
- CacheEngineType
- CacheTopology
- CacheTopologyType
- CacheTier
- CacheEngineView
- CacheTopologyView
- initial topology stubs

No production wiring is added in this PR.
No behavior change intended.

JIRA: HBASE-30019

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
@@ -1,0 +1,292 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.io.hfile.cache;
+
+import java.util.Optional;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.io.hfile.BlockCacheKey;
+import org.apache.hadoop.hbase.io.hfile.BlockType;
+import org.apache.hadoop.hbase.io.hfile.CacheStats;
+import org.apache.hadoop.hbase.io.hfile.Cacheable;
+import org.apache.hadoop.hbase.io.hfile.HFileBlock;
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * Storage abstraction for a concrete HBase block cache backend.
+ * <p>
+ * A {@code CacheEngine} represents the storage layer only. It is responsible for storing,
+ * retrieving, invalidating, and reporting statistics for cached blocks. It does not perform tier
+ * orchestration, admission control, placement decisions, or promotion/demotion across cache levels.
+ * </p>
+ * <p>
+ * This interface is intentionally aligned with the storage-oriented subset of the current
+ * {@code BlockCache} contract so that existing implementations such as LruBlockCache and
+ * BucketCache can be migrated incrementally with minimal behavioral risk.
+ * </p>
+ * <p>
+ * Responsibilities of a {@code CacheEngine} include:
+ * </p>
+ * <ul>
+ * <li>block lookup</li>
+ * <li>block insertion</li>
+ * <li>targeted invalidation / eviction</li>
+ * <li>capacity and occupancy reporting</li>
+ * <li>engine-local statistics</li>
+ * <li>optional implementation-specific fit/capability checks</li>
+ * </ul>
+ * <p>
+ * Non-responsibilities include:
+ * </p>
+ * <ul>
+ * <li>L1/L2 topology orchestration</li>
+ * <li>admission policy</li>
+ * <li>tier placement decisions</li>
+ * <li>promotion or demotion across tiers</li>
+ * </ul>
+ */
+@InterfaceAudience.Private
+public interface CacheEngine {
+
+  /**
+   * Returns a human-readable name for this cache engine instance.
+   * <p>
+   * The name is intended for logging, metrics, diagnostics, and configuration reporting. It should
+   * be stable for the lifetime of the engine instance.
+   * </p>
+   * @return engine name
+   */
+  String getName();
+
+  /**
+   * Adds a block to the cache.
+   * @param cacheKey block cache key
+   * @param buf      block contents
+   * @param inMemory whether the block should be treated as in-memory
+   */
+  void cacheBlock(BlockCacheKey cacheKey, Cacheable buf, boolean inMemory);
+
+  /**
+   * Adds a block to the cache, optionally waiting for asynchronous cache backends.
+   * <p>
+   * This is primarily useful for implementations such as BucketCache that may buffer writes
+   * asynchronously.
+   * </p>
+   * @param cacheKey      block cache key
+   * @param buf           block contents
+   * @param inMemory      whether the block should be treated as in-memory
+   * @param waitWhenCache whether to wait for the cache operation to be accepted/flushed
+   */
+  default void cacheBlock(BlockCacheKey cacheKey, Cacheable buf, boolean inMemory,
+    boolean waitWhenCache) {
+    cacheBlock(cacheKey, buf, inMemory);
+  }
+
+  /**
+   * Adds a block to the cache, defaulting to non in-memory treatment.
+   * @param cacheKey block cache key
+   * @param buf      block contents
+   */
+  void cacheBlock(BlockCacheKey cacheKey, Cacheable buf);
+
+  /**
+   * Fetches a block from the cache.
+   * @param cacheKey           block to fetch
+   * @param caching            whether caching is enabled for the request; used for metrics
+   * @param repeat             whether this is a repeated lookup for the same block; used to avoid
+   *                           double-counting misses
+   * @param updateCacheMetrics whether cache metrics should be updated
+   * @return cached block, or {@code null} if not present
+   */
+  Cacheable getBlock(BlockCacheKey cacheKey, boolean caching, boolean repeat,
+    boolean updateCacheMetrics);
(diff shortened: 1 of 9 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngineView.java
```diff
@@ -0,0 +1,97 @@
...
+   */
+  public long getBlockCount() {
+    return engine.getBlockCount();
+  }
+
+  /** 
+   * Returns the number of blocks currently stored in the cache.
+   * @return  the number of blocks currently stored in the cache
+   */
+  public boolean canStore(Cacheable block) {
```
wchevreuil: nit: javadoc copy&paste error.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngineType.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /**
+   * CarrotCache-based engine.
+   */
+  CARROT,
+
+  /**
+   * Other or custom engine type.
+   */
+  CUSTOM
```
wchevreuil: So different custom implementations would all fall under this CUSTOM enum type?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
```diff
@@ -0,0 +1,319 @@
...
+ * <p>Non-responsibilities include:</p>
+ * <ul>
+ *   <li>L1/L2 topology orchestration</li>
+ *   <li>admission policy</li>
+ *   <li>tier placement decisions</li>
+ *   <li>promotion or demotion across tiers</li>
+ * </ul>
+ */
+@InterfaceAudience.Private
+public interface CacheEngine {
```
wchevreuil, line 63: Is this the future substitute for the current existing BlockCache interface?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
```diff
@@ -0,0 +1,319 @@
...
+   * @return free size
+   */
+  long getFreeSize();
+
+  /**
+   * Returns the currently occupied cache size, in bytes.
+   *
+   * @return occupied size
+   */
+  long getCurrentSize();
```
wchevreuil: I got a bit confused about what is the difference between these. Looking at current BlockCache implementations, my understanding is that "size" computes the "logical" data size in the cache, simply, the sum of all cached blocks length. 

MaxSize is the one that maps to the configured capacity. 

The "current size" is the real cache occupation, and in some implementations like BucketCache, it will usually be higher than the "size" due to some fragmentation.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
VladRodionov: Partially yes, but overall CacheAccessService will be an access point to a caching system. Future ticket.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngineType.java
VladRodionov: Its just a stub, any engine should has its own type. I will remove it. Its confusing.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
VladRodionov: getMaxSize is configuration, getCurrentSize is actual usage, and getFreeSize
is an engine-provided fast/accurate view of remaining capacity. These are not
interchangeable due to differences in allocation models across engines. size() does seem redundant - will remove it. Actually I will keep size() as a  replacement for getCurrentSize and will remove getCurrentSize

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopology.java
```diff
@@ -0,0 +1,177 @@
...
+  /**
+   * Returns a read-only view of this topology.
+   *
+   * <p>The returned view is intended for policy implementations, diagnostics, and metrics. It should
+   * expose topology and engine state without allowing callers to mutate cache contents or bypass
+   * topology behavior.</p>
+   *
+   * @return read-only topology view
+   */
+  CacheTopologyView getView();
```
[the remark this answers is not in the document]
VladRodionov: Good catch — you’re absolutely right.

Returning a package-private type from a public interface makes the method effectively unusable
outside the package and is not a valid API design.
The intent here was to keep CacheTopologyView / CacheEngineView as lightweight read-only
helpers for policy and internal use, without introducing additional abstraction via interfaces.
To fix this, I will make CacheTopologyView and CacheEngineView public while keeping them
annotated with @InterfaceAudience.Private to indicate that they are not intended for external
consumption.
This preserves the simple “view as class” design without leaking internal implementation details
or forcing an additional interface layer.
I’ll update the patch accordingly.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopologyView.java
```diff
@@ -0,0 +1,87 @@
...
+  /**
+   * Returns the list of cache tiers in this topology. For a single-tier topology, it returns a list
+   * containing only CacheTier.SINGLE. For a multi-tier topology, it returns a list containing
+   * CacheTier.L1 and CacheTier.L2.
+   * @return the list of cache tiers in this topology
+   */
+  public List<CacheTier> getTiers() {
+    return topology.getEngines().size() == 1
+      ? List.of(CacheTier.SINGLE)
+      : List.of(CacheTier.L1, CacheTier.L2);
```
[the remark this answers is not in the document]
VladRodionov: Good point. I removed CUSTOM topology, but I agree the current implementation is still too implicit.

Inferring tiers from getEngines().size() is brittle. I will add an explicit getTiers() method to CacheTopology and have CacheTopologyView delegate to it instead of deriving tiers from engine count.

That keeps tier ownership in the topology implementation and avoids assumptions in the view layer.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopologyView.java
```diff
@@ -0,0 +1,87 @@
...
+  
+  public String getName() {
+    return topology.getName();
+  }
+  /**
+   * Returns the cache topology type, which can be SINGLE, L1_L2, or CUSTOM.
+   * @return the cache topology type
+   */
+  public CacheTopologyType getType() {
+    return topology.getType();
```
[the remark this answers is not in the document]
VladRodionov: Done.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngineType.java
```diff
@@ -0,0 +1,41 @@
...
+
+  /**
+   * Bucket-based off-heap / file-backed cache engine.
+   */
+  BUCKET,
+
+  /**
+   * CarrotCache-based engine.
+   */
+  CARROT
```
taklwu: nit: will carrot be the dependencies of hbase? if not, maybe this could be removed?

or should we use interface factory with getName instead of enum?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
```diff
@@ -0,0 +1,302 @@
...
+   * @param cacheKey           block to fetch
+   * @param caching            whether caching is enabled for the request; used for metrics
+   * @param repeat             whether this is a repeated lookup for the same block; used to avoid
+   *                           double-counting misses
+   * @param updateCacheMetrics whether cache metrics should be updated
+   * @param blockType          optional block type hint
+   * @return cached block, or {@code null} if not present
+   */
+  default Cacheable getBlock(BlockCacheKey cacheKey, boolean caching, boolean repeat,
+    boolean updateCacheMetrics, BlockType blockType) {
```
taklwu, line 132: is this `BlockType blockType` an optimization for fast lookup?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
taklwu: yeah, I was about to ask some of the interface in `BlockCache` wasn't defined here, e.g. `iterator` and `getBlockCaches`, but let's see if the future task would tell us about those functions

taklwu (commented):
few questions, mostly it looks good.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngineType.java
VladRodionov: Good point. CarrotCache should not need to be a hard dependency of hbase-server for this foundational API.

I think using a fixed enum for engine types is too restrictive for a pluggable architecture anyway. It works for built-ins like LRU/Bucket, but does not scale well for optional or third-party engines.

I’ll remove CacheEngineType and rely on getName() / configuration identity instead. That keeps the core API independent of optional engines and avoids baking specific external implementations into HBase.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
VladRodionov: BlockType here is a hint rather than part of the lookup key.
It is used by some cache implementations to avoid re-inspecting the block payload
(e.g. distinguishing data vs index/meta blocks), enabling minor optimizations in
metrics, prioritization, or eviction behavior.

It is not required for correctness and does not affect lookup, which is based
solely on BlockCacheKey. Implementations are free to ignore it, which is why a
default method is provided. This will be moved later into Context object

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheEngine.java
VladRodionov: API will evolve for sure. We will see later if anything is missing.

wchevreuil (approved):

taklwu (approved):

PR #8156, opened 2026-04-28
HBASE-30119 Stop JmxCacheBuster in HBTU

Apache9:


Apache9:
@liuxiaocs7 I found that JmxCacheBuster is in our own code base so we can change its code...

I revert the previous PR, PTAL on this new one.

Thanks.

liuxiaocs7 (approved):

PR #8157, opened 2026-04-28
HBASE-30127 Improve Website's SEO

yuriipalam:
There is also a small change in the `hbase-site` repo, please take a look:
https://github.com/apache/hbase-site/pull/13

This change improves SEO (Search Engine Optimization) by adding proper cache headers, good looking 404 page, and sitemap.

404 page in dark mode:
image


hbase-website/public/sitemap.xml
PDavid, line 1: As I understand we have a script to generate this file. Can I ask why is this committed? I guess new builds will generate new version of this file with different timestamps so we will have local modifications on this file. 🤔 
Should we rather git ignore this file instead?

hbase-website/package.json
```diff
@@ -75,6 +76,7 @@
     "remark-gfm": "^4.0.1",
     "remark-math": "^6.0.0",
     "scroll-into-view-if-needed": "^3.1.0",
+    "sitemap": "^9.0.1",
```
PDavid: sitemap is added to `dependencies` rather than `devDependencies`. Since it's only used in the build script (not at runtime), I guess it should be a dev dependency.

PDavid (commented):
Many thanks, looks really nice already. 👍 

Just two small questions / comments.

PDavid:
It seems that the build complains about these files missing the ASF licence header:

> Files with unapproved licenses:
> 
>   public/robots.txt
>   public/sitemap.xml

hbase-website/package.json
yuriipalam: you're right. thanks

hbase-website/public/sitemap.xml
yuriipalam: Actually, yes. It sounds good for our workflow

PDavid (approved):
Many thanks, looks good to me. 👍

PDavid:
Some tests failed in the PR build but they are not because of the changes in this branch.

hbase-website/app/pages/_docs/docs/_mdx/(multi-page)/upgrading/version-number.mdx
```diff
@@ -149,6 +154,4 @@ When we say two HBase versions are compatible, we mean that the versions are wir
 
 [^2]: See http://docs.oracle.com/javase/specs/jls/se8/html/jls-13.html.
 
-[^3]: Note that this indicates what could break, not that it will break. We will/should add specifics in our release notes.
```
PDavid, line 152: Can I ask why did we remove this footnote from here? It seems to be unrelated to the website SEO to me.

yuriipalam: Because this footnote was added to the heading, what leads to a nested link problem.
In docs frameworks headings are links because they are anchors, same for footnotes.
So we had something like

```html
<a>
  <h4>Heading...</h4>
  <a>^3</a>
<a>
```

It's wrong HTML structure which leads to a hydration error. Due to the async nature of the e2e tests, we couldn't reliably catch this issue before. It might be a new task to work on for me actually, to make it more deterministic.

So, I just moved the footnote right under the heading since it's the only place where it's used. This way we fix the broken structure.

PDavid: Ah, I see, many thanks for the response. I overlooked that you moved the footnote up.

PDavid:
Hi @yuriipalam,
Since https://github.com/apache/hbase-site/pull/13 is related should we merge this PR first and then wait until the website is deployed or should we merge https://github.com/apache/hbase-site/pull/13 first or the order does not matter?

yuriipalam:
Hi @PDavid 
I think it's not important. We can just merge both of them at the same time. The order doesn't really matter.

PDavid:
Test failures in the PR build are unrelated since this is a website-only change.

PR #8179, opened 2026-04-30
HBASE-30102 Add metric to account for region data classified as cold by the Time Based Priority logic (#8128)

wchevreuil:


wchevreuil:
The reported failures are all unrelated flakeys.

PR #8158, opened 2026-04-28
HBASE-30128 Fix WALCellCodec off-heap KV tags compression test mismatch

liuxiaocs7:
- see: HBASE-30128

PR #8183, opened 2026-05-01
HBASE-30124 Addendum Upgrade hbase-server to use junit5 Part14

liuxiaocs7:
- see: HBASE-30124


hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/wal/TestWALCellCodecWithCompression.java
```diff
@@ -73,7 +73,7 @@ public void testEncodeDecodeKVsWithTagsWithTagsCompression() throws Exception {
 
   @TestTemplate
   public void testEncodeDecodeOffKVsWithTagsWithTagsCompression() throws Exception {
-    doTest(true, true);
+    doTest(true, false);
```
liuxiaocs7, line 76: will be updated in https://github.com/apache/hbase/pull/8158

Apache9 (approved):

PR #8184, opened 2026-05-01
HBASE-30020 Introduce cache placement and admission policy API

VladRodionov:
This PR targets the HBASE-30018 feature branch.

Introduces the cache placement/admission policy layer for the pluggable block cache architecture.

Adds:
- CachePlacementAdmissionPolicy
- AdmissionDecision
- AdmissionPriority
- TierDecision
- PromotionDecision
- RepresentationDecision
- CacheRequestContext
- CacheWriteContext
- CacheWriteSource
- DefaultHBaseCachePlacementAdmissionPolicy

The default policy preserves current HBase behavior:
- admission is allowed by default after existing CacheConfig decisions
- metadata/index/bloom blocks prefer L1, with fallback to L2
- data blocks prefer L2, with fallback to L1
- representation preserves current HBase behavior
- no promotion by default

No read/write path migration is included in this PR.
No behavior change intended.

JIRA: HBASE-30020

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/AdmissionDecision.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/AdmissionDecision.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.io.hfile.cache;
+
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * Result of cache admission evaluation.
+ */
+@InterfaceAudience.Private
+public final class AdmissionDecision {
+
+  private static final AdmissionDecision ADMIT = new AdmissionDecision(true, "ADMIT");
+
+  private static final AdmissionDecision REJECT = new AdmissionDecision(false, "REJECT");
+
+  private final boolean admit;
+  private final String reason;
+
+  private AdmissionDecision(boolean admit, String reason) {
+    this.admit = admit;
+    this.reason = reason;
+  }
+
+  /**
+   * Returns a normal-priority admission decision.
+   * @return admit decision
+   */
+  public static AdmissionDecision admit() {
+    return ADMIT;
+  }
+
+  /**
+   * Returns an admission decision with a reason.
+   * @param reason reason text
+   * @return admit decision
+   */
+  public static AdmissionDecision admit(String reason) {
+    return new AdmissionDecision(true, reason);
+  }
+
+  /**
+   * Returns a rejection decision. * @return reject decision
+   */
+  public static AdmissionDecision reject() {
+    return REJECT;
+  }
+
+  /**
+   * Returns a rejection decision with a reason.
+   * @param reason rejection reason
+   * @return reject decision
+   */
+  public static AdmissionDecision reject(String reason) {
+    return new AdmissionDecision(false, reason);
+  }
+
+  /**
+   * Returns whether the block should be admitted.
+   * @return true when admitted
+   */
+  public boolean isAdmitted() {
+    return admit;
+  }
+
+  /**
+   * Returns the decision reason.
+   * @return decision reason
+   */
+  public String getReason() {
+    return reason;
+  }
+
+}
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheRequestContext.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheRequestContext.java
@@ -1,0 +1,187 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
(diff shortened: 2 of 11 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/AdmissionDecision.java
```diff
@@ -0,0 +1,89 @@
...
+   * Returns an admission decision with a reason.
+   * @param reason reason text
+   * @return admit decision
+   */
+  public static AdmissionDecision admit(String reason) {
+    return new AdmissionDecision(true, reason);
+  }
+
+  /**
+   * Returns a rejection decision. * @return reject decision
```
petersomogyi:
```suggestion
   * Returns a rejection decision. 
   * @return reject decision
```

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/RepresentationDecision.java
```diff
@@ -0,0 +1,63 @@
...
+  /**
+   * Returns a representation decision for the supplied representation kind.
+   * @param kind representation kind
+   * @return representation decision
+   */
+  public static RepresentationDecision of(RepresentationKind kind) {
+    return kind == RepresentationKind.CURRENT_HBASE_DEFAULT
+      ? CURRENT_HBASE_DEFAULT
+      : new RepresentationDecision(kind);
+  }
```
petersomogyi: I'm not sure if this will be on a hot path but if so then you can cache all 4 RepresentationKind instances since the enum is small and fixed.

petersomogyi (approved):
LGTM. Clean API design with good Javadocs.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheWriteContext.java
```diff
@@ -0,0 +1,187 @@
...
+public final class CacheWriteContext {
+
+  private final boolean inMemory;
+  private final boolean waitWhenCache;
+  private final boolean cacheCompressed;
+  private final boolean cacheOnWrite;
+  private final BlockCategory blockCategory;
+  private final CacheWriteSource source;
+
+  private CacheWriteContext(Builder builder) {
```
wchevreuil, line 39: Currently, we have also a separate "cacheCompactedDataOnWrite", which governs cache on write behaviour separately for compaction writes. Should we keep that behaviour here?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/RepresentationDecision.java
wchevreuil: This class seems redundant with RepresentationKind, is it planned to implement any additional logic in the future? Otherwise, can these two just be merged?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheWriteContext.java
VladRodionov: I think "cacheCompactedDataOnWrite" and "cacheOnWrite" in general are redundant if CacheWriteSource already captures the insertion origin. The source tells us whether the insertion came from READ_MISS, FLUSH, COMPACTION, PREFETCH, or PROMOTION. Existing CacheConfig logic should already decide whether cache population is allowed before this context is created.
So the policy should not use cacheOnWrite and cacheCompactedDataOnWrite as a second eligibility gate. For compatibility behavior, it should assume that the caller invokes cache insertion only after CacheConfig has allowed it. I’ll remove cacheOnWrite from CacheWriteContext and rely on CacheWriteSource plus existing call-site logic.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/RepresentationDecision.java
VladRodionov: Yep, looks over engineered. Will remove it.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheRequestContext.java
```diff
@@ -0,0 +1,187 @@
...
+    return caching;
+  }
+
+  /**
+   * Returns whether this is a repeated lookup for the same block.
+   * @return true when this is a repeated lookup
+   */
+  public boolean isRepeat() {
+    return repeat;
+  }
```
taklwu, line 70: question:  how would this repeat be used ? is it something we will define in the each implementation that more than X times lookup would considered as repeat?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/PromotionAction.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package org.apache.hadoop.hbase.io.hfile.cache;
+
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * Promotion action selected by cache placement policy.
+ */
+@InterfaceAudience.Private
+public enum PromotionAction {
```
taklwu, line 26: nit: will we have a demotion action ? IIRC that we may have block moving from L1 -> L2, ,if so and if need to record it, where should we action be annotated ?

taklwu (approved):
just two questions , but LGTM

wchevreuil (approved):
LGTM.

VladRodionov:
Thank you, @taklwu @wchevreuil and @petersomogyi.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/PromotionAction.java
VladRodionov: I think demotion is a separate topology operation and should not be encoded in PromotionAction.
PromotionAction is intentionally scoped to cache-hit handling: after a hit in one tier, the policy may decide whether to promote the block to a higher-priority tier. Demotion, if needed later, would likely be modeled separately, for example as DemotionDecision / DemotionAction, or recorded in topology-level metrics when CacheTopology#demote(...) is invoked. I would avoid mixing promotion and demotion into the same enum because they are triggered by different events: promotion is hit-driven, while demotion is usually eviction/pressure-driven. Besides this, demotion is quite expensive (you will need to expose CacheEngine internal eviction pipeline to listen to these events, not all cache implementations can be efficient)

For this PR, I’d keep PromotionAction limited to NONE and PROMOTE, and add demotion-specific decision/metrics later when we wire actual eviction callbacks or topology movement.

Btw, promotion is hard as well, you have to make sure that object being promoted is really worth it, otherwise you introduce additional layer of cache data trashing.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheRequestContext.java
VladRodionov: `repeat` preserves an existing HBase cache lookup semantic and is not intended as a hotness/frequency signal. Today HBase uses it mainly to avoid double-counting cache misses during retry or double-check-locking lookup paths.

PR #8185, opened 2026-05-03
HBASE-30136 Upgrade hbase-server to use junit5 Part15

liuxiaocs7:
- see HBASE-30136

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCacheOnWriteInSchema.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCacheOnWriteInSchema.java
@@ -17,17 +17,15 @@
  */
 package org.apache.hadoop.hbase.regionserver;
 
-import static org.junit.Assert.assertTrue;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
 import java.util.Random;
+import java.util.stream.Stream;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
-import org.apache.hadoop.hbase.HBaseClassTestRule;
+import org.apache.hadoop.hbase.HBaseParameterizedTestTemplate;
 import org.apache.hadoop.hbase.HBaseTestingUtil;
 import org.apache.hadoop.hbase.KeyValue;
 import org.apache.hadoop.hbase.TableName;
@@ -53,16 +51,12 @@ import org.apache.hadoop.hbase.testclassification.SmallTests;
 import org.apache.hadoop.hbase.util.Bytes;
 import org.apache.hadoop.hbase.util.CommonFSUtils;
 import org.apache.hadoop.hbase.wal.AbstractFSWALProvider;
-import org.junit.After;
-import org.junit.Before;
-import org.junit.ClassRule;
-import org.junit.Rule;
-import org.junit.Test;
-import org.junit.experimental.categories.Category;
-import org.junit.rules.TestName;
-import org.junit.runner.RunWith;
-import org.junit.runners.Parameterized;
-import org.junit.runners.Parameterized.Parameters;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.TestInfo;
+import org.junit.jupiter.api.TestTemplate;
+import org.junit.jupiter.params.provider.Arguments;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -70,17 +64,13 @@ import org.slf4j.LoggerFactory;
  * Tests {@link HFile} cache-on-write functionality for data blocks, non-root index blocks, and
  * Bloom filter blocks, as specified by the column family.
  */
-@RunWith(Parameterized.class)
-@Category({ RegionServerTests.class, SmallTests.class })
+@Tag(RegionServerTests.TAG)
+@Tag(SmallTests.TAG)
+@HBaseParameterizedTestTemplate(name = "{index}: cacheOnWrite={0}")
 public class TestCacheOnWriteInSchema {
 
-  @ClassRule
-  public static final HBaseClassTestRule CLASS_RULE =
-    HBaseClassTestRule.forClass(TestCacheOnWriteInSchema.class);
-
   private static final Logger LOG = LoggerFactory.getLogger(TestCacheOnWriteInSchema.class);
-  @Rule
-  public TestName name = new TestName();
+  private String name;
 
   private static final HBaseTestingUtil TEST_UTIL = new HBaseTestingUtil();
   private static final String DIR = TEST_UTIL.getDataTestDir("TestCacheOnWriteInSchema").toString();
@@ -141,19 +131,15 @@ public class TestCacheOnWriteInSchema {
     System.out.println(testDescription);
   }
 
-  @Parameters
-  public static Collection<Object[]> getParameters() {
-    List<Object[]> cowTypes = new ArrayList<>();
-    for (CacheOnWriteType cowType : CacheOnWriteType.values()) {
-      cowTypes.add(new Object[] { cowType });
-    }
-    return cowTypes;
+  public static Stream<Arguments> parameters() {
+    return Stream.of(CacheOnWriteType.values()).map(Arguments::of);
   }
 
-  @Before
-  public void setUp() throws IOException {
+  @BeforeEach
+  public void setUp(TestInfo testInfo) throws IOException {
+    this.name = testInfo.getTestMethod().get().getName();
     // parameterized tests add [#] suffix get rid of [ and ].
-    table = Bytes.toBytes(name.getMethodName().replaceAll("[\\[\\]]", "_"));
+    table = Bytes.toBytes(name.replaceAll("[\\[\\]]", "_"));
 
     conf = TEST_UTIL.getConfiguration();
     conf.setInt(HFile.FORMAT_VERSION_KEY, HFile.MAX_FORMAT_VERSION);
@@ -183,7 +169,7 @@ public class TestCacheOnWriteInSchema {
     store = region.getStore(hcd.getName());
   }
 
-  @After
+  @AfterEach
   public void tearDown() throws IOException {
     IOException ex = null;
     try {
@@ -203,7 +189,7 @@ public class TestCacheOnWriteInSchema {
     }
   }
 
-  @Test
+  @TestTemplate
   public void testCacheOnWriteInSchema() throws IOException {
     // Write some random data into the store
     StoreFileWriter writer = store.getStoreEngine()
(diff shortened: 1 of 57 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactorMemLeak.java
```diff
@@ -89,16 +88,16 @@ public static void tearDown() throws Exception {
...
     List<HRegion> regions = UTIL.getHBaseCluster().getRegionServerThreads().get(0).getRegionServer()
       .getRegions(tableName);
-    Assert.assertEquals(regions.size(), 1);
+    assertEquals(regions.size(), 1);
     HRegion region = regions.get(0);
-    Assert.assertEquals(region.getStores().size(), 1);
+    assertEquals(region.getStores().size(), 1);
     HStore store = region.getStore(FAMILY);
-    Assert.assertEquals(store.getStorefilesCount(), 1);
+    assertEquals(store.getStorefilesCount(), 1);
```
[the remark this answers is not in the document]
liuxiaocs7: Keep it consistent with before

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCleanupMetaWAL.java
```diff
@@ -31,32 +30,28 @@
...
+  @BeforeAll
   public static void before() throws Exception {
     TEST_UTIL.startMiniCluster(2);
   }
 
-  @AfterClass
+  @AfterAll
   public static void after() throws Exception {
-    TEST_UTIL.shutdownMiniZKCluster();
+    TEST_UTIL.shutdownMiniCluster();
```
Apache9, line 54: OK, we fixed an old bug here...

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestClearRegionBlockCache.java
```diff
@@ -95,12 +92,12 @@ public void setup() throws Exception {
     HTU.flush(TABLE_NAME);
   }
 
-  @After
+  @AfterEach
   public void teardown() throws Exception {
     HTU.shutdownMiniCluster();
```
Apache9, line 97: This means we need to restart the cluster for every test method? Can file another issue to dig more, maybe we should split this to two tests, where we start cluster with different cacheType.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactingToCellFlatMapMemStore.java
```diff
@@ -73,13 +74,9 @@ public TestCompactingToCellFlatMapMemStore(String type) {
...
   @Override
-  public void tearDown() throws Exception {
-    chunkCreator.clearChunksInPool();
-    super.tearDown();
-  }
-
-  @Override
-  public void setUp() throws Exception {
+  @BeforeEach
+  public void setUp(TestInfo testInfo) throws Exception {
```
Apache9: Is the behavior expected? Better introduce some dummy abstract method and call it in the parent test class, and implement the dummy method in this sub class, so it will be more clear?

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompaction.java
```diff
@@ -747,7 +742,7 @@ public List<Path> compact(ThroughputController throughputController, User user)
             this.wait();
           }
         } catch (InterruptedException e) {
-          Assume.assumeNoException(e);
+          Assumptions.abort(e.getMessage());
```
Apache9, line 745: This is a bit strange, anyway, can be a separated issue.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactionFileNotFound.java
```diff
@@ -193,9 +187,9 @@ public void testCompactionAfterRefresh() throws Exception {
       } catch (IOException e) {
         LOG.error("Got an exception during compaction", e);
         if (e instanceof FileNotFoundException) {
-          Assert.fail("Got a FNFE during compaction");
+          fail("Got a FNFE during compaction");
```
Apache9: A bit strange that why not just throw the exception out...

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactSplitThread.java
```diff
@@ -123,7 +117,7 @@ public void testThreadPoolSizeTuning() throws Exception {
       try {
         regionServer.getCompactSplitThread().onConfigurationChange(conf);
       } catch (IllegalArgumentException iae) {
-        Assert.fail("Update bigger configuration failed!");
+        fail("Update bigger configuration failed!");
```
Apache9: It is not good that we eat the exception information...

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestDefaultMemStore.java
```diff
@@ -1083,10 +1072,10 @@ private long runSnapshot(final AbstractMemStore hmc) throws UnexpectedStateExcep
     int oldHistorySize = hmc.getSnapshot().getCellsCount();
     MemStoreSnapshot snapshot = hmc.snapshot();
     // Make some assertions about what just happened.
-    assertTrue("History size has not increased",
-      oldHistorySize < hmc.getSnapshot().getCellsCount());
+    assertTrue(oldHistorySize < hmc.getSnapshot().getCellsCount(),
+      "History size has not increased");
     long t = memstore.timeOfOldestEdit();
-    assertTrue("Time of oldest edit is not Long.MAX_VALUE", t == Long.MAX_VALUE);
+    assertTrue(t == Long.MAX_VALUE, "Time of oldest edit is not Long.MAX_VALUE");
```
Apache9: assertEquals

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestClearRegionBlockCache.java
liuxiaocs7: I agree this is worth improving separately. We can file a follow-up issue to investigate splitting this into separate tests for different cache types, thanks!

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactingToCellFlatMapMemStore.java
liuxiaocs7: Good point. I refactored this further so that the JUnit lifecycle method is only owned by TestDefaultMemStore.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompaction.java
liuxiaocs7: Will file another issue to resolve this problem

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestDefaultMemStore.java
liuxiaocs7: Done!

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactionFileNotFound.java
liuxiaocs7: make sense, fixed

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactSplitThread.java
liuxiaocs7: Agreed. I changed this to `assertDoesNotThrow `so we keep the original exception as the cause if the update fails, instead of swallowing it with `fail(message)`

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCacheOnWriteInSchema.java
```diff
@@ -141,19 +131,15 @@ public TestCacheOnWriteInSchema(CacheOnWriteType cowType) {
...
   }
 
-  @Before
-  public void setUp() throws IOException {
+  @BeforeEach
+  public void setUp(TestInfo testInfo) throws IOException {
+    this.name = testInfo.getTestMethod().get().getName();
     // parameterized tests add [#] suffix get rid of [ and ].
-    table = Bytes.toBytes(name.getMethodName().replaceAll("[\\[\\]]", "_"));
+    table = Bytes.toBytes(name.replaceAll("[\\[\\]]", "_"));
```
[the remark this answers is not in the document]
liuxiaocs7: Done

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactingMemStore.java
```diff
@@ -59,49 +58,37 @@
...
   private static final Logger LOG = LoggerFactory.getLogger(TestCompactingMemStore.class);
   protected static ChunkCreator chunkCreator;
   protected HRegion region;
   protected RegionServicesForStores regionServicesForStores;
   protected HStore store;
+  private Configuration conf;
 
-  @After
+  @AfterEach
   public void tearDown() throws Exception {
```
Apache9: Better use the same pattern with BeforeEach?

Apache9: PTAL at this one? @liuxiaocs7

Apache9:
Any updates here? @liuxiaocs7 

Thanks.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactingMemStore.java
liuxiaocs7: I'm very sorry for missing the message, I'll check it now, thanks for the comments

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestCompactingMemStore.java
```diff
@@ -59,49 +58,35 @@
...
   protected static ChunkCreator chunkCreator;
   protected HRegion region;
   protected RegionServicesForStores regionServicesForStores;
   protected HStore store;
+  private Configuration conf;
 
-  @After
-  public void tearDown() throws Exception {
+  @Override
+  protected void internalTearDown() throws Exception {
```
liuxiaocs7, line 82: this is the addition

liuxiaocs7:
> Any updates here? @liuxiaocs7
> 
> Thanks.

Sorry for the late reply, please help take a look when free, other separate issues will continue to be addressed after this PR is merged, thanks!

Apache9 (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestClearRegionBlockCache.java
liuxiaocs7: see: HBASE-30204

PR #8196, opened 2026-05-05
HBASE-30090: Table on replica cluster not refreshing after flipping r…

Kota-SH:
Jira: https://issues.apache.org/jira/browse/HBASE-30090


anmolnar (approved):
lgtm.

PR #8176, opened 2026-04-30
HBASE-30123 Upgrade hbase-server to use junit5 Part13 (#8153)

Apache9:
(cherry picked from commit eae4887cd41dc180e02f77bfb8522372a39a6f87)

PR #8153, opened 2026-04-27
HBASE-30123 Upgrade hbase-server to use junit5 Part13

Apache9:

PR #8224, opened 2026-05-12
HBASE-30123 Upgrade hbase-server to use junit5 Part13 (#8153) (#8176)

Apache9:
(cherry picked from commit eae4887cd41dc180e02f77bfb8522372a39a6f87)


Apache9:
I've tried locally, TestCellBasedImportExport2 can always pass.

I downloaded the log from github ci, the failure is because of ConnectionLoss on ZooKeeper, which is very strange...

PR #8231, opened 2026-05-12
HBASE-30021 Introduce cache access service API

VladRodionov:
## Summary

This PR introduces `CacheAccessService`, an HBase-facing cache access abstraction for block cache read/write path callers.

The new service layer preserves the core operational shape of the existing `BlockCache` API while moving callers toward context-based cache access using `CacheRequestContext` and `CacheWriteContext`.

## Changes

- Adds `CacheAccessService`
- Adds `BlockCacheBackedCacheAccessService` as a compatibility adapter over existing `BlockCache`
- Adds `NoOpCacheAccessService` for disabled-cache wiring
- Adds `CacheAccessServices` helper factory methods
- Adds unit tests for block-cache-backed and disabled-cache behavior
- Adds `TopologyBackedCacheAccessService` to connect the service layer to `CacheTopology`, `CacheEngine`, and `CachePlacementAdmissionPolicy`
- Adds tests for topology-backed lookup, insertion, eviction, sizing, and factory behavior

## Notes

This PR does not migrate `HFileReaderImpl`, `BlockCacheFactory`, or write-path callers yet. It only introduces the service abstraction and compatibility adapter so later PRs can migrate call sites incrementally.

No behavior change is intended for existing cache implementations.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/BlockCacheBackedCacheAccessService.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/BlockCacheBackedCacheAccessService.java
@@ -1,0 +1,298 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.io.hfile.cache;
+
+import java.util.Objects;
+import java.util.Optional;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.hbase.io.hfile.BlockCache;
+import org.apache.hadoop.hbase.io.hfile.BlockCacheKey;
+import org.apache.hadoop.hbase.io.hfile.BlockType;
+import org.apache.hadoop.hbase.io.hfile.CacheStats;
+import org.apache.hadoop.hbase.io.hfile.Cacheable;
+import org.apache.hadoop.hbase.io.hfile.HFileBlock;
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * {@link CacheAccessService} implementation backed by an existing {@link BlockCache} instance.
+ * <p>
+ * This adapter is the compatibility bridge for the first migration step. It allows new callers to
+ * depend on {@link CacheAccessService} while the runtime implementation still uses the current
+ * {@link BlockCache} hierarchy, including LruBlockCache, BucketCache, CombinedBlockCache, and other
+ * existing implementations.
+ * </p>
+ * <p>
+ * The adapter should not introduce new policy, placement, admission, representation, promotion, or
+ * topology behavior. Its purpose is to translate the new context-based service API into the current
+ * {@code BlockCache} API with no intentional behavior change.
+ * </p>
+ * <p>
+ * A future topology-backed service can replace this adapter after call sites have migrated to
+ * {@link CacheAccessService}. That future implementation may use {@link CacheTopology},
+ * {@link CachePlacementAdmissionPolicy}, and {@link CacheEngine}; this adapter deliberately does
+ * not.
+ * </p>
+ */
+@InterfaceAudience.Private
+public class BlockCacheBackedCacheAccessService implements CacheAccessService {
+
+  private final BlockCache blockCache;
+
+  /**
+   * Creates a cache access service backed by the supplied legacy block cache.
+   * @param blockCache block cache to wrap
+   */
+  public BlockCacheBackedCacheAccessService(BlockCache blockCache) {
+    this.blockCache = Objects.requireNonNull(blockCache, "blockCache must not be null");
+  }
+
+  /**
+   * Returns the wrapped {@link BlockCache} instance.
+   * <p>
+   * This accessor is intended for tests and transitional wiring only. New read/write path code
+   * should use {@link CacheAccessService} methods instead of unwrapping the legacy cache.
+   * </p>
+   * @return wrapped block cache
+   */
+  public BlockCache getBlockCache() {
+    return blockCache;
+  }
+
+  /**
+   * Returns a human-readable service name.
+   * @return service name
+   */
+  @Override
+  public String getName() {
+    return blockCache.getClass().getSimpleName();
+  }
+
+  /**
+   * Fetches a block by delegating to the wrapped {@link BlockCache}.
+   * @param cacheKey block to fetch
+   * @param context  cache request context
+   * @return cached block, or {@code null} if not present
+   */
+  @Override
+  public Cacheable getBlock(BlockCacheKey cacheKey, CacheRequestContext context) {
+    Objects.requireNonNull(cacheKey, "cacheKey must not be null");
+    Objects.requireNonNull(context, "context must not be null");
+
+    Optional<BlockType> blockType = context.getBlockType();
+    if (blockType.isPresent()) {
+      return blockCache.getBlock(cacheKey, context.isCaching(), context.isRepeat(),
+        context.isUpdateCacheMetrics(), blockType.get());
+    }
+    return blockCache.getBlock(cacheKey, context.isCaching(), context.isRepeat(),
+      context.isUpdateCacheMetrics());
+  }
+
+  /**
+   * Caches a block by delegating to the wrapped {@link BlockCache}.
+   * @param cacheKey block cache key
+   * @param block    block contents
+   * @param context  cache write context
+   */
+  @Override
+  public void cacheBlock(BlockCacheKey cacheKey, Cacheable block, CacheWriteContext context) {
+    Objects.requireNonNull(cacheKey, "cacheKey must not be null");
+    Objects.requireNonNull(block, "block must not be null");
+    Objects.requireNonNull(context, "context must not be null");
+
+    blockCache.cacheBlock(cacheKey, block, context.isInMemory(), context.isWaitWhenCache());
+  }
+
+  /**
+   * Evicts a single block by delegating to the wrapped {@link BlockCache}.
+   * @param cacheKey block to remove
+   * @return {@code true} if the block existed and was removed, {@code false} otherwise
+   */
+  @Override
+  public boolean evictBlock(BlockCacheKey cacheKey) {
+    Objects.requireNonNull(cacheKey, "cacheKey must not be null");
+    return blockCache.evictBlock(cacheKey);
+  }
+
+  /**
+   * Evicts all cached blocks for the given HFile by delegating to the wrapped {@link BlockCache}.
+   * @param hfileName HFile name
+   * @return number of blocks removed
+   */
+  @Override
+  public int evictBlocksByHfileName(String hfileName) {
+    Objects.requireNonNull(hfileName, "hfileName must not be null");
+    return blockCache.evictBlocksByHfileName(hfileName);
+  }
+
+  /**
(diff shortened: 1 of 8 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/BlockCacheBackedCacheAccessService.java
```diff
@@ -0,0 +1,298 @@
...
+
+  /**
+   * Evicts cached blocks for a region if the wrapped cache supports it.
+   * @param regionName region name
+   * @return number of blocks removed
+   */
+  @Override
+  public int evictBlocksByRegionName(String regionName) {
+    return 0; // BlockCache does not support region-based eviction, so we return 0 to indicate no
+              // blocks removed
```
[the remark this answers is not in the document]
taklwu: nit: please fix the comment line above the return line.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/NoOpCacheAccessService.java
```diff
@@ -0,0 +1,216 @@
...
+ * no-ops.
+ * </p>
+ * <p>
+ * This implementation should not be used to hide configuration mistakes. It represents an explicit
+ * disabled-cache state and should be selected only when the caller has determined that no block
+ * cache is available or desired.
+ * </p>
+ */
+@InterfaceAudience.Private
+public final class NoOpCacheAccessService implements CacheAccessService {
```
taklwu, line 49: nit: is this class used for testing purpose?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/TopologyBackedCacheAccessService.java
```diff
@@ -0,0 +1,476 @@
...
+
+    AdmissionDecision admission =
+      policy.shouldAdmit(cacheKey, block, context, AdmissionPriority.NORMAL, topologyView);
+    if (!admission.isAdmitted()) {
+      return;
+    }
+
+    policy.selectRepresentation(cacheKey, block, context, topologyView);
+    TierDecision tierDecision = policy.selectTier(cacheKey, block, context, topologyView);
+    for (CacheTier tier : tierDecision.getTiers()) {
```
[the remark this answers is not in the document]
taklwu: should be add a comment that `selectRepresentation` would be default to `CURRENT_HBASE_DEFAULT` ? or even remove this line of `selectRepresentation` ?

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/cache/TestBlockCacheBackedCacheAccessService.java
```diff
@@ -0,0 +1,210 @@
...
+  }
+
+  /**
+   * Verifies that invalidation methods delegate to the wrapped block cache.
+   */
+  @Test
+  void testEvictionDelegatesToBlockCache() {
+    BlockCache blockCache = mock(BlockCache.class);
+    CacheAccessService service = new BlockCacheBackedCacheAccessService(blockCache);
+    BlockCacheKey key = new BlockCacheKey("file", 1L);
```
taklwu: nit: make `file` and `1L` into parameters and avoid typos in the future ? or even move `file` to class level, since may test are using the same filename
```suggestion
    String fileName = "file";
    long offset = 1L;
    BlockCacheKey key = new BlockCacheKey(fileName, offset);
```

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/BlockCacheBackedCacheAccessService.java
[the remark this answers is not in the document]
VladRodionov: Done

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/cache/TestTopologyBackedCacheAccessService.java
```diff
@@ -0,0 +1,414 @@
...
+    CachePlacementAdmissionPolicy policy = mock(CachePlacementAdmissionPolicy.class);
+    CacheEngine engine = mock(CacheEngine.class);
+    CacheWriteContext context = writeContext();
+
+    when(topology.getView()).thenReturn(topologyView);
+    when(topology.getEngine(CacheTier.SINGLE)).thenReturn(Optional.of(engine));
+    when(policy.shouldAdmit(key, block, context, AdmissionPriority.NORMAL, topologyView))
+      .thenReturn(AdmissionDecision.admit());
+    when(policy.selectRepresentation(key, block, context, topologyView))
+      .thenReturn(RepresentationDecision.CURRENT_HBASE_DEFAULT);
```
taklwu, line 167: nit: I assumed this is just a mock call and the representation isn't used at the moment, will we have another alternative `TopologyBackedCacheAccessService` for other implementation? 

also, can you briefly explain how `TopologyBackedCacheAccessService` or `BlockCacheBackedCacheAccessService` could be configured in future ?

taklwu (changes requested):

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/TopologyBackedCacheAccessService.java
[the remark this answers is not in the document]
VladRodionov: I removed this selection and added this comment to JavaDoc: 
 Representation selection is intentionally not invoked by this initial implementation. Until the
 service can actually apply representation decisions safely, especially around HFileBlock
 lifecycle and packed/unpacked storage, representation policy is left to a later integration step.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/TopologyBackedCacheAccessService.java
```diff
@@ -0,0 +1,476 @@
...
+    }
+    return engine.getBlock(cacheKey, context.isCaching(), context.isRepeat(),
+      context.isUpdateCacheMetrics());
+  }
+
+  private void maybePromote(BlockCacheKey cacheKey, Cacheable block, CacheTier sourceTier,
+    CacheEngine sourceEngine, CacheRequestContext context) {
+    PromotionDecision decision =
+      policy.shouldPromote(cacheKey, block, sourceTier, context, topologyView);
+    if (decision == null || !decision.shouldPromote()) {
```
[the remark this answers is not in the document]
VladRodionov: Fixed

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/NoOpCacheAccessService.java
VladRodionov: The intent is to have a null-object implementation for the disabled-cache case.
Once read/write path callers start depending on CacheAccessService, having a disabled implementation lets the runtime wire a non-null service even when block cache is disabled or unavailable. That keeps future call sites simpler because they can call the service uniformly instead of carrying repeated null checks around BlockCache.

So instead of patterns like:

```java 
if (blockCache != null) {   blockCache.getBlock(...); } 
```
the wiring can select either:

```java 
CacheAccessServices.fromBlockCache(blockCache) 
```
or:

```java 
CacheAccessServices.disabled() 
```
and callers can use CacheAccessService consistently.
It is not strictly required for the first adapter PR. If you think it adds unnecessary surface area before we wire CacheAccessService into runtime code, I can remove NoOpCacheAccessService and the disabled() helper from this PR and add it later when the first disabled-cache wiring needs it.

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/cache/TestTopologyBackedCacheAccessService.java
VladRodionov: Good question. No, I do not expect multiple TopologyBackedCacheAccessService implementations per cache engine.

The intended model is:

```text 
CacheAccessService
  -> CacheTopology
      -> CacheEngine(s)
  -> CachePlacementAdmissionPolicy
``` 

So TopologyBackedCacheAccessService would be the common topology-based service. Different implementations plug in below it as different CacheEngine implementations, and different behavior plugs in as different CachePlacementAdmissionPolicy implementations.

For example, future wiring could choose either:

```text
 BlockCacheBackedCacheAccessService -> existing BlockCache 
```
for compatibility mode, or:

```text
TopologyBackedCacheAccessService
  -> TieredExclusiveTopology
  -> LruCacheEngine + BucketCacheEngine
  -> Default policy
```
and later swap only the engine/policy, for example:

```text
 LruCacheEngine + CarrotCacheEngine Carrot-aware policy
``` 

This PR does not wire runtime configuration yet, but future configuration would likely happen in BlockCacheFactory, using settings such as service type, topology type, L1/L2 engines, and policy.

Example:
```properties
hbase.blockcache.access.service = topology
hbase.blockcache.topology = tiered-exclusive
hbase.blockcache.l1.engine = lru
hbase.blockcache.l2.engine = bucket
hbase.blockcache.policy = default
```

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/NoOpCacheAccessService.java
```diff
@@ -0,0 +1,216 @@
...
+  /**
+   * Always returns {@code null} because this service does not store blocks.
+   * @param cacheKey block to fetch
+   * @param context  cache request context
+   * @return always {@code null}
+   */
+  @Override
+  public Cacheable getBlock(BlockCacheKey cacheKey, CacheRequestContext context) {
+    return null;
+  }
```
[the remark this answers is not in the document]
VladRodionov: Fixed.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/TopologyBackedCacheAccessService.java
```diff
@@ -0,0 +1,476 @@
...
+   * @param hfileName  HFile name
+   * @param initOffset inclusive start offset
+   * @param endOffset  inclusive end offset
+   * @return total number of blocks removed across all engines
+   */
+  @Override
+  public int evictBlocksRangeByHfileName(String hfileName, long initOffset, long endOffset) {
+    Objects.requireNonNull(hfileName, "hfileName must not be null");
+
+    int evicted = 0;
```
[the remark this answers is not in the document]
VladRodionov: Done.

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/cache/TestBlockCacheBackedCacheAccessService.java
VladRodionov: Good point. I’ll replace the repeated string/offset literals with class-level constants so the test is less typo-prone.

Kota-SH (approved):
lgtm

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/cache/TestTopologyBackedCacheAccessService.java
```diff
@@ -0,0 +1,464 @@
...
+  void testEvictBlocksRangeByHfileNameSumsEngineResults() {
+    CacheTopology topology = mock(CacheTopology.class);
+    CacheTopologyView topologyView = mock(CacheTopologyView.class);
+    CachePlacementAdmissionPolicy policy = mock(CachePlacementAdmissionPolicy.class);
+    CacheEngine l1 = mock(CacheEngine.class);
+    CacheEngine l2 = mock(CacheEngine.class);
+
+    when(topology.getView()).thenReturn(topologyView);
+    when(topology.getEngines()).thenReturn(List.of(l1, l2));
+    when(l1.evictBlocksRangeByHfileName("file", 10L, 100L)).thenReturn(2);
```
taklwu: nit: for `"file"` I forgot to mention this test class also need to replace this inline plaintext with parameters.

taklwu (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/cache/TestTopologyBackedCacheAccessService.java
VladRodionov: Done.

taklwu:
TestPrefetchWithBucketCache failed but it's related to https://github.com/apache/hbase/pull/8198

PR #8233, opened 2026-05-13
HBASE-30154: Fix hlink missing backreference causing dataloss

shrinidhi-arista:
Fix includes -

- RestoreSnapshotHelper: write a back-reference to the archive when creating an
HFileLink Reference file, so HFileLinkCleaner knows the archived HFile is still live.

Added test TestCloneSnapshotFromClientAfterMergingRegion which fails with FileNotFound
exception while checking if all the hfiles referenced by cloned table are available. Disabling and
enabling the table also causes the test to hang because of master retrying indefinitely. The test
passes with the fix.

- HFileLinkCleaner: when evaluating a back-reference, also check for HFileLink Reference
files (glob .*) in the link directory — these are created by
restoreReferenceFile() and protect the archived HFile even when no zero-byte HFileLink
file exists alongside them.

Added testHFileLinkReferenceFileProtectsArchivedHFile() which fails while asserting that Hfile
is not cleaned up but passes with the fix.


hbase-server/src/main/java/org/apache/hadoop/hbase/master/cleaner/HFileLinkCleaner.java
```diff
@@ -79,7 +79,17 @@ public boolean isFileDeletable(FileStatus fStat) {
...
           hfilePath =
             HFileLink.getHFileFromBackReference(CommonFSUtils.getRootDir(getConf()), filePath);
-          return !fs.exists(hfilePath);
+          if (fs.exists(hfilePath)) {
+            return false;
+          }
+          // Also protect HFileLink Reference files created by
+          // RestoreSnapshotHelper.restoreReferenceFile(). These are named
+          // <hfileLinkName>.<encodedRegion> and live in the same directory as the
+          // zero-byte HFileLink. The zero-byte file does not exist (only the Reference
```
Apache9: 'The zero-byte file does not exist', mind explaining more about this? I do not fully understand...

shrinidhi-arista: Updated the comment to make it more clear. What I mean is that when RestoreSnapshotHelper.restoreReferenceFile() clones a snapshot that contains a merged region, instead of creating zero-byte Hfile link of the form <table>=<region>-<hfile> , it creates a reference of the form <table>=<region>-<hfile>.<encodedRegion> (that that points through an HfileLink ) in the same directory. fs.exists() only checks for the zero-byte HfileLink and not the reference and hence that check is not sufficient to protect the references.

PR #8236, opened 2026-05-14
HBASE-30161 Add paginated, single-RPC RegionLocator.getRegionLocations(startKey, limit) API for bulk meta-cache warmup

sanjeet006py:
JIRA: [HBASE-30161](https://issues.apache.org/jira/browse/HBASE-30161)

Generated-by: Claude Opus 4.7


hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotRegionLocator.java
```diff
@@ -119,6 +119,12 @@ public List<HRegionLocation> getAllRegionLocations() throws IOException {
     return rawLocations;
   }
 
+  @Override
+  public List<HRegionLocation> getRegionLocations(byte[] startKey, int limit) throws IOException {
+    // No need to page as region locations are already in-memory.
+    return getAllRegionLocations();
+  }
```
haridsv: This is actually breaking the contract established by the javadoc by always returning all regions.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionLocator.java
```diff
@@ -130,6 +131,44 @@ default List<HRegionLocation> getRegionLocations(byte[] row) throws IOException
...
+   * region-cache warmup.
+   * @param startKey region start-key to begin scanning from (inclusive); {@code null} or empty
+   *                 starts from the first region
+   * @param limit    maximum number of regions to return; if &lt;= 0, falls back to
+   *                 {@code hbase.meta.scanner.caching}
+   * @return up to {@code limit} {@link HRegionLocation}s in start-key order, possibly empty when no
+   *         more regions exist
+   * @throws IOException if a remote or network exception occurs
+   */
+  List<HRegionLocation> getRegionLocations(byte[] startKey, int limit) throws IOException;
```
haridsv: Since this is changing a public interface on a stable branch, this will break external code implementing the interface, why not have a default implementation perhaps throw unsupported or return a blank list?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionLocator.java
```diff
@@ -130,6 +131,44 @@ default List<HRegionLocation> getRegionLocations(byte[] row) throws IOException
...
+   * regions in start-key order.
+   * <p/>
+   * The returned list includes all replicas of each region (matching
+   * {@link #getAllRegionLocations()}), and the result is also written to the connection's region
+   * location cache.
+   * <p/>
+   * Ordering: regions are returned in ascending region start-key order (the natural order of
+   * {@code hbase:meta} rows for a single table). Within each region, replicas are returned in
+   * ascending replica-id order (replica 0, then 1, then 2, ...). Split parents and offline regions
+   * are filtered out, which may cause a page to contain fewer than {@code limit} regions but never
```
haridsv, line 146: Where is this filtering happening? I didn't see any test coverage either. The existing methods don't do any such filtering correct, so is this even needed?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionLocator.java
haridsv: Also, this is overloading an existing methods that takes a data row to find its locations, which is a totally different semantic and so reusing that name for this can be very confusing. Perhaps just call this `getLocations` or may be you can come up with a better name.

hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotRegionLocator.java
sanjeet006py: Yeah, I was also wondering same. Thanks for bringing this up. Instead of implementing paging here, shall I throw an exception saying use getAllRegionLocations as I don't see advantage of paging in here.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/RegionLocator.java
sanjeet006py: Filtering is already implemented and happens in inside `MetaTableAccessor.DefaultVisitorBase#visit()`. Call chain: `HRegionLocator#getRegionLocations()` -> `MetaTableAccessor.TableVisitorBase#visit()` -> `MetaTableAccessor.DefaultVisitorBase#visit()`.

sanjeet006py: I am reusing the filtering logic so, no test coverage needed.

hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotRegionLocator.java
haridsv: Sure, that works too, as long as you call out the fact that all implementation may not support it and that they should fallback to getAllRegionLocations.

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestRegionLocatorPagedScanRpcCount.java
```diff
@@ -0,0 +1,215 @@
...
+/**
+ * Asserts the single-RPC promise of the paginated meta-scan path
+ * ({@link MetaTableAccessor#scanMetaForTableRegions(Connection, MetaTableAccessor.Visitor, TableName, byte[], int, CatalogReplicaMode)})
+ * by wrapping the meta {@link Table} so we can read {@link ScanMetrics#countOfRPCcalls} from the
+ * issued {@link ResultScanner} - one per ScannerNext server response.
+ * <p/>
+ * Cluster runs with {@code hbase.meta.scanner.caching = 2} so the {@code limit > caching} branch is
+ * exercised cheaply with a small table (5 user regions).
+ */
+@Category({ MediumTests.class, ClientTests.class })
```
haridsv: Considering that the purpose is to verify that the scanner caching value is correct, I feel a mini cluster test is unnecessary and adds too much time on the test duration. Please check if you can achieve this using a mock.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/ConnectionImplementation.java
```diff
@@ -1185,6 +1181,19 @@ void takeUserRegionLock() throws IOException {
     }
   }
 
+  /**
+   * Release {@link #userRegionLock} previously acquired via {@link #takeUserRegionLock()} and
+   * record the held duration in metrics.
+   * @param lockStartTimeMs value of {@link EnvironmentEdgeManager#currentTime()} captured
+   *                        immediately after {@link #takeUserRegionLock()} returned
```
Umeshkumar9414, line 1188: wonder how spotless didn't take care of spaces.

sanjeet006py: I think this is valid space as we are listing method parameters and so, spotless also didn't flag it. On viewing in editor it is aligned rightly.
Screenshot 2026-05-15 at 9 10 20 PM

hbase-server/src/test/java/org/apache/hadoop/hbase/client/TestRegionLocatorPagedScanRpcCount.java
sanjeet006py: I tried to find a way to do this with a mock and got a bit stuck — wanted to share where I landed.

The property the test is asserting is "exactly one `ScannerNext` RPC fires for a paged scan, regardless of `hbase.meta.scanner.caching`." That isn't really a property of `MetaTableAccessor` in isolation — it's an emergent property of `ClientScanner` + `ScannerCallable` + the RegionServer responding to the scan's `caching`/`limit`, and `ScanMetrics.countOfRPCcalls` is incremented inside that machinery on real responses.

Two mock-based shapes I considered, both of which seem to weaken the test:

1. Capture the `Scan` and assert `scan.getCaching() == rowLimit`. This restates the one-line fix in `getMetaScan` and wouldn't catch a future regression where caching is set correctly but the client-side batching contract changes.
2. Hand-roll a fake `ResultScanner` that simulates `ScannerNext` batching from `caching`/`limit`. That re-implements the contract we're verifying inside the test fake and asserts our fake matches our expectation — feels tautological.

If you see a shape I missed that preserves the RPC-count assertion without a cluster, I'd be happy to switch. Otherwise I'd lean toward keeping this as an integration test.

haridsv (approved):
LGTM

virajjasani:
we have merge conflicts

virajjasani (approved):
+1, pending latest build

PR #8237, opened 2026-05-14
HBASE-30161 Add paginated, single-RPC RegionLocator.getRegionLocations(startKey, limit) API for bulk meta-cache warmup

sanjeet006py:
JIRA: [HBASE-30161](https://issues.apache.org/jira/browse/HBASE-30161)

Generated-by: Claude Opus 4.7


hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncTableRegionLocator.java
```diff
@@ -112,6 +113,54 @@ default CompletableFuture<List<HRegionLocation>> getRegionLocations(byte[] row)
...
+   * regions in start-key order.
+   * <p/>
+   * The returned list includes all replicas of each region (matching
+   * {@link #getAllRegionLocations()}), and the result is also written to the connection's region
+   * location cache.
+   * <p/>
+   * Ordering: regions are returned in ascending region start-key order (the natural order of
+   * {@code hbase:meta} rows for a single table). Within each region, replicas are returned in
+   * ascending replica-id order (replica 0, then 1, then 2, ...). Split parents are filtered out,
+   * which may cause a page to contain fewer than {@code limit} regions but never disturbs ordering
```
Umeshkumar9414, line 128: what about merged regions ?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncTableRegionLocatorImpl.java
```diff
@@ -81,6 +83,39 @@ public CompletableFuture<List<HRegionLocation>> getRegionLocations(byte[] row, b
...
+        ? limit
+        : conn.getConfiguration().getInt(HConstants.HBASE_META_SCANNER_CACHING,
+          HConstants.DEFAULT_HBASE_META_SCANNER_CACHING);
+      CompletableFuture<List<HRegionLocation>> future =
+        ClientMetaTableAccessor.getTableHRegionLocations(conn.getTable(TableName.META_TABLE_NAME),
+          tableName, startKey, effectiveLimit);
+      addListener(future, (locs, error) -> {
+        if (error != null || locs == null) {
+          return;
+        }
```
Umeshkumar9414, line 107: what if there is error ?

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncTableRegionLocator.java
sanjeet006py: Thanks for flagging — let me share my understanding and please correct me if I've got this wrong.

I think merge parents don't need filtering here because of how splits vs merges record cleanup state in `hbase:meta`:

- **Split:** the parent row stays around with `setOffline(true).setSplit(true)` (see `RegionStateStore#splitRegion`) and carries `info:splitA`/`info:splitB` pointers. The catalog janitor seems to use that row to track whether daughters still hold HFile references, and only deletes the parent row + HDFS dir once references are gone. So during that window, meta scans do see the parent — which is why `excludeOfflinedSplitParents` exists.

- **Merge:** from what I can see in `RegionStateStore#mergeRegions`, the same multi-mutate that commits the merge *deletes* every parent row and writes a single child row with `info:merge_0000`, `info:merge_0001`, … pointing back at the parents. So the "waiting on HFile cleanup" bookkeeping lives on the child row, not on the parents.

If that's right, a meta scan should never see merge parents, and the merged child row is itself the live region for that key range — which is what we'd want to return. This also seems to match `getAllRegionLocations()`'s existing behavior; both share `CollectRegionLocationsVisitor(excludeOfflinedSplitParents=true)`.

Does this line up with your understanding? Happy to add a comment near the visitor call site noting this if it'd help future readers, or revisit the filter if I'm missing a case.

hbase-client/src/main/java/org/apache/hadoop/hbase/client/AsyncTableRegionLocatorImpl.java
sanjeet006py: Good catch — let me explain the intent and please push back if it's not the right call.

The listener at `AsyncTableRegionLocatorImpl.java:104` is a side-effect-only hook for cache warm-up. The `future` returned to the caller is the one from `ClientMetaTableAccessor.getTableHRegionLocations`, which already completes exceptionally on any meta-scan error (see `getTableRegionsAndLocations`'s `future.completeExceptionally(error)` paths). So the error propagates to the caller through the returned future regardless of what the listener does.

On the listener side, if `error != null` there's simply nothing valid to cache, and if `locs == null` there's nothing to iterate. Either way, returning early is the safe behavior — we don't want to mask the error or double-fail the future, just skip cache population for this call. The next call to `getRegionLocationsPage` (or any locator path) will populate normally.

That said, I can add a `LOG.debug` (or `LOG.warn`) in the error branch so the silent skip is observable, e.g.:
```java
if (error != null) {
  LOG.debug("Skipping cache warm-up for {}: {}", tableName, error.toString());
  return;
}
if (locs == null) {
  return;
}
```
Happy to add that if you think it's worth it.

virajjasani (approved):
+1

PR #8264, opened 2026-05-21
HBASE-30169 Split completed parent region added to RIT list on host RS and master crash

Umeshkumar9414:
Jira: HBASE-30169

- Bug: When a host RegionServer and master crashes, regionCrashed() blindly added all hosted regions to RIT, including split-completed parent regions that should never be reassigned.
  - Fix: Added isSplitOrMerged() guard in regionCrashed() to skip split/merged regions. Extracted isSplitOrMerged() and isReplica() helpers to reduce duplication with handleRegionStateNodeOperation().
  - Removed broken isRegionInTransition(RegionInfo) API — RegionInfo.COMPARATOR considers offline/split flags, so lookups failed when those flags differed between the query and stored instance. Replaced
  with encoded-name-based lookup in test utility.
  - Added test testRITWithSplitTableRegion — splits a region, crashes the hosting RS, restarts the master, and verifies the split parent is NOT in the RIT list.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
@@ -35,6 +35,7 @@ import org.apache.hadoop.hbase.client.AsyncRegionServerAdmin;
 import org.apache.hadoop.hbase.client.RegionInfo;
 import org.apache.hadoop.hbase.client.RegionReplicaUtil;
 import org.apache.hadoop.hbase.favored.FavoredNodesManager;
+import org.apache.hadoop.hbase.master.RegionState;
 import org.apache.hadoop.hbase.master.procedure.MasterProcedureEnv;
 import org.apache.hadoop.hbase.util.FutureUtils;
 import org.apache.hadoop.hbase.wal.WALSplitUtil;
@@ -297,5 +298,17 @@ final class AssignmentManagerUtil {
       throw new IOException("Recovered.edits are found in Region: " + regionInfo
         + ", abort split/merge to prevent data loss");
     }
+  }
+
+  /**
+   * For splitting, need to test both region info and state, and will return true if either of the
+   * test returns true. Please see the comments in
+   * {@link AssignmentManager#markRegionAsSplit(RegionInfo, ServerName, RegionInfo, RegionInfo)} for
+   * more details on why we need to test two conditions.
+   */
+  static boolean isSplitOrMerged(RegionStateNode regionStateNode) {
+    return regionStateNode.getState() == RegionState.State.SPLIT
+      || regionStateNode.getRegionInfo().isSplit()
+      || regionStateNode.getState() == RegionState.State.MERGED;
   }
 }
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
@@ -41,8 +41,10 @@ public class RegionInTransitionTracker {
 
   private final List<RegionState.State> ENABLE_TABLE_REGION_STATE = List.of(RegionState.State.OPEN);
 
-  // DO NOT USE containsKey() on regionInTransition map as MutableRegionInfo.COMPARATOR considers
-  // offline/split flags.
+  // DO NOT USE containsKey()/remove() on regionInTransition with a different RegionInfo instance:
+  // this map is ordered by RegionInfo.COMPARATOR, and that comparator includes the offline flag.
+  // Lookups can therefore fail if the RegionInfo used as the key has a different offline value,
+  // even when it refers to the same region. Offline value changes with splitting.
   private final ConcurrentSkipListMap<RegionInfo, RegionStateNode> regionInTransition =
     new ConcurrentSkipListMap<>(RegionInfo.COMPARATOR);
 
@@ -54,7 +56,7 @@ public class RegionInTransitionTracker {
    * other servers.
    */
   public void regionCrashed(RegionStateNode regionStateNode) {
-    if (isReplica(regionStateNode) || isSplitOrMerged(regionStateNode)) {
+    if (isReplica(regionStateNode)) {
       return;
     }
 
@@ -84,7 +86,7 @@ public class RegionInTransitionTracker {
       tableEnabled ? ENABLE_TABLE_REGION_STATE : DISABLE_TABLE_REGION_STATE;
 
     // if region is merged or split it should not be in RIT list
-    if (isSplitOrMerged(regionStateNode)) {
+    if (AssignmentManagerUtil.isSplitOrMerged(regionStateNode)) {
       if (removeRegionInTransition(regionStateNode.getRegionInfo())) {
         LOG.debug("Removed {} from RIT list as it is split or merged",
           regionStateNode.getRegionInfo().getEncodedName());
@@ -100,12 +102,6 @@ public class RegionInTransitionTracker {
           regionStateNode.getRegionInfo().getEncodedName(), currentState);
       }
     }
-  }
-
-  private static boolean isSplitOrMerged(RegionStateNode regionStateNode) {
-    return regionStateNode.getState() == RegionState.State.SPLIT
-      || regionStateNode.getState() == RegionState.State.MERGED
-      || regionStateNode.getRegionInfo().isSplit();
   }
 
   private boolean isTableEnabled(TableName tableName) {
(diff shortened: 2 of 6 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/RegionInTransitionTracker.java
```diff
@@ -41,22 +41,20 @@ public class RegionInTransitionTracker {
 
   private final List<RegionState.State> ENABLE_TABLE_REGION_STATE = List.of(RegionState.State.OPEN);
 
+  // DO NOT USE containsKey() on regionInTransition map as MutableRegionInfo.COMPARATOR considers
+  // offline/split flags.
```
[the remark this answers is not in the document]
apurtell: Let's do this one.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/AssignmentTestingUtil.java
```diff
@@ -49,12 +49,17 @@ private AssignmentTestingUtil() {
   }
 
   public static void waitForRegionToBeInTransition(final HBaseTestingUtil util,
-    final RegionInfo hri) throws Exception {
-    while (!getMaster(util).getAssignmentManager().isRegionInTransition(hri)) {
+    final RegionInfo hri) {
+    while (!isRegionInTransition(hri, getMaster(util).getAssignmentManager())) {
       Threads.sleep(10);
```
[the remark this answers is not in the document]
apurtell: This is a good suggestion, if unrelated to the change, but might cause issues with other tests once you start pulling on the thread. I would ignore this one.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRegionSplit.java
```diff
@@ -155,6 +155,54 @@ public void testSplitTableRegion() throws Exception {
...
+    // As there are only 3 RS, start one more RS before expiring one
+    UTIL.getHBaseCluster().startRegionServer();
+
+    // stop RS holding split parent
+    UTIL.getHBaseCluster().getMaster().getServerManager().expireServer(targetRS);
+
+    // stop master
+    UTIL.getHBaseCluster().stopMaster(0);
+    UTIL.getHBaseCluster().waitOnMaster(0);
+    Thread.sleep(500);
```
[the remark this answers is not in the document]
apurtell: Also one to look at.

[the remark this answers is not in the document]
Umeshkumar9414: In this I just want to submit the SCP to master. So it include the RS into dead server list. I don't need the SCP to complete.

Umeshkumar9414:
Failing test is related to compaction. Unrelated to PR. Can look seperatly later.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
```diff
@@ -298,4 +299,16 @@ static void checkClosedRegion(MasterProcedureEnv env, RegionInfo regionInfo) thr
...
     }
   }
+
+  /**
+   * For splitting, need to test both region info and state, and will return true if either of the
+   * test returns true. Please see the comments in
+   * {@link AssignmentManager#markRegionAsSplit(RegionInfo, ServerName, RegionInfo, RegionInfo)} for
+   * more details on why we need to test two conditions.
+   */
+  static boolean isSplitOrMerged(RegionStateNode regionStateNode) {
```
apurtell, line 309: An AI assisted review finds:

This check `isSplitOrMerged` has been moved out of `regionCrashed` , where it formerly guarded calls to both `visitRegionState` and `markRegionsAsCrashed`.  After this refactor, now only `visitRegionState` is guarded. A split parent in `SPLIT` state whose location is still equal to the crashed server (the parent stays in `regionStates` in this state until the catalog janitor cleans it) will now be processed in `regionCrashed` when this didn't used to be possible.

Recommendation: 

Add `& !AssignmentManagerUtil.isSplitOrMerged(node)` to the `if (crashedServerName.equals(node.getRegionLocation()))` branch in `markRegionsAsCrashed`, just like the same check guard `visitRegionState`.

Umeshkumar9414: This was intentional @apurtell . As soon as region got splitted or merged we remove that from the RegionStates. And even if HMaster is restarted while [loading the regions in ](https://github.com/apache/hbase/pull/8264/changes/BASE..a6fa2651419f2422b771268bd734f96a8b4b6524#diff-8701f497f12ec66e8fe4cfcc91b1876d95793172eecf5a269f40f126283ac145L1886)visitRegionState, we don't add those regions back to visitRegionState RegionStates. So I think we are fine not adding the extra check. 
Let me know if I am missing something.

apurtell (approved):

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
apurtell: Yes I read through the code and I think visitRegionState is enough.

Umeshkumar9414:
@virajjasani , @Apache9 is there any concern on this PR or can we merge this ?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRegionSplit.java
```diff
@@ -155,6 +156,57 @@ public void testSplitTableRegion() throws Exception {
...
+    // Split region of the table
+    long procId = procExec.submitProcedure(
+      new SplitTableRegionProcedure(procExec.getEnvironment(), regions[0], splitKey));
+    // Wait the completion
+    ProcedureTestingUtility.waitProcedure(procExec, procId);
+    ProcedureTestingUtility.assertProcNotFailed(procExec, procId);
+
+    assertEquals(2, UTIL.getHBaseCluster().getRegions(tableName).size(), "not able to split table");
+    assertFalse(AssignmentTestingUtil.isRegionInTransition(regions[0],
+      UTIL.getHBaseCluster().getMaster().getAssignmentManager()));
```
virajjasani, line 188: After this, can we also check for region[0] to be split?
e.g.
```
    assertTrue(UTIL.getHBaseCluster().getMaster()
      .getAssignmentManager().getRegionStates().getRegionState(regions[0]).isSplit());
```

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRegionSplit.java
```diff
@@ -155,6 +156,57 @@ public void testSplitTableRegion() throws Exception {
...
+    UTIL.getHBaseCluster().waitOnMaster(0);
+
+    // restart master
+    UTIL.getHBaseCluster().startMaster();
+    assertTrue(UTIL.getHBaseCluster().waitForActiveAndReadyMaster(30000),
+      "Master failed to initialize in in 30 seconds");
+    UTIL.invalidateConnection();
+
+    assertFalse(AssignmentTestingUtil.isRegionInTransition(regions[0],
+      UTIL.getHBaseCluster().getMaster().getAssignmentManager()));
```
virajjasani, line 211: same here after this? check for region[0] being split and offline?

virajjasani (approved):
Left minor nits, +1 otherwise

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRegionSplit.java
Umeshkumar9414: instead of this I am using getOrCreateRegionStateNode(regions[0]).isSplit() as it do both the necessary check for split

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRegionSplit.java
Umeshkumar9414: added

Umeshkumar9414:
PR for branch-2 : https://github.com/apache/hbase/pull/8350
for branch-2.6 : https://github.com/apache/hbase/pull/8351
for branch-2.5 : https://github.com/apache/hbase/pull/8352

virajjasani:
> PR for branch-2 : #8350 for branch-2.6 : #8351 for branch-2.5 : #8352

Thanks @Umeshkumar9414, all backport PRs have `TestRegionMover2` failures

Umeshkumar9414:
> PR for branch-2 : #8350 for branch-2.6 : #8351 for branch-2.5 : #8352

Will look into it.

Umeshkumar9414:
There was a issue with this and Duo did pushed a fix for this - https://issues.apache.org/jira/browse/HBASE-30223

Umeshkumar9414:
> There was a issue with this and Duo did pushed a fix for this - https://issues.apache.org/jira/browse/HBASE-30223

Updated all branches. Please check now.

virajjasani:
Is `TestMajorCompaction` failing in your local? Is it failing without the PR?

virajjasani:
It seems to be present in other master branch PRs too, not relevant here, merging the PR soon

PR #8269, opened 2026-05-23
HBASE-30173 Add AGENTS.md for LLM-assisted codebase navigation

ndimiduk:
LLM-based tools (code assistants, security scanners, review agents) increasingly operate against large codebases. This adds an AGENTS.md convention file that teaches these tools how to navigate HBase's 37+ Maven modules, understand the client/server divide, use existing @InterfaceAudience annotations for orientation, and locate key entry points. The security model section points to the existing documentation rather than duplicating it.

Also adds a dev-support/README.md index and fixes stale content across 12 existing READMEs — incorrect paths, removed modules, outdated versions, broken URLs, and a complete rewrite of hbase-endpoint and hbase-rest READMEs that described code no longer present. Three Java comment references to the renamed hbase-hadoop2-compat module were corrected as well.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,187 @@
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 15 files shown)


liuxiaocs7 (approved):

Apache9 (approved):
Thanks Nick, let's use this AGENTS.md as a start point.

Do we need this on all branches? Or only on master branch and include the branch strateg in AGENTS.md?

ndimiduk:
> Do we need this on all branches? Or only on master branch and include the branch strateg in AGENTS.md?

I was thinking we'd start with just master and see how it goes. Maybe we can drop a file on each active branch that says "refer to AGENTS.md on master branch" and let agents track back to the source. Or maybe there's enough deviation between branches that we want to customize sections for each branch.

ndimiduk:
Merged just to master. Thanks for the reviews!

PR #8276, opened 2026-05-27
HBASE-30177 Make block caching configurable for MOB compaction reads

liuxiaocs7:
- see: HBASE-30177

### DIFF
--- a/hbase-common/src/main/resources/hbase-default.xml
+++ b/hbase-common/src/main/resources/hbase-default.xml
@@ -1927,12 +1927,12 @@ possible configurations would overwhelm and obscure the important.
     </description>
   </property>
   <property>
-    <name>hbase.mob.compaction.cache.blocks</name>
+    <name>hbase.mob.compaction.read.cache.blocks</name>
     <value>true</value>
     <description>
-      Whether MOB compaction should cache blocks when resolving existing MOB file references.
-      The default value is true to preserve the previous behavior. Set to false to avoid polluting
-      the block cache with background compaction reads.
+      Whether MOB compaction should cache blocks while reading existing MOB files to resolve
+      references during compaction. The default value is true to preserve the previous behavior.
+      Set to false to avoid polluting the block cache with background compaction reads.
     </description>
   </property>
   <property>
--- a/hbase-common/src/main/resources/hbase-default.xml
+++ b/hbase-common/src/main/resources/hbase-default.xml
@@ -1927,6 +1927,15 @@ possible configurations would overwhelm and obscure the important.
     </description>
   </property>
   <property>
+    <name>hbase.mob.compaction.cache.blocks</name>
+    <value>true</value>
+    <description>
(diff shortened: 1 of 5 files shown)


hbase-common/src/main/resources/hbase-default.xml
```diff
@@ -1926,6 +1926,15 @@ possible configurations would overwhelm and obscure the important.
       The default value is one week.
     </description>
   </property>
+  <property>
+    <name>hbase.mob.compaction.cache.blocks</name>
```
VladRodionov: Minor naming/documentation question: the config name hbase.mob.compaction.cache.blocks is understandable, but the behavior is specifically about caching blocks while reading existing MOB files during compaction, not cache-on-write of compaction output. To avoid confusion with cache-on-write settings, would a more explicit name such as hbase.mob.compaction.read.cache.blocks or hbase.mob.compaction.cache.blocks.on.read be clearer?

VladRodionov (commented):
Thanks for the patch. The change looks good to me. The new config is useful, and keeping the default as true preserves the existing behavior.

Minor: the current test verifies that the compactor reads the configuration value, which is probably enough for this change since the production code simply passes that field into mobStore.resolve(...). No blocker from my side.

VladRodionov (approved):
lgtm.

hbase-common/src/main/resources/hbase-default.xml
liuxiaocs7: Thanks for the comments, agree with your points. Renamed it to `hbase.mob.compaction.read.cache.blocks` and updated the description to make it explicit that this only applies to reads of existing MOB files during compaction.

hbase-server/src/main/java/org/apache/hadoop/hbase/mob/DefaultMobStoreCompactor.java
```diff
@@ -377,7 +379,7 @@ protected boolean performCompaction(FileDetails fd, InternalScanner scanner, Cel
               String fName = MobUtils.getMobFileName(c);
               // Added to support migration
               try {
-                mobCell = mobStore.resolve(c, true, false).getCell();
+                mobCell = mobStore.resolve(c, cacheMobBlocksOnCompaction, false).getCell();
```
guluo2016, line 382: LGTM for this changes，Thanks!

With the help of IDEA inspections, I noted an additional issue: `mobStore.resolve()` would return `MobCell` object, and `MobCell implements Closeable`, so, should we explicitly close it here?

liuxiaocs7: Hi, @guluo2016, thanks for the comments, agree with your points, this looks pre-existing, not introduced by HBASE-30177. I’ll open a separate JIRA and put the fix in a separate commit, WDYT.

Thanks!

guluo2016 (approved):

liuxiaocs7:
Failed UTs seems unrelated, let me merge

PR #8278, opened 2026-05-27
HBASE-30175 Upgrade hbase-server to use junit5 Part19

Apache9:


hbase-server/src/test/java/org/apache/hadoop/hbase/replication/regionserver/TestReplicationSourceManagerJoin.java
```diff
@@ -31,40 +30,36 @@
...
-    HBaseClassTestRule.forClass(TestReplicationSourceManagerJoin.class);
-
-  @Rule
-  public TestName testName = new TestName();
+@Tag(ReplicationTests.TAG)
+@Tag(MediumTests.TAG)
+public class TestReplicationSourceManagerJoin extends TestReplicationBaseNoBeforeAll {
 
+  @BeforeAll
   @BeforeClass
```
liuxiaocs7: We still need `BeforeClass` here?

liuxiaocs7 (approved):

PR #8279, opened 2026-05-27
HBASE-30022 Refactor CacheConfig and HFileReaderImpl to use CacheAcce…

VladRodionov:
## Summary

This PR refactors HFile reader cache access to go through `CacheAccessService`.

`CacheConfig` now exposes a `CacheAccessService` backed by the existing `BlockCache`, or by a disabled no-op service when block cache is not configured. `HFileReaderImpl` and `HFilePreadReader` use this service for block cache lookup and cache population instead of accessing `BlockCache` directly.

## Changes

- Add `CacheAccessService` wiring to `CacheConfig`
- Keep existing `CacheConfig#getBlockCache()` unchanged for compatibility
- Refactor `HFileReaderImpl` cache lookup/population to use `CacheAccessService`
- Refactor `HFilePreadReader` cache access to use `CacheAccessService`
- Preserve existing cache flags and behavior

## Notes

This PR does not change cache construction, `BlockCacheFactory`, topology-backed runtime selection, or writer-path cache population.

The intent is to keep behavior unchanged while moving the HFile read path from direct `BlockCache` access to the new service abstraction introduced in HBASE-30021.

## Testing

Ran:

```bash
mvn -pl hbase-server spotless:check
mvn -pl hbase-server checkstyle:check
mvn -pl hbase-server -DskipTests compile
mvn -pl hbase-server test
```

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
@@ -47,6 +47,7 @@ import org.apache.hadoop.hbase.io.compress.Compression;
 import org.apache.hadoop.hbase.io.encoding.DataBlockEncoder;
 import org.apache.hadoop.hbase.io.encoding.DataBlockEncoding;
 import org.apache.hadoop.hbase.io.encoding.HFileBlockDecodingContext;
+import org.apache.hadoop.hbase.io.hfile.cache.CacheAccessService;
 import org.apache.hadoop.hbase.monitoring.ThreadLocalServerSideScanMetrics;
 import org.apache.hadoop.hbase.nio.ByteBuff;
 import org.apache.hadoop.hbase.regionserver.KeyValueScanner;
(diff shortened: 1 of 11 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
```diff
@@ -1104,14 +1105,14 @@ public HFileBlock getCachedBlock(BlockCacheKey cacheKey, boolean cacheBlock, boo
     boolean updateCacheMetrics, BlockType expectedBlockType,
     DataBlockEncoding expectedDataBlockEncoding) throws IOException {
     // Check cache for block. If found return.
-    BlockCache cache = cacheConf.getBlockCache().orElse(null);
+    CacheAccessService cacheAccessService = cacheConf.getCacheAccessService();
     long cachedBlockBytesRead = 0;
-    if (cache != null) {
+    if (cacheAccessService != null) {
       HFileBlock cachedBlock = null;
```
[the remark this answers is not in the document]
taklwu: nit: can you check what we should do here? I agreed that it's non null and default to `NoOpCacheAccessService` and null check does not make any difference.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopology.java
```diff
@@ -198,4 +198,50 @@ default boolean demote(BlockCacheKey cacheKey, Cacheable block, CacheEngine sour
...
+   * <p>
+   * Resizing is an optional per-engine capability. A topology may contain a mix of resizable and
+   * non-resizable engines, for example a resizable on-heap L1 and a fixed-size off-heap L2. This
+   * helper allows admin/control code to discover resize support without exposing resizing on the
+   * read/write path facade.
+   * </p>
+   * @param tier cache tier to inspect
+   * @return resizable engine for the tier, or empty if the tier does not exist or is not resizable
+   */
+  default Optional<ResizableCacheEngine> getResizableEngine(CacheTier tier) {
```
taklwu: what is this about ? I cannot find usage in this PR

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopology.java
```diff
@@ -198,4 +198,50 @@ default boolean demote(BlockCacheKey cacheKey, Cacheable block, CacheEngine sour
...
+   * <p>
+   * This helper is intended for diagnostics, metrics, and admin/control code. Normal HFile read and
+   * write path callers should use {@link CacheAccessService} and should not depend on this optional
+   * capability.
+   * </p>
+   * @param tier cache tier to inspect
+   * @return file-tracking engine for the tier, or empty if the tier does not exist or does not
+   *         support file/region cache tracking
+   */
+  default Optional<CacheFileTrackingEngine> getFileTrackingEngine(CacheTier tier) {
```
taklwu: ditto, if we don't have any usage at the moment, can you remove it?

taklwu (changes requested):
I found other `BlockCache blockCache` e.g. in HRegion and assumed they will be addressed in followup JIRA/PR.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopology.java
VladRodionov: ResizableBlockCache interface conversion. Yes, it will be used later in subsequent PRs.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopology.java
VladRodionov: The same as above. Ground work for next PR.

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestCacheConfig.java
```diff
@@ -417,4 +423,28 @@ public void testGetOnHeapCacheSize() {
     onHeapCacheSize = MemorySizeUtil.getOnHeapCacheSize(copyConf);
     assertEquals(fixedSize, onHeapCacheSize);
   }
+
+  @Test
+  void testCacheAccessServiceBackedByBlockCacheWhenBlockCacheIsConfigured() {
+    Configuration conf = new Configuration(false);
+    BlockCache blockCache = BlockCacheFactory.createBlockCache(this.conf);
+    CacheConfig cacheConfig = new CacheConfig(conf, blockCache);
+
```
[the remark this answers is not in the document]
wchevreuil: Should apply this?

wchevreuil (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestCacheConfig.java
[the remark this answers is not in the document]
VladRodionov: Yes, fixed

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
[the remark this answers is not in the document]
VladRodionov: Done.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheTopology.java
VladRodionov: OK, I will remove this code.

VladRodionov:
> I found other `BlockCache blockCache` e.g. in HRegion and assumed they will be addressed in followup JIRA/PR.

Yes. Other call sites will be addressed  in a followup PRs.

Kota-SH (approved):
LGTM

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
```diff
@@ -1104,120 +1105,120 @@ public HFileBlock getCachedBlock(BlockCacheKey cacheKey, boolean cacheBlock, boo
...
-              compressedBlock.release();
-            }
-          }
-          try {
-            validateBlockType(cachedBlock, expectedBlockType);
-          } catch (IOException e) {
-            returnAndEvictBlock(cache, cacheKey, cachedBlock);
-            cachedBlock = null;
-            throw e;
+    HFileBlock cachedBlock = null;
```
taklwu, line 1110: nit: I'm wondered if we should do this instead? basically, if we let the code follow throw the `non-null`/ non-`NoOpCacheAccessService` , either it's still using any code logic to do no ops, or if `NoOpCacheAccessService` has any bug or human error in the future, it may be easier to cause another issue.

what do you think?

```suggestion
    HFileBlock cachedBlock = null;
    if (cacheAccessService instanceof NoOpCacheAccessService) {
      return null;
    }
```

VladRodionov: Thanks. I think I would avoid the explicit `instanceof NoOpCacheAccessService` check here.
`CacheAccessService` is intended to be the abstraction boundary, so `HFileReaderImpl` should not need
to know which concrete service implementation it received. The disabled-cache behavior belongs in
`NoOpCacheAccessService`; its `getBlock(...)` returns null, so the current flow already exits
without doing cache work.
If there is concern about future bugs in `NoOpCacheAccessService`, I think the better protection is
unit coverage for that implementation rather than adding implementation-specific checks in the read
path. This keeps the reader code independent of whether the service is block-cache-backed,
topology-backed, or no-op.

taklwu (approved):
LGTM

PR #8280, opened 2026-05-28
HBASE-30180: Can still add data to read-only region after flipping read-only flag multiple times

kgeisz:
https://issues.apache.org/jira/browse/HBASE-30180

**Note:  Unit tests were generated using AI**
- `testRegionOnConfigurationChangeLoadsReadOnlyCoprocessorsAfterRepeatedToggle()` was generated with Cursor
- `testRegionCoprocessorHostUsesCompoundConfigurationAfterConfigChange()` was generated with Claude Opus

## Summary

This pull request fixes an issue with the new Read-Replica feature where if you create a table on the active cluster and flip the read-only flag multiple times, then sometimes it is still possible to add data to a table despite the cluster being in read-only mode.  This issue was happening due to how the configuration for `HRegion` was being updated after running `update_all_config` in the HBase shell.

### Key Changes
- Delete the `CoprocessorConfigurationUtil.updateCoprocessorListInConf()` method since running this in `HMaster`, `HRegionServer`, and `HRegion`'s `onConfigurationChange()` method is redundant.
    - This method was also improperly unsetting the region coprocessors for HRegion's `CompoundConfiguration`, which was the main cause of HBASE-30180.
- Properly create a new `RegionCoprocessorHost` object in `HRegion.onConfigurationChange()` after a dynamic configuration update.
    - The new object was being created with the dynamically updated `Configuration` object provided as an arg in `HRegion.onConfigurationChange()`.  However, it needs to be created with an updated version of HRegion's `this.conf` `CompoundConfiguration` object.
- Add documentation to HMaster, HRegionServer, and HRegion's `onConfigurationChange()` methods regarding how the `Configuration` objects work.
    - For HMaster and HRegionServer, the `Configuration` object provided in their `onConfigurationChange()` methods is the same object reference as their respective `this.conf` `Configuration` objects.
- Rename some variables for more clarity.

## More Details

Read-only mode is determined by checking whether the ReadOnly coprocessors are loaded on the cluster.  If they are present, then read-only mode is active, and vice-vera if they are not present.  HBASE-30180 was being caused by how the `CoprocessorConfigurationUtil.updateCoprocessorListInConf()` method was unsetting the read-only coprocessors.  HRegion's `this.conf` object is of type `CompoundConfiguration`.  This differs from HMaster and HRegionServer, which use type `Configuration`.  When disabling read-only mode, the `updateCoprocessorListInConf()` was running `unset()` on the configuration object, which does not properly unset the config property for `CompoundConfiguration`s.  Instead, the property needs to explicitly be set to an empty string.

After further investigation, the `updateCoprocessorListInConf()` method was redundant and not necessary since the read-only coprocessors were already being added/removed in `CoprocessorConfigurationUtil.syncReadOnlyConfigurations()`.  That is why this PR removes the `updateCoprocessorListInConf()` entirely.

One additional issue was with how HRegion's `this.coprocessorHost` object was being updated after a dynamic configuration update.  The new `RegionCoprocessorHost` object created for `this.coprocessorHost` was using the dynamically updated `Configuration` object when it should have been using an updated version of HRegion's `CompoundConfiguration` object.  This PR corrects that discrepancy.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
@@ -17,10 +17,13 @@
  */
 package org.apache.hadoop.hbase.util;
 
+import static org.apache.hadoop.hbase.HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY;
+
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashSet;
 import java.util.List;
+import java.util.Objects;
 import java.util.Set;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HConstants;
@@ -178,6 +181,25 @@ public final class CoprocessorConfigurationUtil {
   }
 
   /**
+   * Updates the coprocessors in one Configuration object to match the coprocessors in the other
+   * @param srcConf            the Configuration object we are getting coprocessors from
+   * @param dstConf            the Configuration object we are updating
+   * @param coprocessorConfKey the type of coprocessors we are updating
+   */
+  private static void syncCoprocessorsWithConf(Configuration srcConf, Configuration dstConf,
+    String coprocessorConfKey) {
+    String configuredCps = srcConf.get(coprocessorConfKey);
+    dstConf.set(coprocessorConfKey, Objects.requireNonNullElse(configuredCps, ""));
+
+    // A configuration with region coprocessors may also have user region coprocessors
+    if (CoprocessorHost.REGION_COPROCESSOR_CONF_KEY.equals(coprocessorConfKey)) {
+      String configuredUserCps = srcConf.get(CoprocessorHost.USER_REGION_COPROCESSOR_CONF_KEY);
+      dstConf.set(CoprocessorHost.USER_REGION_COPROCESSOR_CONF_KEY,
+        Objects.requireNonNullElse(configuredUserCps, ""));
+    }
+  }
+
+  /**
    * This method adds or removes relevant ReadOnlyController coprocessors to the provided
    * configuration based on whether read-only mode is enabled in the provided Configuration.
    * @param conf               The up-to-date configuration used to determine how to handle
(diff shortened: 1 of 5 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/util/CoprocessorConfigurationUtil.java
```diff
@@ -264,28 +246,43 @@ public static String getComponentName(String coprocessorConfKey) {
...
       .checkConfigurationChange(coprocessorHost, newConf, coprocessorConfKey);
 
-    // update region server coprocessor if the configuration has changed.
     if ((hasCoprocessorConfigChanged || hasReadOnlyModeChanged) && !isMaintenanceMode) {
       LOG.info("Updating coprocessors for {} {} because the configuration has changed",
-        getComponentName(coprocessorConfKey), instance);
-      CoprocessorConfigurationUtil.syncReadOnlyConfigurations(newConf, coprocessorConfKey);
-      reloadTask.reload(newConf);
+        componentName, instance);
+      if (hasReadOnlyModeChanged) {
```
taklwu: this reminded me about #7514 were facing the same issue, IMO you're doing the right change.

taklwu (changes requested):
nit: Would you be able to come up with a unit test to prevent this error from happening?

wchevreuil (approved):

kgeisz:
The unit test failures are not relevant.

kgeisz:
I force-pushed another commit to fix a poorly formatted comment.  We can run the tests again if needed.  However, all relevant tests passed on the last run.

I also updated the PR description.

kgeisz:
Two more additions:

- Gave `HRegion.getConfiguration()` the `@RestrictedApi` annotation
- Added unit test:  `testRegionCoprocessorHostUsesCompoundConfigurationAfterConfigChange()`
    - This test ensures HRegion's `this.coprocessorHost` variable is updated with a new `RegionCoprocessorHost` object that uses a `CompoundConfiguration` object in its constructor (rather than a `Configuration` object).

taklwu (approved):
LGTM

PR #8281, opened 2026-05-28
HBASE-30138 Upgrade hbase-server to use junit5 Part16

liuxiaocs7:
- see: HBASE-30138


liuxiaocs7:
`TestClientScannerTimeouts` fails on JDK1.8, seems not related, let me merge

PR #8283, opened 2026-05-28
HBASE-30136 Upgrade hbase-server to use junit5 Part15

liuxiaocs7:
- for branch-2
- see: HBASE-30136


liuxiaocs7:
Same question, `TestClientScannerTimeouts` fails on JDK8, and it is unrelated, let me merge first!

PR #8284, opened 2026-05-29
HBASE-30183 Refactoring scanner related code in hbase-thrift and hbase-rest

Apache9:


hbase-thrift/src/test/java/org/apache/hadoop/hbase/thrift/TestThriftSpnegoHttpServer.java
```diff
@@ -194,11 +222,16 @@ protected void talkToThriftServer(String url, int customHeaderSize) throws Excep
       }
       TestThriftServer.createTestTables(client);
       TestThriftServer.checkTableList(client);
+
+      TProtocol prop2 = new TBinaryProtocol(tHttpClient2);
+      Hbase.Client client2 = new Hbase.Client(prot, prop2);
```
liuxiaocs7: Should this be `new Hbase.Client(prop2)` instead?

hbase-thrift/src/test/java/org/apache/hadoop/hbase/thrift/TestThriftSpnegoHttpServer.java
```diff
@@ -167,11 +174,32 @@ public void testRunThriftServerWithHeaderBufferLength() throws Exception {
...
+    mutations
+      .add(new Mutation(false, TestThriftServer.columnAAname, TestThriftServer.valueAname, true));
+    client.mutateRow(TestThriftServer.tableAname, TestThriftServer.rowAname, mutations,
+      Collections.emptyMap());
+
+    int id = client.scannerOpen(TestThriftServer.tableAname, ByteBuffer.allocate(0),
+      Collections.emptyList(), Collections.emptyMap());
+
+    assertThrows(Exception.class, () -> client2.scannerGet(id));
+    assertThrows(Exception.class, () -> client2.scannerClose(id));
```
liuxiaocs7: Could we narrow this assertion?

liuxiaocs7 (approved):

PR #8286, opened 2026-05-30
HBASE-30188 Upgrade hbase-server to use junit5 Part20

Apache9:


hbase-server/src/test/java/org/apache/hadoop/hbase/replication/ReplicationEndpointTestBase.java
```diff
@@ -0,0 +1,644 @@
...
+
+  /**
+   * Not used by unit tests, helpful for manual testing with replication.
+   * <p>
+   * Snippet for `hbase shell`:
+   *
+   * <pre>
+   * create 't', 'f'
+   * add_peer '1', ENDPOINT_CLASSNAME =&gt; 'org.apache.hadoop.hbase.replication.' + \
+   *    'TestReplicationEndpoint$SleepingReplicationEndpointForTest'
```
liuxiaocs7: We need to update the doc here to `ReplicationEndpointTestBase$SleepingReplicationEndpointForTest` after refactoring?

liuxiaocs7 (approved):

PR #8287, opened 2026-05-30
HBASE-30205 Optimize Bloom Filter bit operations by replacing array lookup with bitwise shift

xuliang0317:
Optimize Bloom Filter bit operations by replacing array lookup with bitwise shift

While this is functional, it introduces unnecessary overhead in the read path, which is highly performance-sensitive:

Array Bounds Checking: The JVM must perform bounds checking for every array access to prevent ArrayIndexOutOfBoundsException, which adds minor but cumulative overhead.
Memory Access: Even though bitvals is likely cached in L1/L2 CPU cache, it still requires a memory load instruction.
Division/Modulo Overhead: In BloomFilterChunk#set(long pos), the code uses / 8 and % 8 which compiles to relatively expensive division instructions.
Solution
This PR optimizes the Bloom Filter bit operations by replacing the array lookup with direct bitwise shifts and optimizing the division/modulo operations:

Direct Bitwise Shift: Replaced BloomFilterUtil.bitvals[bitPos] with (1 << bitPos).
This eliminates array bounds checks completely.
The JIT compiler can optimize 1 << bitPos into a single, extremely fast CPU instruction (SHL), executing entirely within CPU registers with zero memory access latency.
Remove Unused Array: Removed the BloomFilterUtil.bitvals array completely to clean up the code.
Division to Shift Optimization: In BloomFilterChunk#set(long pos), replaced pos / 8 and pos % 8 with pos >> 3 and pos & 7. Since pos is always non-negative in this context, this is a safe and much faster alternative to division instructions.
Proposed Changes

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/util/BloomFilterChunk.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/util/BloomFilterChunk.java
@@ -203,12 +203,12 @@ public class BloomFilterChunk implements BloomFilterBase {
    * @param pos index of bit
    */
   void set(long pos) {
-    int bytePos = (int) (pos / 8);
-    int bitPos = (int) (pos % 8);
+    int bytePos = (int) (pos >> 3);
+    int bitPos = (int) (pos & 7);
     byte curByte = bloom.get(bytePos);
-    curByte |= BloomFilterUtil.bitvals[bitPos];
+    curByte |= (1 << bitPos);  
     bloom.put(bytePos, curByte);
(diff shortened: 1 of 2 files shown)


xuliang0317:
@Apache9 Could you review this PR when you get a chance?

xuliang0317:
@apurtell Could you please take a look at this PR when you have a moment? I'd appreciate any feedback or suggestions you might have. Thanks!

VladRodionov:
@xuliang0317 Do you observe any measurable performance improvements?

PR #8299, opened 2026-06-01
HBASE-30195 Update Thrift gateway to use new hbase-shaded-thrift in thirdparty

apurtell:
On https://github.com/apache/hbase-thirdparty/pull/160 we onboard `libthrift` to `hbase-thirdparty` so we can consume libthrift security fixes while maintaining Java 8 compatibility for HBase branch-2, branch-2.5, and branch-2.6. 

Use the new `hbase-shaded-thrift` in `hbase-thirdparty`.


apurtell:
I applied this locally and built with `-Dhbase-thirdparty=4.1.14-SNAPSHOT` and the `hbase-thrift` test results look good. 

```text
Tests run: 155, Failures: 0, Errors: 2, Skipped: 0
```

The two erroring tests are `TestThriftSpnegoHttpFallbackServer` and `TestThriftSpnegoHttpServer` failing on `NoMatchingRule: No rules applied to xxx@XXX.XXX.COM`

This is an unrelated environmental problem because Kerberos tests using the miniKDC don't install matching rules to handle real kerberos credentials that might be found in a user's environment. I recently fixed that issue in the Phoenix tests and will follow up for HBase on another JIRA

apurtell:
Rebased. 

Still waiting for hbase-thirdparty release vote before this can come out of draft.

hbase-thrift/src/main/java/org/apache/hadoop/hbase/thrift2/generated/NamespaceDescriptor.java
Apache9, line 1: Why this file is gone?

Apache9 (commented):
Is it possible to ban libthrift dependencies completely in maven enforcer?

hbase-thrift/src/main/java/org/apache/hadoop/hbase/thrift2/generated/NamespaceDescriptor.java
apurtell: As part of this work I wiped the generated sources directory and then regenerated the IDL with the `thrift` compiler version 0.23.0. The IDL has never defined a `NamespaceDescriptor` as far as I know it has always been `TNamespaceDescriptor`. I think the old file was an orphan from HBASE-21650 that has been harmlessly hanging around but is not needed.

apurtell:
Added an enforcer import ban for `org.apache.thrift:*`

Rebased on current `master`.

Apache9 (commented):
You need to add a repository section in pom.xml for building with staging release.

We can remove this section when 4.1.14 is finally out.

apurtell:
Done, thank you for the reminder @Apache9

Apache9 (approved):

PDavid (approved):

apurtell:
Rebased on latest `master` after release of `hbase-thirdparty-4.1.14`. Dropped staging repo.

pankaj72981 (approved):

liuxiaocs7 (approved):

PR #8306, opened 2026-06-03
HBASE-30198 Javadoc build failures in branch-2/2.5/2.6

apurtell:
There are three fairly recent regressions in Javadoc processing for branch-2 / 2.5 / 2.6. 

The duplicate test class name `org.apache.hadoop.hbase.client.TestBufferedMutator` in both `hbase-client` and `hbase-server` modules causes the JDK8 doclet to NPE. This is a break fix. 

ApacheDS dependencies were added only to the Hadoop 3 profile by HBASE-29244 so the aggregate Test Javadoc can't resolve `org.apache.directory.*` imports. This is a cleanup. 

`TestLeaseRenewal` imports `DummyDFSOutputStream`, which can't compile under hadoop-2, but it can be excluded. This is a cleanup.


Apache9 (approved):
Thanks @apurtell for the quick fix!

PR #8321, opened 2026-06-06
HBASE-30159 Make hash algorithm configurable for HashTable and SyncTable

jinhyukify:
https://issues.apache.org/jira/browse/HBASE-30159

### DIFF
--- a/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/HashTable.java
+++ b/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/HashTable.java
@@ -58,7 +58,6 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.apache.hbase.thirdparty.com.google.common.base.Charsets;
-import org.apache.hbase.thirdparty.com.google.common.base.Throwables;
 import org.apache.hbase.thirdparty.com.google.common.collect.Ordering;
 
 @InterfaceAudience.Private
@@ -68,7 +67,14 @@ public class HashTable extends Configured implements Tool {
 
   private static final int DEFAULT_BATCH_SIZE = 8000;
 
+  /**
+   * Default hash algorithm. Kept as MD5 so that manifests produced by older versions, which did not
+   * record an algorithm, remain readable.
+   */
+  static final String DEFAULT_HASH_ALGORITHM = "MD5";
+
   private final static String HASH_BATCH_SIZE_CONF_KEY = "hash.batch.size";
+  final static String HASH_ALGORITHM_CONF_KEY = "hash.algorithm";
   final static String PARTITIONS_FILE_NAME = "partitions";
   final static String MANIFEST_FILE_NAME = "manifest";
   final static String HASH_DATA_DIR = "hashes";
@@ -99,6 +105,7 @@ public class HashTable extends Configured implements Tool {
     long endTime = 0;
     boolean ignoreTimestamps;
     boolean rawScan;
+    String hashAlgorithm = DEFAULT_HASH_ALGORITHM;
 
     List<ImmutableBytesWritable> partitions;
 
@@ -138,6 +145,7 @@ public class HashTable extends Configured implements Tool {
         p.setProperty("endTimestamp", Long.toString(endTime));
       }
       p.setProperty("rawScan", Boolean.toString(rawScan));
+      p.setProperty("hashAlgorithm", hashAlgorithm);
 
       try (OutputStreamWriter osw = new OutputStreamWriter(fs.create(path), Charsets.UTF_8)) {
         p.store(osw, null);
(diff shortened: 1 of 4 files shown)


hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestHashTable.java
```diff
@@ -181,4 +183,173 @@ ImmutableMap.<Integer, ImmutableBytesWritable> builder()
...
+        "--hashAlgorithm=SHA-256", tableName.getNameAsString(), testDir.toString() });
+    assertEquals(0, code, "test job failed");
+
+    FileSystem fs = TEST_UTIL.getTestFileSystem();
+    HashTable.TableHash tableHash = HashTable.TableHash.read(fs.getConf(), testDir);
+    assertEquals("SHA-256", tableHash.hashAlgorithm,
+      "manifest must record the algorithm used to produce the digests");
+
+    ImmutableMap<Integer, ImmutableBytesWritable> expectedHashes =
+      ImmutableMap.<Integer, ImmutableBytesWritable> builder()
```
jinhyukify, line 230: Generated SHA-256 hashes using this script.
https://gist.github.com/jinhyukify/705b9b6b81d06973556d5645353efcbb

hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/HashTable.java
```diff
@@ -665,7 +679,7 @@ private boolean doCommandLine(final String[] args) {
 
       for (int i = 0; i < args.length - NUM_ARGS; i++) {
         String cmd = args[i];
-        if (cmd.equals("-h") || cmd.startsWith("--h")) {
+        if (cmd.equals("-h") || cmd.equals("--help")) {
```
jinhyukify, line 683: Without this change, we cannot use options that start with `--h` including `--hashAlgorithm`

hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/HashTable.java
```diff
@@ -189,6 +197,8 @@ void readPropertiesFile(FileSystem fs, Path path) throws IOException {
       if (endTimeString != null) {
         endTime = Long.parseLong(endTimeString);
       }
+
+      hashAlgorithm = p.getProperty("hashAlgorithm", DEFAULT_HASH_ALGORITHM);
```
jinhyukify, line 201: For old HashTable files, we fall back to the original MD5 algorithm.

hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestHashTable.java
```diff
@@ -181,4 +183,173 @@ ImmutableMap.<Integer, ImmutableBytesWritable> builder()
...
+      p.store(out, null);
+    }
+
+    // write an empty partitions file so TableHash.read() succeeds
+    HashTable.TableHash empty = new HashTable.TableHash();
+    empty.partitions = new java.util.ArrayList<>();
+    empty.writePartitionFile(fs.getConf(), new Path(testDir, HashTable.PARTITIONS_FILE_NAME));
+
+    HashTable.TableHash tableHash = HashTable.TableHash.read(fs.getConf(), testDir);
+    assertEquals("MD5", tableHash.hashAlgorithm,
```
junegunn: Can we refer to `DEFAULT_HASH_ALGORITHM` here?

hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/HashTable.java
```diff
@@ -640,6 +652,8 @@ private static void printUsage(final String errorMsg) {
     System.err.println(" families          comma-separated list of families to include");
     System.err.println(" ignoreTimestamps  if true, ignores cell timestamps");
     System.err.println("                   when calculating hashes");
+    System.err.println(" hashAlgorithm     MessageDigest algorithm to use for batch hashes");
+    System.err.println("                   (defaults to " + DEFAULT_HASH_ALGORITHM + ")");
```
junegunn, line 657: Would it help to show the available candidates?

```
jshell> System.out.println(java.security.Security.getAlgorithms("MessageDigest"));
[SHA3-512, SHA-1, SHA-384, SHA3-384, SHA-224, SHA-512/256, SHA-256, MD2, SHA-512/224, SHA3-256, SHA-512, SHA3-224, MD5]
```

Too many?

jinhyukify: Thanks for the review! 😄 
I updated the usage text to list a few popular hash algorithms from the supported set.

junegunn (approved):

PR #8324, opened 2026-06-08
HBASE-30209 MobFileCleanerChore is not registered for online configur…

liuxiaocs7:
…ation updates

- see: HBASE-30209

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
@@ -1964,9 +1964,6 @@ public class HMaster extends HBaseServerBase<MasterRpcServices> implements Maste
   }
 
   protected void stopChores() {
-    if (mobFileCleanerChore != null) {
-      configurationManager.deregisterObserver(mobFileCleanerChore);
-    }
     shutdownChore(mobFileCleanerChore);
     shutdownChore(mobFileCompactionChore);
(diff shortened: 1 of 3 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/master/HMaster.java
```diff
@@ -1963,6 +1964,9 @@ private void stopProcedureExecutor() {
   }
 
   protected void stopChores() {
+    if (mobFileCleanerChore != null) {
+      configurationManager.deregisterObserver(mobFileCleanerChore);
+    }
```
junegunn: We don't deregister the other chores in `stopChores`. Any reason to deregister this one specifically? We could drop this for consistency if there's no strong reason to keep it.

liuxiaocs7: Thanks for the comments, agree with your point, removed!

junegunn (approved):

VladRodionov (approved):
LGTM. MobFileCleanerChore already supports onConfigurationChange(...), so registering it with the master ConfigurationManager in initMobCleaner() makes sense. The added tests cover both sides: the chore is registered by HMaster and the chore actually reacts to thread-count changes.

liuxiaocs7:
Many thanks @junegunn and @VladRodionov for the reviewing!

PR #8325, opened 2026-06-09
HBASE-30023 Refactor write path, prefetch, and compaction cache popul…

VladRodionov:
### Summary

This PR continues the pluggable block cache refactoring by migrating cache population paths to use CacheAccessService instead of calling BlockCache directly where the code is part of HFile cache access behavior.

The change keeps the runtime behavior legacy-compatible. Existing BlockCache implementations are still used through BlockCacheBackedCacheAccessService. This PR does not refactor BlockCacheFactory, does not introduce CacheEngine-backed runtime wiring, and does not attempt to migrate all remaining BlockCache references.

The main goal is to make CacheAccessService the HBase-facing facade for cache lookup, insertion, invalidation, admission checks, and related cache operations while preserving existing behavior.

### Changes

* Migrates additional cache population code paths from direct BlockCache usage to CacheAccessService.
* Preserves simple cacheBlock(...) overloads on CacheAccessService so legacy-style call sites do not need to construct CacheWriteContext manually.
* Keeps CacheWriteContext available for future richer policy-aware cache insertion paths.
* Adds/uses compatibility forwarding for cache admission hooks such as shouldCacheBlock(...) and shouldCacheFile(...) where existing paths depend on them.
* Adds/uses CacheAccessService.ifEnabled(...) for call sites that need the old Optional#ifPresent(...) style behavior when cache is disabled.
* Preserves disabled-cache behavior through NoOpCacheAccessService.
* Updates affected tests.

### Compatibility notes

This PR intentionally does not force every cache insertion call site to specify a precise CacheWriteSource.

Existing HBase cache population code does not always expose whether a block insertion comes from flush, compaction, bulk load, or another writer path. Guessing that source could introduce incorrect policy behavior later. Therefore, this PR keeps compatibility-oriented overloads and uses richer context only where the information is already available and meaningful.

### Out of scope

* No BlockCacheFactory refactoring.
* No CacheEngine adapter migration.
* No migration of LruBlockCache, LruAdaptiveBlockCache, TinyLfuBlockCache, or BucketCache to CacheEngine.
* No removal of BlockCache from CacheConfig.
* No removal of getBlockCache().
* No migration of BlockCache.iterator().
* No diagnostic JSP/admin page migration.
* No broad migration of implementation-specific cache tests.
* No topology-backed runtime wiring.

### Testing

Ran the affected HFile cache-on-write test successfully:

bash mvn -pl hbase-server -Dtest=TestHFile#testWriterCacheOnWriteSkipDoesNotLeak test 

Also verified build/package after resolving a stale local class output issue unrelated to this PR.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheWriteSource.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheWriteSource.java
@@ -36,6 +36,16 @@ public enum CacheWriteSource {
   READ_MISS,
 
   /**
+   * Block was inserted from the HFile write/cache-on-write path.
+   * <p>
+   * This source covers write-side population where the caller does not expose whether the HFile is
+   * being produced by flush, compaction, bulk load, or another writer. More specific write-side
+   * sources can be added later when that context is available at the call site.
+   * </p>
+   */
+  WRITE_PATH,
+
+  /**
    * Cache population during flush output generation.
    */
   FLUSH,
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestLazyDataBlockDecompression.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestLazyDataBlockDecompression.java
@@ -132,6 +132,10 @@ public class TestLazyDataBlockDecompression {
     reader.close();
   }
 
+  /*
+   * TODO: migrate this test to use new HBASE-30018 APIs and remove the need to cast to
+   * LruBlockCache
+   */
   @TestTemplate
   public void testCompressionIncreasesEffectiveBlockCacheSize() throws Exception {
     // enough room for 2 uncompressed block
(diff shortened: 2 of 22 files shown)


wchevreuil (approved):
LGTM. UT failures look unrelated.

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/TestLazyDataBlockDecompression.java
```diff
@@ -132,6 +132,10 @@ private static void cacheBlocks(Configuration conf, CacheConfig cacheConfig, Fil
     reader.close();
   }
 
+  /*
+   * TODO: migrate this test to use new HBASE-30018 APIs and remove the need to cast to
+   * LruBlockCache
+   */
```
taklwu, line 138: nit: if you have plan to update it , we can have this in the upcoming JIRA without this reminder.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/cache/CacheWriteSource.java
```diff
@@ -35,6 +35,16 @@ public enum CacheWriteSource {
...
+  /**
+   * Block was inserted from the HFile write/cache-on-write path.
+   * <p>
+   * This source covers write-side population where the caller does not expose whether the HFile is
+   * being produced by flush, compaction, bulk load, or another writer. More specific write-side
+   * sources can be added later when that context is available at the call site.
+   * </p>
+   */
+  WRITE_PATH,
+
```
taklwu, line 47: nit: I got this is a new type, but in this PR, we didn't use it and IMO we should add new type when it's being used?

taklwu (approved):

PR #8326, opened 2026-06-09
HBASE-30212 Netty should allow every supported TLS ciphers by default

meszibalu:
Netty will allow to use every supported cipher at the client side by default, so clients can use the widest range of ciphers.


PDavid (approved):
Many thanks for fixing this. 👍

petersomogyi (approved):

PR #8327, opened 2026-06-09
HBASE-29774 incremental backup fails on empty WAL

DieterDP-ng:
Add handling for WALHeaderEOFException in WALInputFormat so WALPlayer now skips empty WAL files rather than crashing. Empty WAL files (also lacking a header) either originate from crashed Region Servers, or might exist for a very short period when using the legacy WAL writer on newly created active WALs.

This solves a problem where incremental backups could fail when converting WAL files to HFiles. In this use case, all WAL files passed to the WALPlayer are rolled (so no longer active), so empty WAL files were due to crashed Region Servers.

(This is an updated PR of the [one originally created by my colleague @thomassngdata](https://github.com/apache/hbase/pull/7762). It applies some review remarks I had when reviewing that PR, but Thomas is currently unavailable to process them himself.)

### DIFF
--- a/hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALInputFormat.java
+++ b/hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALInputFormat.java
@@ -18,6 +18,7 @@
 package org.apache.hadoop.hbase.mapreduce;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
 
 import java.util.ArrayList;
 import java.util.List;
@@ -36,6 +37,7 @@ import org.apache.hadoop.hbase.util.EnvironmentEdgeManager;
 import org.apache.hadoop.mapreduce.InputSplit;
 import org.apache.hadoop.mapreduce.Job;
 import org.apache.hadoop.mapreduce.JobContext;
+import org.apache.hadoop.mapreduce.TaskAttemptContext;
 import org.apache.hadoop.mapreduce.lib.input.FileInputFormat;
 import org.junit.jupiter.api.BeforeAll;
 import org.junit.jupiter.api.Tag;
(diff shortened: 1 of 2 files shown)


DieterDePaepe:
As I've helped edit this PR myself, we need someone else of HBase community to give an approval before I can merge. @ndimiduk or @hgromer - you've reviewed backup related PRs before. Would appreciate if either of you had time for a review.

hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALInputFormat.java
```diff
@@ -120,4 +122,44 @@ public void testHandlesArchivedWALFiles() throws Exception {
...
+      WALInputFormat wif = new WALInputFormat();
+      List<InputSplit> splits = wif.getSplits(ctx);
+      assertEquals(1, splits.size());
+      WALInputFormat.WALSplit split = (WALInputFormat.WALSplit) splits.get(0);
+
+      // This should not throw WALHeaderEOFException - it should return false for nextKeyValue()
+      reader.initialize(split, taskCtx);
+      // nextKeyValue() should return false since the file is empty (reader is null)
+      assertFalse(reader.nextKeyValue());
+    }
```
VladRodionov, line 166: nit: Could we delete emptyWalFile in a finally block? The test creates a WAL file under the mini-cluster WAL root and does not appear to clean it up. Even if the test cluster cleanup eventually removes it, local cleanup would make the test more isolated.

Otherwise, lgtm.

VladRodionov (commented):
lgtm.

hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALInputFormat.java
DieterDP-ng: Added.

DieterDePaepe:
@VladRodionov - mind giving a formal approval?

ndimiduk (approved):

ndimiduk:
Wait what's this, we need to manually approve PR workflows now??

PR #8331, opened 2026-06-10
HBASE-30217 Remove dead code getRegionId and isTemporaryDirectoryName…

liuxiaocs7:
… in HRegionFileSystem

- see: HBASE-30217

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionFileSystem.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionFileSystem.java
@@ -78,10 +78,20 @@ public class HRegionFileSystem {
   /** Name of the region info file that resides just under the region directory. */
   public final static String REGION_INFO_FILE = ".regioninfo";
 
-  /** Temporary subdirectory of the region directory used for merges. */
+  /**
+   * Temporary subdirectory of the region directory used for merges.
+   * @deprecated Since 3.0.0, will be removed in 4.0.0. Merges now write directly to the merged
+   *             region directory under the table dir since HBASE-26187.
+   */
+  @Deprecated
   public static final String REGION_MERGES_DIR = ".merges";
 
-  /** Temporary subdirectory of the region directory used for splits. */
+  /**
+   * Temporary subdirectory of the region directory used for splits.
+   * @deprecated Since 3.0.0, will be removed in 4.0.0. Splits now write directly to the daughter
+   *             region directory under the table dir since HBASE-26187.
+   */
+  @Deprecated
   public static final String REGION_SPLITS_DIR = ".splits";
 
   /** Temporary subdirectory of the region directory used for compaction output. */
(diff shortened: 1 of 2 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionFileSystem.java
```diff
@@ -78,10 +78,20 @@ public class HRegionFileSystem {
...
   public static final String REGION_MERGES_DIR = ".merges";
 
-  /** Temporary subdirectory of the region directory used for splits. */
+  /**
+   * Temporary subdirectory of the region directory used for splits.
+   * @deprecated Since 3.0.0, will be removed in 4.0.0. Splits now write directly to the daughter
+   *             region directory under the table dir since HBASE-26187.
+   */
+  @Deprecated
   public static final String REGION_SPLITS_DIR = ".splits";
```
wchevreuil: Do we really need to deprecate those rather than simply remove it? HRegionFileSystem is marked as InterfaceAudience.Private, so isn't intended for public use.

wchevreuil (approved):

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionFileSystem.java
liuxiaocs7: Thanks for the comments, agree with your points, I've removed the two constants directly. The only usage was in `TestHdfsSnapshotHRegion`, which I've updated to use string literals with a comment explaining they represent legacy directories from before HBASE-26187.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestHdfsSnapshotHRegion.java
```diff
@@ -92,10 +92,9 @@ public void testSnapshottingWithTmpSplitsAndMergeDirectoriesPresent() throws Exc
     Path regionDirectoryPath = new Path(tableDir, encodedName);
     TEST_UTIL.getTestFileSystem()
       .create(new Path(regionDirectoryPath, HRegionFileSystem.REGION_TEMP_DIR));
-    TEST_UTIL.getTestFileSystem()
-      .create(new Path(regionDirectoryPath, HRegionFileSystem.REGION_SPLITS_DIR));
-    TEST_UTIL.getTestFileSystem()
-      .create(new Path(regionDirectoryPath, HRegionFileSystem.REGION_MERGES_DIR));
+    // Legacy directories created by split/merge before HBASE-26187
+    TEST_UTIL.getTestFileSystem().create(new Path(regionDirectoryPath, ".splits"));
```
Apache9: Do we still need these two directories in this test?

liuxiaocs7: Thanks for your comments, agree with your points, removed them.

guluo2016 (approved):
Thanks

PR #8333, opened 2026-06-10
HBASE-30166 ADDENDUM TestPrefetchPersistence may still fail due to prefetch delay

wchevreuil:


PDavid (approved):
Thanks, looks good to me. 👍

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/bucket/TestPrefetchPersistence.java
```diff
@@ -138,14 +142,8 @@ public void cleanup() {
...
-    int retries = 0;
-    while (
-      !reader.prefetchComplete()
-        && !bucketCache.fullyCachedFiles.containsKey(storeFilePath.getName()) && retries < 5
-    ) {
-      Thread.sleep(500);
-      retries++;
-    }
+    Waiter.waitFor(conf, 30000, () -> reader.prefetchComplete()
+      || bucketCache.fullyCachedFiles.containsKey(storeFilePath.getName()));
```
guluo2016, line 146: Nit: the polling interval is 100ms now by default

guluo2016 (approved):
Thanks

PR #8340, opened 2026-06-11
HBASE-30221 FNFE may occur when executing `hbase snapshot info --list-snapshots`

guluo2016:
Details see: [HBASE-30221](https://issues.apache.org/jira/browse/HBASE-30221)

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
@@ -628,6 +628,9 @@ public final class SnapshotInfo extends AbstractHBaseTool {
     Path rootDir = CommonFSUtils.getRootDir(conf);
     FileSystem fs = FileSystem.get(rootDir.toUri(), conf);
     Path snapshotDir = SnapshotDescriptionUtils.getSnapshotsDir(rootDir);
+    if (!fs.exists(snapshotDir)) {
+      return Collections.emptyList();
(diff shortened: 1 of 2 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
```diff
@@ -628,6 +628,9 @@ public static List<SnapshotDescription> getSnapshotList(final Configuration conf
     Path rootDir = CommonFSUtils.getRootDir(conf);
     FileSystem fs = FileSystem.get(rootDir.toUri(), conf);
     Path snapshotDir = SnapshotDescriptionUtils.getSnapshotsDir(rootDir);
+    if (!fs.exists(snapshotDir)) {
+      return Collections.emptyList();
+    }
     FileStatus[] snapshots = fs.listStatus(snapshotDir,
```
Apache9: Is it possible to test whether snapshots is null or empty, so we can save on request to HDFS namenode?

guluo2016: Good point, thanks.
I checked the code again and luckily found that `CommonFSUtils.listStatus` can handle this case directly, so let me update the code, thanks again.

Apache9 (approved):

guluo2016:
Thanks for reviewing @Apache9 
The failed UTs seem unrelated to this change.
Let me merge it later.

PR #8341, opened 2026-06-11
HBASE-30209 MobFileCleanerChore is not registered for online configur…

liuxiaocs7:
…… (#8324)

- for branch-2
- see: HBASE-30209

PR #8342, opened 2026-06-12
HBASE-30223 TestRegionMover2.testWithMergedRegions may merge disjoint…

Apache9:
… regions and cause tests to fail

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMover2.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMover2.java
@@ -17,6 +17,10 @@
  */
 package org.apache.hadoop.hbase.util;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMover2.java
```diff
@@ -112,23 +119,36 @@ public void testWithMergedRegions() throws Exception {
     table.put(puts);
     admin.flush(tableName);
     HRegionServer regionServer = cluster.getRegionServer(0);
+    RegionInfo regionA = regionServer.getRegions(tableName).get(0).getRegionInfo();
```
liuxiaocs7: If `regionA` is the last region of the table, in which case its endKey is empty, and regionB would then match the first region so we'd end up merging two disjoint regions again?

liuxiaocs7:
https://github.com/apache/hbase/pull/8341 encountered the same error, thanks for the fixing!

hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMover2.java
Apache9: Not likely but OK, let me make it more robust, there is also possible that the region server does not contain any regions of this table, let me deal with it too.

liuxiaocs7 (approved):

Apache9:
The failed UTs are not touched and not related.

Let me merge.

PR #8347, opened 2026-06-12
HBASE-30142 Resolve NPE while running recoverUnknown command because null RegionLocation

Umeshkumar9414:
MasterRpcServices.scheduleSCPsForUnknownServers() has no null guard before calling isServerUnknown(). With the old code, regions with null serverName (OFFLINE/FAILED_OPEN in-between state) would spuriously trigger a ServerCrashProcedure. The fix prevents that.


virajjasani (approved):

virajjasani (approved):
+1

PR #8353, opened 2026-06-12
HBASE-30225 Performance degradation observed on ycsb reads benchmark after HBASE-29727

wchevreuil:


hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
```diff
@@ -115,7 +114,7 @@ public BlockCacheKey(String hfileName, String cfName, String regionName, long of
   public BlockCacheKey(Path hfilePath, long offset, boolean isPrimaryReplica, BlockType blockType) {
     this(hfilePath.getName(), hfilePath.getParent().getName(),
       hfilePath.getParent().getParent().getName(), offset, isPrimaryReplica, blockType,
-      HFileArchiveUtil.isHFileArchived(hfilePath));
```
Apache9, line 118: Why here we can use false directly?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
```diff
@@ -97,9 +96,9 @@ public BlockCacheKey(String hfileName, String cfName, String regionName, long of
     this.offset = offset;
     this.blockType = blockType;
     // Use string pool for file, region and cf values
-    this.hfileName = HFILE_NAME_POOL.intern(hfileName);
-    this.regionName = (regionName != null) ? REGION_NAME_POOL.intern(regionName) : null;
-    this.cfName = (cfName != null) ? CF_NAME_POOL.intern(cfName) : null;
+    this.hfileName = hfileName;
```
Apache9: If we do not do intern here, why we still put the above 3 string pools in this class?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
```diff
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hbase.io.hfile;
 
+import org.apache.commons.lang3.mutable.MutableBoolean;
```
Apache9: Where do we use this import?

Apache9 (commented):
Mind explaining more about the changes here? I do not fully get what you are trying to fix here...

wchevreuil:
> Mind explaining more about the changes here? I do not fully get what you are trying to fix here...

As explained in the jira description: 

When executing ycsb read workloads, we observed a ~30% latency degradation (please see flamegrahs attached to the jira).

The problem was that we added logic for parsing the file Path into region name, family name, as well checks for archiving all on the BlockCacheKey constructor used by HFileReaderImpl on the beginning of each block read. As seen on the flame graphs attached covering a five minutes window on one of the RSes, around 30% of the CPU time was spent on the BlockCacheKey constructor, either calling Path.getParent() or HFileUtils.isHFileArchived().

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
wchevreuil: We are doing intern from readers and writers, we need those to access a single pool per file/region/cf. Since both reader and writer classes already reference BlockCacheKey, and the string re-use is relevant in the BlockCacheKey usage, I found it makes sense to keep the pools in the BlockCacheKey.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
wchevreuil: We are actually only using this constructor on tests now. Because the HFileUtils.isArchived call seems too expensive to be performed on every block cache key creation, that check should be delegated to the reader, which can do it once during it's initialisation. Writers don't even need it, since a being written file is guaranteed to be "non-archived", so writers can benefit from constructors that alwas set archived as false. Let me add javadoc to explain this.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
Apache9: Why not introduce a new class to do this?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
Apache9: Then why not just remoev this flag in BlockCacheKey if it does not change?

Apache9:
> > Mind explaining more about the changes here? I do not fully get what you are trying to fix here...
> 
> As explained in the jira description:
> 
> When executing ycsb read workloads, we observed a ~30% latency degradation (please see flamegrahs attached to the jira).
> 
> The problem was that we added logic for parsing the file Path into region name, family name, as well checks for archiving all on the BlockCacheKey constructor used by HFileReaderImpl on the beginning of each block read. As seen on the flame graphs attached covering a five minutes window on one of the RSes, around 30% of the CPU time was spent on the BlockCacheKey constructor, either calling Path.getParent() or HFileUtils.isHFileArchived().

So the intention here, is to not always call intern when creating BlockCacheKey, and move intern to other places?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
wchevreuil: It might change when it's created by a Reader (see HFileReaderImpl#1352).

wchevreuil:
> > > Mind explaining more about the changes here? I do not fully get what you are trying to fix here...
> > 
> > 
> > As explained in the jira description:
> > When executing ycsb read workloads, we observed a ~30% latency degradation (please see flamegrahs attached to the jira).
> > The problem was that we added logic for parsing the file Path into region name, family name, as well checks for archiving all on the BlockCacheKey constructor used by HFileReaderImpl on the beginning of each block read. As seen on the flame graphs attached covering a five minutes window on one of the RSes, around 30% of the CPU time was spent on the BlockCacheKey constructor, either calling Path.getParent() or HFileUtils.isHFileArchived().
> 
> So the intention here, is to not always call intern when creating BlockCacheKey, and move intern to other places?

Sort of. The main problem is not the intern call itself, but the parsing of file name, region and CF name from the path. Doing it inside the BlockCacheKey constructor is too costly, we saw 10% degradation on latency and throughput of ycsb workloads. Moving this parsing to HFileWriterImpl/HFileReaderImpl initialisation means we only need to do it once, as all block cache keys on each writer/reader instance will refer to same values. 

We could maybe leave the pools and intern calls private to BlockCacheKey, but we would miss the original parsed strings from getting pooled. Less efficient, but I think it's not a big deal, as we would only be doing it at the reader/writer level. Code-wise, I think it would be cleaner and less confusing. Let me do it.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
```diff
@@ -1336,9 +1346,8 @@ public HFileBlock readBlock(long dataBlockOffset, long onDiskBlockSize, final bo
     // the other choice is to duplicate work (which the cache would prevent you
     // from doing).
 
-    BlockCacheKey cacheKey =
-      new BlockCacheKey(path, dataBlockOffset, this.isPrimaryReplicaReader(), expectedBlockType);
-
+    BlockCacheKey cacheKey = new BlockCacheKey(path.getName(), family, region, dataBlockOffset,
```
VladRodionov: A bit off top, I am really surprised that that parsing qualified file name can be expensive at all, but you see that in your benchmarks. Q: why did not you cache file name in this case? path.getName()?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
```diff
@@ -105,17 +104,18 @@ public BlockCacheKey(String hfileName, String cfName, String regionName, long of
 
   /**
    * Construct a new BlockCacheKey using a file path. File, column family and region information
-   * will be extracted from the passed path. This should be used when inserting keys into the cache,
-   * so that region cache metrics are recorded properly.
+   * will be extracted from the passed path, so that region cache metrics are recorded properly/
```
VladRodionov: nit: properly

VladRodionov (commented):
The direction makes sense to me, but I think HFileReaderImpl should also cache the HFile name once.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
wchevreuil: The culprits, as seem by the async profiler flamegraphs, are Path.getParent() and HFileArchiveUtil.isFileArchived(). I didn't think of doing the same for the file name because that didn't seem to cause problems, but yeah, could do it for the sake of code consistency.

wchevreuil:
Current compile failures were introduced by HBASE-30221. There's already an addendum PR for fixing it: https://github.com/apache/hbase/pull/8361

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
Apache9: Then can we just remove this constructor? Or leave the code as is since it only calls from test code? It does not affect real performance if it only calls in test code right?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
```diff
@@ -1246,8 +1259,8 @@ public HFileBlock getMetaBlock(String metaBlockName, boolean cacheBlock) throws
     synchronized (metaBlockIndexReader.getRootBlockKey(block)) {
       // Check cache for block. If found return.
       long metaBlockOffset = metaBlockIndexReader.getRootBlockOffset(block);
-      BlockCacheKey cacheKey =
-        new BlockCacheKey(path, metaBlockOffset, this.isPrimaryReplicaReader(), BlockType.META);
+      BlockCacheKey cacheKey = new BlockCacheKey(fileName, family, region, metaBlockOffset,
+        this.isPrimaryReplicaReader(), BlockType.META, archived);
```
Apache9, line 1263: So this is one of the optimization right? We calculate archived once when constructing the HFileReaderImpl, and do not need to calcuate it every time when constructing a BlockCacheKey?

Apache9 (commented):
Where are the intern calls now?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/BlockCacheKey.java
wchevreuil: Yes.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/HFileReaderImpl.java
wchevreuil: Correct. Same for region and CF name.

wchevreuil:
> Where are the intern calls now?

Back to the "BlockCacheKey(String hfileName, String cfName, String regionName, long offset, boolean isPrimaryReplica, BlockType blockType, boolean archived)" (lines 99 to 101).

Apache9:
> > Where are the intern calls now?
> 
> Back to the "BlockCacheKey(String hfileName, String cfName, String regionName, long offset, boolean isPrimaryReplica, BlockType blockType, boolean archived)" (lines 99 to 101).

OK, I missed your previous comment about keeping the intern calls inside BlockCacheKey. I'm OK with moving the intern calls out if there are benefits, the only concern is to also move the StringPools out. If you think keeping it in BlockCacheKey is not the biggest bottleneck here I'm also OK with keeping it as is.

Thanks.

petersomogyi (approved):
I suggest adding visibility annotation for the BlockCacheKey's test-intended constructor.

petersomogyi (approved):

PR #8357, opened 2026-06-15
HBASE-29555 TestRollbackSCP.testFailAndRollback fails sometimes because non empty Scheduler queue

SwaraliJoshi:
## Summary

`TestRollbackSCP.testFailAndRollback` is flaky: it intermittently fails with
`java.lang.IllegalArgumentException: scheduler queue not empty` from
`ProcedureExecutor.load()` while restarting the master procedure executor.

The test restarts the `ProcedureExecutor` **in place**, reusing the same
executor and `MasterProcedureScheduler` instances to simulate a failover. While
the executor is being reloaded, other still-running threads of the live
mini-cluster master can push a procedure back into the shared scheduler in the
small window between `scheduler.clear()` and `ProcedureExecutor.load()`'s
`Preconditions.checkArgument(scheduler.size() == 0, ...)`.

Two producers were identified:
- the `asyncTaskExecutor` callback that wakes a procedure after an async meta
  update (e.g. `AssignmentManager.persistToMeta`), and
- an incoming `reportRegionStateTransition` RPC from a live region server,
  handled on an `RpcServer` handler thread, which wakes a procedure through
  `ProcedureEvent.wake` -> `scheduler.addFront`.

This is a test-infrastructure issue: a real master failover starts a fresh
process with a fresh executor/scheduler, so the production `load()` precondition
is not affected. The fix is therefore confined to test code.

## Solution (`ProcedureTestingUtility.restart()`, test-only)

The fix makes the test's in-place reload faithfully reproduce a real master failover by closing the two gaps, with no retry loop.

1. Gap 1 — no stale async wake survives the reload (awaitAsyncTaskExecutorTermination)
A new helper that, after the executor is stopped (which shuts the asyncTaskExecutor down), waits for that executor to fully terminate (bounded 60s) before the reload clears the scheduler. This guarantees any pending persistToMeta/regionClosedAbnormally wake-up callback has finished, and any future completing afterward is rejected by the shut-down executor. It reproduces production's guarantee that "the old process's async work is gone before load()."

2. Gap 2 — no region report during the reload (RELOAD_LOCK + report override)
A fair ReentrantReadWriteLock RELOAD_LOCK.
AssignmentManagerForTest.reportRegionStateTransition(...) is overridden to take the read lock via non-blocking tryLock; if the reload holds the write lock it throws PleaseHoldException (the RegionServer retries). This mirrors production exactly — MasterRpcServices.checkServiceStarted() rejects reports until the master finishes initializing (which is after load()). tryLock (not a blocking lock) is deliberate: it makes a report never block, so the AM re-init inside the reload can't deadlock behind a blocked report, and RPC handler threads stay free.

## Test plan

- [x] Ran `TestRollbackSCP` 100x consecutively with the fix: 100/100 passed (twice).
- [x] Control: reverted the fix and reran: reproduced the failure within a few iterations.
- [x] Regression: ran affected `hbase-procedure` tests and a representative
      `hbase-server` subset (TestSCP, TestProcedureAdmin,
      TestTransitRegionStateProcedure, TestCreateTableProcedure): all pass.

### DIFF
--- a/hbase-procedure/src/test/java/org/apache/hadoop/hbase/procedure2/ProcedureTestingUtility.java
+++ b/hbase-procedure/src/test/java/org/apache/hadoop/hbase/procedure2/ProcedureTestingUtility.java
@@ -28,8 +28,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Set;
 import java.util.concurrent.Callable;
-import java.util.concurrent.ExecutorService;
-import java.util.concurrent.TimeUnit;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
@@ -127,50 +125,14 @@ public final class ProcedureTestingUtility {
       stopAction.call();
     }
     procExecutor.join();
-    // HBASE-29555: Callbacks that wake up a procedure after an async operation (such as updating
-    // meta) run on the asyncTaskExecutor, not on a PEWorker thread. ProcedureExecutor.stop() shuts
-    // the asyncTaskExecutor down, but ProcedureExecutor.join() only waits a bounded amount of time
-    // for it to terminate before giving up. If such a wake up task is still pending, it can call
-    // scheduler.addFront() after we clear the scheduler below and after the executor has been
-    // restarted, leaving the scheduler queue non-empty and tripping the Preconditions check in
-    // ProcedureExecutor.load(). Since we reuse the same executor (and scheduler) instance across
-    // this simulated restart, wait for the already shutdown asyncTaskExecutor to fully terminate so
-    // any pending wake up task has finished before we clear the scheduler.
-    awaitAsyncTaskExecutorTermination(procExecutor);
+    procExecutor.getScheduler().clear();
 
     // nothing running...
 
     // re-start
     LOG.info("RESTART - Start");
-    // HBASE-29555: External events (for example a region server reporting a region state transition
-    // over RPC) can wake a procedure and push it back into the scheduler in the small window
-    // between clearing the scheduler and ProcedureExecutor.load() checking that it is empty. Those
-    // producers run on threads we do not own here (RPC handlers etc.), so we cannot reliably stop
-    // them from this thread. Instead, reload in a retry loop: if load() fails only because the
-    // scheduler is not empty, stop/drain/clear again and retry. ProcedureExecutor.stop() is
-    // explicitly safe to call after a failed init(), so this is a clean redo of the reload.
-    final int maxRestartAttempts = 20;
-    for (int attempt = 1;; attempt++) {
-      procExecutor.getScheduler().clear();
-      procStore.start(storeThreads);
-      try {
-        procExecutor.init(execThreads, failOnCorrupted);
-        break;
-      } catch (IllegalArgumentException e) {
-        if (
-          attempt >= maxRestartAttempts || e.getMessage() == null
-            || !e.getMessage().contains("scheduler queue not empty")
-        ) {
-          throw e;
-        }
-        LOG.warn("RESTART - scheduler was repopulated during reload (attempt {}/{}), retrying",
-          attempt, maxRestartAttempts, e);
-        procExecutor.stop();
-        procStore.stop(abort);
-        procExecutor.join();
-        awaitAsyncTaskExecutorTermination(procExecutor);
-      }
-    }
+    procStore.start(storeThreads);
+    procExecutor.init(execThreads, failOnCorrupted);
     if (actionBeforeStartWorker != null) {
       actionBeforeStartWorker.call();
     }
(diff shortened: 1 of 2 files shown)


Apache9:
I prefer we just change the code the TestRollbackSCP.

We have lots of other tests which rely on this restart, changing the logic may hide some other bugs we want to expose in these tests.

Thanks.

SwaraliJoshi:
> I prefer we just change the code the TestRollbackSCP.
> 
> We have lots of other tests which rely on this restart, changing the logic may hide some other bugs we want to expose in these tests.
> 
> Thanks.

@Apache9 Yes, I am checking if we have missed any condition to stimulate the real production restart of master due to which this issue is happening only in the test. Please allow me sometime to make the necessary changes.

hbase-procedure/src/test/java/org/apache/hadoop/hbase/procedure2/ProcedureTestingUtility.java
```diff
@@ -125,14 +127,50 @@ public static <TEnv> void restart(ProcedureExecutor<TEnv> procExecutor,
...
-    procExecutor.init(execThreads, failOnCorrupted);
+    // HBASE-29555: External events (for example a region server reporting a region state transition
+    // over RPC) can wake a procedure and push it back into the scheduler in the small window
+    // between clearing the scheduler and ProcedureExecutor.load() checking that it is empty. Those
+    // producers run on threads we do not own here (RPC handlers etc.), so we cannot reliably stop
+    // them from this thread. Instead, reload in a retry loop: if load() fails only because the
+    // scheduler is not empty, stop/drain/clear again and retry. ProcedureExecutor.stop() is
+    // explicitly safe to call after a failed init(), so this is a clean redo of the reload.
+    final int maxRestartAttempts = 20;
+    for (int attempt = 1;; attempt++) {
```
Umeshkumar9414: Emptying scheduler is not out goal here. Check cause of this and try to resolve that. If we are missing something while simulating the Master restart lets add that.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
```diff
@@ -75,12 +85,45 @@ public class TestRollbackSCP {
 
   private static final AtomicBoolean INJECTED = new AtomicBoolean(false);
 
+  // HBASE-29555: guards the in-place procedure-executor reload against region servers concurrently
+  // reporting region state transitions. See restartMasterProcedureExecutorLikeFailover(). Fair so
+  // that, while the reload is waiting for / holding the write lock, new reports are rejected rather
+  // than barging in. A region state report takes the read lock; the reload takes the write lock.
+  private static final ReentrantReadWriteLock RELOAD_LOCK = new ReentrantReadWriteLock(true);
```
Apache9: Actually I do not get the point why here we need fair lock?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
```diff
@@ -75,12 +85,45 @@ public class TestRollbackSCP {
...
+      // reports (via HMaster.checkServiceStarted) until it has finished initializing, which happens
+      // only after the procedure executor has loaded. The test instead reloads the executor in place
+      // under a live master, so a report can wake a procedure and re-populate the scheduler in the
+      // window between clearing it and ProcedureExecutor.load() asserting it is empty. Mirror
+      // production: while the reload holds the write lock, reject reports with PleaseHoldException
+      // (the region server retries). tryLock (non-blocking) is used so a report never blocks the
+      // reload; acquiring the write lock in the reload still waits for any already in-flight report.
+      boolean locked = false;
+      try {
+        locked = RELOAD_LOCK.readLock().tryLock(0, TimeUnit.SECONDS);
```
Apache9: What does timeout 0 means here? No timeout or same with tryLock without any parameters?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
```diff
@@ -160,6 +203,67 @@ public void describeTo(Description description) {
...
+   * before we clear the scheduler, reproducing the "no async work survives the restart" guarantee.
+   */
+  private static void awaitAsyncTaskExecutorTermination(ProcedureExecutor<?> procExec) {
+    ExecutorService asyncTaskExecutor = procExec.getAsyncTaskExecutor();
+    if (asyncTaskExecutor == null) {
+      return;
+    }
+    long deadline = System.currentTimeMillis() + 60000;
+    try {
+      while (!asyncTaskExecutor.awaitTermination(1, TimeUnit.SECONDS)) {
```
Apache9: Why not just wait for 1 minute?

SwaraliJoshi: You are right, let me change it to 1 minute.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
SwaraliJoshi: Fairness is there to prevent the reload thread (the writer) from being starved by a steady stream of region-state reports (the readers). The report path uses the timed tryLock(0, …), which honors the fairness policy: with a fair lock, once the reload is queued waiting for the write lock, new tryLock attempts return false and the report is rejected (PleaseHoldException, RS retries), so the writer acquires promptly. With a non-fair lock, readers "barge" ahead of the waiting writer, so under continuous reports the reload could be delayed.

Fairness prevents reload starvation under report load, happy to drop it if you think that's over-engineering.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
SwaraliJoshi: Per the JDK docs, the no-arg tryLock() always "barges" — it grabs the read lock if the write lock isn't currently held, ignoring fairness even on a fair lock. The timed tryLock(0, unit) honors the fairness policy — on a fair lock it will refuse to acquire if a writer is already waiting. That difference is precisely why the code uses tryLock(0, …) together with the fair lock: it's non-blocking (a report never blocks → no deadlock, RPC handler threads stay free) and it respects the waiting reload writer (so reports get rejected as soon as the reload starts). If you used the no-arg tryLock(), the fair flag would be meaningless (it would barge anyway).

Fair lock + timed tryLock(0) = non-blocking and fairness-honoring; that combination is what makes reports reject-and-retry during reload without starving the reload.

hbase-procedure/src/test/java/org/apache/hadoop/hbase/procedure2/ProcedureTestingUtility.java
SwaraliJoshi: @Umeshkumar9414 Please re-review now. I have made the necessary changes.

PR #8361, opened 2026-06-16
HBASE-30221 ADDENDUM use junit5 import in TestSnapshotInfo

guluo2016:

PR #8362, opened 2026-06-16
HBASE-30227 Disallow JUnit4 imports in source code

liuxiaocs7:
- see: HBASE-30227

image

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -2761,6 +2761,14 @@
                   <reason>Use junit5 instead</reason>
                   <bannedImports>
                     <bannedImport>junit.framework.**</bannedImport>
+                    <bannedImport>org.junit.*</bannedImport>
+                    <bannedImport>org.junit.rules.**</bannedImport>
+                    <bannedImport>org.junit.runner.**</bannedImport>
+                    <bannedImport>org.junit.runners.**</bannedImport>
+                    <bannedImport>org.junit.experimental.**</bannedImport>
+                    <bannedImport>org.junit.internal.**</bannedImport>
+                    <bannedImport>static org.junit.Assert.**</bannedImport>
+                    <bannedImport>static org.junit.Assume.**</bannedImport>
                   </bannedImports>
                 </restrictImports>
               </rules>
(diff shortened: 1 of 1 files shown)


liuxiaocs7:
wait https://github.com/apache/hbase/pull/8361 in

wchevreuil (approved):

Apache9:
In HBASE-30200, we have already banned all junit4 dependencies. Since there are still junit4 assert references in hadoop mini cluster code, I introduced dummy Assert class to delegate all calls to junit5 Assertions. And all the method in this class has RestrictedApi annotation to disable usage in HBase code base.

I think this is enough?

liuxiaocs7:
> In [HBASE-30200](https://issues.apache.org/jira/browse/HBASE-30200), we have already banned all junit4 dependencies. Since there are still junit4 assert references in hadoop mini cluster code, I introduced dummy Assert class to delegate all calls to junit5 Assertions. And all the method in this class has RestrictedApi annotation to disable usage in HBase code base.
> 
> I think this is enough?

Hi, @Apache9, the extra import ban seems still useful as an earlier and clearer guardrail, it fails the build when new JUnit4 imports are added, see https://github.com/apache/hbase/pull/8340 and https://github.com/apache/hbase/pull/8361, WDYT

image

Apache9:
> > In [HBASE-30200](https://issues.apache.org/jira/browse/HBASE-30200), we have already banned all junit4 dependencies. Since there are still junit4 assert references in hadoop mini cluster code, I introduced dummy Assert class to delegate all calls to junit5 Assertions. And all the method in this class has RestrictedApi annotation to disable usage in HBase code base.
> > I think this is enough?
> 
> Hi, @Apache9, the extra import ban seems still useful as an earlier and clearer guardrail, it fails the build when new JUnit4 imports are added, see #8340 and #8361, WDYT
> 
> image

In #8361 , the branch compile is already failing right? You should see an error in pre commit check when adding the calls.

But yeah. this does not fail local builds if you do not enable error prone, with this PR in the local build will also fail too.

Apache9 (approved):

guluo2016 (approved):
Thanks!

liuxiaocs7:
Failed UTs seems unrelated, let me merge.

PR #8377, opened 2026-06-17
HBASE-30220: A replica cluster can have read-only mode disabled even when another active cluster already exists

kgeisz:
https://issues.apache.org/jira/browse/HBASE-30220

## AI Usage

These changes were generated using Claude Opus 4.6 and then modified by hand.

## Summary

In a Read-Replica cluster setup, the active cluster allows both reads and writes, while the replica clusters only allow reads. In this type of setup, only one cluster should be an active cluster at a time. Although the current code prevents starting an active cluster while another one exists, it is still possible to simple turn a running replica into an active cluster despite another active cluster already existing. The changes in this pull request prevent this behavior.

A cluster is promoted from replica to active by changing the `hbase.global.readonly.enabled` variable from `true` to `false` in `hbase-site.xml` and then running `update_all_config` in the HBase shell.  The current active cluster is tracked by having the cluster's ID in a file called `active.cluster.suffix.id`.  This file does not exist when all clusters are replicas.

## Key Changes

1. Adds a `volatile boolean` flag called `readOnlyTransitionBlocked` to HBaseServerBase. This is used to track whether a cluster was prevented from becoming an active cluster due to another active cluster already existing.
    - If another active cluster exists when changing read-only from `false` to `true`, then `readOnlyTransitionBlocked` is set to `true`, and a new exception called `ReadOnlyTransitionException` is thrown in the HBase shell.
2. Adds a `static` method to `AbstractReadOnlyController` called `isAnotherClusterActive()`.  This returns true when the cluster ID in `active.cluster.suffix.id` does not match the current cluster.
3. Adds logic to `onConfigurationChange()` in `HMaster`, `HRegionServer`, and `HRegion` to prevent the replica cluster from being promoted to an active cluster when another active cluster exists.
    - These methods does not return early when the read-only transition is blocked in case other coprocessors or variables need to be dynamically loaded.
4. Read-only mode is still active and the ReadOnlyController coprocessors are still loaded if a cluster is blocked from becoming an active cluster.

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/ReadOnlyTransitionException.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/ReadOnlyTransitionException.java
@@ -28,8 +28,11 @@ public class ReadOnlyTransitionException extends DoNotRetryIOException {
 
   private static final long serialVersionUID = 1L;
 
+  private final String activeClusterId;
+
   public ReadOnlyTransitionException() {
     super();
+    this.activeClusterId = null;
   }
 
   /**
@@ -37,6 +40,16 @@ public class ReadOnlyTransitionException extends DoNotRetryIOException {
    */
   public ReadOnlyTransitionException(String message) {
     super(message);
+    this.activeClusterId = null;
+  }
+
+  /**
+   * @param message         the message for this exception
+   * @param activeClusterId the ID of the cluster that is currently active
+   */
+  public ReadOnlyTransitionException(String message, String activeClusterId) {
+    super(message);
+    this.activeClusterId = activeClusterId;
   }
 
   /**
@@ -45,6 +58,7 @@ public class ReadOnlyTransitionException extends DoNotRetryIOException {
    */
   public ReadOnlyTransitionException(String message, Throwable throwable) {
     super(message, throwable);
+    this.activeClusterId = null;
   }
 
   /**
@@ -52,5 +66,19 @@ public class ReadOnlyTransitionException extends DoNotRetryIOException {
    */
   public ReadOnlyTransitionException(Throwable throwable) {
     super(throwable);
+    this.activeClusterId = null;
+  }
+
+  /** Returns the ID of the cluster that is currently active, or null if unknown */
+  public String getActiveClusterId() {
+    return activeClusterId;
+  }
+
+  @Override
+  public String getMessage() {
+    if (activeClusterId != null) {
+      return super.getMessage() + " Current active cluster ID: " + activeClusterId;
+    }
+    return super.getMessage();
   }
 }
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/ReadOnlyTransitionException.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/ReadOnlyTransitionException.java
@@ -1,0 +1,56 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase;
+
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * Thrown when a replica cluster attempts to disable read-only mode while another active cluster
+ * already exists on the same storage location.
+ */
+@InterfaceAudience.Public
+public class ReadOnlyTransitionException extends DoNotRetryIOException {
+
+  private static final long serialVersionUID = 1L;
+
+  public ReadOnlyTransitionException() {
+    super();
+  }
+
+  /**
+   * @param message the message for this exception
+   */
+  public ReadOnlyTransitionException(String message) {
+    super(message);
+  }
+
+  /**
+   * @param message   the message for this exception
+   * @param throwable the {@link Throwable} to use for this exception
+   */
+  public ReadOnlyTransitionException(String message, Throwable throwable) {
+    super(message, throwable);
+  }
+
+  /**
+   * @param throwable the {@link Throwable} to use for this exception
+   */
+  public ReadOnlyTransitionException(Throwable throwable) {
+    super(throwable);
+  }
+}
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/HBaseServerBase.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/HBaseServerBase.java
@@ -667,7 +667,7 @@ public abstract class HBaseServerBase<R extends HBaseRpcServicesBase<?>> extends
       HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME, this.blockingActiveClusterId.get(),
       HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY);
     this.readOnlyTransitionBlocked.set(true);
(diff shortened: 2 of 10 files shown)


kgeisz:
These unit test failures are not relevant.

hbase-client/src/main/java/org/apache/hadoop/hbase/ReadOnlyTransitionException.java
```diff
@@ -0,0 +1,56 @@
...
+package org.apache.hadoop.hbase;
+
+import org.apache.yetus.audience.InterfaceAudience;
+
+/**
+ * Thrown when a replica cluster attempts to disable read-only mode while another active cluster
+ * already exists on the same storage location.
+ */
+@InterfaceAudience.Public
+public class ReadOnlyTransitionException extends DoNotRetryIOException {
```
anmolnar, line 27: nit: Could encapsulate a String field identifying the current active cluster's ID and emitted to log when the exception is converted to string.

hbase-server/src/main/java/org/apache/hadoop/hbase/util/ConfigurationUtil.java
```diff
@@ -116,8 +116,22 @@ public static List<Map.Entry<String, String>> getKeyValues(Configuration conf, S
...
   public static boolean isReadOnlyModeEnabledInConf(Configuration conf) {
     return conf.getBoolean(HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY,
       HConstants.HBASE_GLOBAL_READONLY_ENABLED_DEFAULT);
   }
+
+  /**
+   * Returns a copied version of the provided Configuration object that has
+   * {@link HConstants#HBASE_GLOBAL_READONLY_ENABLED_KEY} set to true.
+   */
+  public static Configuration getReadOnlyEnabledConfigurationCopy(Configuration conf) {
```
anmolnar: I think **set**ReadOnlyEnabledConfigurationCopy would be a better name for this method.

hbase-server/src/main/java/org/apache/hadoop/hbase/HBaseServerBase.java
```diff
@@ -102,15 +103,19 @@ public abstract class HBaseServerBase<R extends HBaseRpcServicesBase<?>> extends
...
 
   // Set when a report to the master comes back with a message asking us to
-  // shutdown. Also set by call to stop when debugging or running unit tests
+  // shut down. Also set by call to stop when debugging or running unit tests
   // of HRegionServer in isolation.
   protected volatile boolean stopped = false;
 
+  // Flag set when a read-only to read-write transition is blocked because another active cluster
+  // exists
+  protected volatile boolean readOnlyTransitionBlocked = false;
```
anmolnar: Probably not strictly necessary here, but I prefer to use `AtomicBoolean` for multi-threaded scenarios.

anmolnar (commented):
Only a few nitpicks from me.

hbase-server/src/main/java/org/apache/hadoop/hbase/util/ConfigurationUtil.java
kgeisz: IMO the name should use `get` instead of `set` since the method returns a modified copy of the configuration.  The original configuration is left untouched.

kgeisz:
I @anmolnar, I addressed two of your comments in a new commit.  Now, if the active cluster ID is included in the `ReadOnlyTransitionException`, then the message will look something like this:

```
org.apache.hadoop.hbase.ReadOnlyTransitionException: Cannot disable read-only mode because another active cluster already exists on this storage location. The read-only coprocessors have not been removed. Current active cluster ID: newClusterId:<blank>
```

Also, I added the active cluster ID to the error message:

```
2026-07-13T09:16:28,286 ERROR [main] hbase.HBaseServerBase(664): Cannot disable read-only mode. The active.cluster.suffix.id file contains a different cluster ID (newClusterId:<blank>), which means that cluster is already the active cluster. Reverting hbase.global.readonly.enabled to true
```

anmolnar (approved):
lgtm.

hbase-server/src/main/java/org/apache/hadoop/hbase/util/ConfigurationUtil.java
anmolnar: Okay, leave it as is.

kgeisz:
@anmolnar, the unit test failure is not relevant.  The failed test was `TestReplicationSource.testAbortTrueOnError`

hbase-server/src/main/java/org/apache/hadoop/hbase/HBaseServerBase.java
```diff
@@ -639,11 +650,35 @@ public void updateConfiguration() throws IOException {
...
   }
 
+  protected Configuration blockReadOnlyTransition(Configuration updatedConf,
+    String activeClusterId) {
+    this.blockingActiveClusterId.set(activeClusterId);
+    LOG.error(
+      "Cannot disable read-only mode. The {} file contains a different cluster ID ({}), which means "
+        + "that cluster is already the active cluster. Reverting {} to true",
+      HConstants.ACTIVE_CLUSTER_SUFFIX_FILE_NAME, this.blockingActiveClusterId.get(),
+      HConstants.HBASE_GLOBAL_READONLY_ENABLED_KEY);
```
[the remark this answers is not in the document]
taklwu: the checkstyle didn't fail, so, this is not related

hbase-server/src/main/java/org/apache/hadoop/hbase/security/access/AbstractReadOnlyController.java
```diff
@@ -68,6 +68,33 @@ public void start(CoprocessorEnvironment env) throws IOException {
...
+    ActiveClusterSuffix localClusterSuffix) {
+    try {
+      ActiveClusterSuffix fileData =
+        FSUtils.getClusterIdFile(fs, rootDir, new ActiveClusterSuffix.Parser());
+      if (fileData == null) {
+        return false;
+      }
+      return !localClusterSuffix.equals(fileData);
+    } catch (IOException e) {
+      LOG.error(
```
taklwu: nit: I'd suggest to use warning because you're returning normally without throwing an exception. 
```suggestion
      LOG.warn(
```

I also found in `manageActiveClusterIdFile` that using error message without throwing exceptions is the pattern......

hbase-server/src/main/java/org/apache/hadoop/hbase/util/ConfigurationUtil.java
taklwu: nit: or maybe use `copyWithReadOnlyModeEnabled` ?

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestReadOnlyManageActiveClusterFile.java
```diff
@@ -146,4 +166,176 @@ public void testDoNotDeleteActiveClusterIdFileWhenSwitchingToReadOnlyIfNotOwnedB
...
+
+    // Try to create a table. This should fail because the cluster is still in read-only mode
+    TableName tableName = TableName.valueOf("testBlockedTransitionTable");
+    TableDescriptor desc = TableDescriptorBuilder.newBuilder(tableName)
+      .setColumnFamily(ColumnFamilyDescriptorBuilder.of("cf")).build();
+    assertThrows(IOException.class, () -> TEST_UTIL.getAdmin().createTable(desc));
+
+    // Delete the foreign active cluster file
+    fs.delete(activeClusterFile, false);
+    assertFalse(activeClusterIdFileExists());
```
taklwu, line 315: nit: so we may need to add a documentation for recovering from blocked read-only mode, the user need to carefully clean up the incorrect activeClusterId by using either the filesystem API.

I knew this is strange, is there any hints to choose or tell which one should be the correct active cluster ID ?

taklwu (approved):
LGTM

hbase-server/src/test/java/org/apache/hadoop/hbase/security/access/TestReadOnlyManageActiveClusterFile.java
anmolnar: Whichever is already present in the active suffix file.

@kgeisz would you like to improve docs as part of this patch?

kgeisz: I can add documentation in this PR

kgeisz:
I have addressed @taklwu's comments.  I also added a new section to the docs that mentions how to recover from a `ReadOnlyTransitionException`.

image

taklwu (approved):

PR #8382, opened 2026-06-19
HBASE-30137 In FSFT, generate a new manifest incase the latest manifest file gets corrupted

gvprathyusha6:


dev-support/design-docs/fsft-manifest-recover.md
```diff
@@ -0,0 +1,268 @@
...
+For FILE SFT, the manifest can in principle reference store members that do not exist as plain
+files in the child family directory:
+
+- plain HFiles do exist on disk
+- virtual split/merge `Reference`s may exist only in the manifest
+- virtual `HFileLink`s may exist only in the manifest plus archive back references
+
+This design adds a single, offline, operator-driven recovery tool that rebuilds a corrupted
+manifest **purely from the store directory listing**, plus a non-authoritative data-loss assessment
+derived from `hbase:meta` split/merge lineage.
```
wchevreuil, line 18: So links/references existing only on the sft file would be lost?

wchevreuil (commented):
Thanks for submitting this. Could we have this moved to hbase-operator-tools repository? I think it would be more consistent with other fixing tooling if it could either be on hbck2 or in the hbase-tools module.

dev-support/design-docs/fsft-manifest-recover.md
gvprathyusha6: Yes, but incase the archival of parents is done by process links/references are already compacted so no data loss in that scenario, else there is scope of data loss similar to a corrupted HFile

PR #8385, opened 2026-06-20
HBASE-30239 Use hamcrest instead of deprecated hamcrest-core and hamc…

Apache9:
…rest-library


liuxiaocs7 (approved):

Apache9:
There are no failed UTs, just flaky ones.

Let me merge.

PR #8386, opened 2026-06-20
HBASE-30240 TestPrefetchPersistence fails because of a file is not fully cached

Apache9:


Apache9:
I also changed some assertions in the test which depends on HBASE-30239( #8385 ), let's get #8385 in first.

I've tried several time locally, the test does not fail any more.

wchevreuil (approved):
That's pretty much possible. LGTM.

PR #8387, opened 2026-06-21
HBASE-30241 TestMobCloneSnapshotFromClientAfterSplittingRegion is fla…

liuxiaocs7:
…ky because a split may build whole-file HFileLinks instead of References

- see: HBASE-30241
- Co-Authored-By: Claude Opus 4.8

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionTestBase.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionTestBase.java
@@ -28,7 +28,6 @@ import org.apache.hadoop.hbase.MetaTableAccessor;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.master.RegionState;
 import org.apache.hadoop.hbase.master.assignment.RegionStates;
-import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.snapshot.SnapshotTestingUtils;
 import org.apache.hadoop.hbase.util.EnvironmentEdgeManager;
 import org.junit.jupiter.api.TestTemplate;
@@ -49,14 +48,10 @@ public class CloneSnapshotFromClientAfterSplittingRegionTestBase
     // point). Without this, the randomly generated row keys may all fall on one side of the split
     // point, leaving the snapshot with no reference files. The cloned table's meta would then be
     // missing the parent split information, which is what HBASE-29111 guards against below. The
-    // region server has compaction disabled (see CloneSnapshotFromClientTestBase), so we compact
-    // the regions directly to bypass that; auto-compaction stays disabled afterwards so the
-    // post-split reference files are not compacted away.
-    for (HRegion region : TEST_UTIL.getHBaseCluster().getRegions(tableName)) {
-      if (RegionReplicaUtil.isDefaultReplica(region.getRegionInfo())) {
-        region.compact(true);
-      }
-    }
+    // region server has compaction disabled (see CloneSnapshotFromClientTestBase), so we
+    // major-compact the table's regions directly to bypass that; auto-compaction stays disabled
+    // afterward so the post-split reference files are not compacted away.
+    TEST_UTIL.compact(tableName, true);
     try (Table k = TEST_UTIL.getConnection().getTable(tableName);
       ResultScanner scanner = k.getScanner(new Scan())) {
       // Split on the second row to make sure that the snapshot contains reference files.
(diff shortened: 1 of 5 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
```diff
@@ -0,0 +1,275 @@
...
+ * <p>
+ * Since HBASE-26421, a split builds an {@code HFileLink} (not a {@code Reference}) for a daughter
+ * whenever a store file lies entirely on one side of the split point. When every store file falls
+ * on one side, the snapshot contains no reference files at all, so the daughters link directly to
+ * the snapshot files and do not depend on the cloned parent region. This is the complement of the
+ * {@code Reference} case verified by {@code CloneSnapshotFromClientAfterSplittingRegionTestBase}
+ * (the regression HBASE-29111 guards against): here there is no parent-to-daughter mapping to
+ * record, and the cloned table must still be safe after the source table and snapshot are removed.
+ */
+public class CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase {
```
Apache9: We do not extend CloneSnapshotFromClientTestBase here?

liuxiaocs7: Thanks for the comments, done!

hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
```diff
@@ -0,0 +1,238 @@
...
+    TEST_UTIL.getConfiguration().setLong(TimeToLiveHFileCleaner.TTL_CONF_KEY, 0);
+  }
+
+  /**
+   * Unlike the parent, build a controlled single-region table with two disjoint store files (in the
+   * test itself) rather than the default randomly-populated, pre-split table and snapshots, so the
+   * split deterministically produces whole-file HFileLinks. Hence we do not call {@code super}.
+   */
+  @Override
+  @BeforeEach
```
Apache9: Is it possible to not override the whole beforeEach method? I prefer we call some sub methods in the setUp method, and we can override some of these methods here to change the behavior.

hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
```diff
@@ -0,0 +1,238 @@
...
+    clonedTableName = TableName.valueOf(methodName + "-clone-" + tid);
+    snapshotName = "snaptb-links-" + tid;
+    // Keep the split parent around so that the snapshot (and therefore the clone) contains it.
+    admin.catalogJanitorSwitch(false);
+  }
+
+  @Override
+  @AfterEach
+  public void tearDown() throws Exception {
+    admin.catalogJanitorSwitch(true);
```
Apache9, line 101: The only differentce is here? Then I think we can introduce a separated AfterEach method here, in JUnit5, the framework will call all AfterEach methods in both parent and sub classes.

hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
liuxiaocs7: Done!

hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
liuxiaocs7: Done!

hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionTestBase.java
```diff
@@ -41,10 +41,20 @@ protected CloneSnapshotFromClientAfterSplittingRegionTestBase(int numReplicas) {
...
+    // region. This guarantees that the split row falls inside an existing store file, so the split
+    // produces a reference file rather than a whole-file link (see HBASE-26421, which builds an
+    // HFileLink instead of a Reference when a store file lies entirely on one side of the split
+    // point). Without this, the randomly generated row keys may all fall on one side of the split
+    // point, leaving the snapshot with no reference files. The cloned table's meta would then be
+    // missing the parent split information, which is what HBASE-29111 guards against below. The
+    // region server has compaction disabled (see CloneSnapshotFromClientTestBase), so we
+    // major-compact the table's regions directly to bypass that; auto-compaction stays disabled
+    // afterward so the post-split reference files are not compacted away.
+    TEST_UTIL.compact(tableName, true);
```
liuxiaocs7, line 54: Here is the fix, we need to compact store files here

Apache9 (approved):

PR #8389, opened 2026-06-21
HBASE-30242 Add Region server level HFile Compression Metrics

Umeshkumar9414:
Local testing 
image

### DIFF
--- a/hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerSource.java
+++ b/hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerSource.java
@@ -233,6 +233,8 @@ public interface MetricsRegionServerSource extends BaseSource, JvmPauseMonitorSo
   String MEMSTORE_OFFHEAP_SIZE_DESC = "Off-heap Size of the memstore";
   String STOREFILE_SIZE = "storeFileSize";
   String STOREFILE_SIZE_GROWTH_RATE = "storeFileSizeGrowthRate";
+  String STOREFILE_UNCOMPRESSED_SIZE = "storeFileUncompressedSize";
+  String STOREFILE_COMPRESSION_RATIO = "storeFileCompressionRatio";
   String MAX_STORE_FILE_AGE = "maxStoreFileAge";
   String MIN_STORE_FILE_AGE = "minStoreFileAge";
   String AVG_STORE_FILE_AGE = "avgStoreFileAge";
@@ -244,6 +246,10 @@ public interface MetricsRegionServerSource extends BaseSource, JvmPauseMonitorSo
   String STOREFILE_SIZE_DESC = "Size of storefiles being served.";
   String STOREFILE_SIZE_GROWTH_RATE_DESC =
     "Bytes per second by which the size of storefiles being served grows.";
+  String STOREFILE_UNCOMPRESSED_SIZE_DESC = "Total uncompressed size of storefiles being served.";
+  String STOREFILE_COMPRESSION_RATIO_DESC =
+    "Compression ratio of storefiles (compressed/uncompressed). Lower values indicate better"
+      + " compression. Returns 1.0 when there is no data.";
   String TOTAL_REQUEST_COUNT = "totalRequestCount";
   String TOTAL_REQUEST_COUNT_DESC =
     "Total number of requests this RegionServer has answered; increments the count once for "
--- a/hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerSource.java
+++ b/hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerSource.java
@@ -248,8 +248,8 @@ public interface MetricsRegionServerSource extends BaseSource, JvmPauseMonitorSo
     "Bytes per second by which the size of storefiles being served grows.";
   String STOREFILE_UNCOMPRESSED_SIZE_DESC = "Total uncompressed size of storefiles being served.";
   String STOREFILE_COMPRESSION_RATIO_DESC =
-    "Compression ratio of storefiles (compressed/uncompressed). Lower values indicate better"
-      + " compression. Returns 1.0 when there is no data.";
+    "Compression ratio of storefiles (uncompressed/compressed). Higher values indicate better"
+      + " compression. Returns 0.0 when there is no data.";
   String TOTAL_REQUEST_COUNT = "totalRequestCount";
   String TOTAL_REQUEST_COUNT_DESC =
     "Total number of requests this RegionServer has answered; increments the count once for "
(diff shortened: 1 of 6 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
```diff
@@ -533,6 +533,20 @@ public long getStoreFileSize() {
     return aggregate.storeFileSize;
   }
 
+  @Override
+  public long getStoreFileUncompressedSize() {
+    return aggregate.storeFileUncompressedSize;
+  }
+
+  @Override
+  public double getStoreFileCompressionRatio() {
```
apurtell, line 542: Generally for compression factor, higher is better. This is what command line compression tools like `xz` or `zstd` or etc. typically report. And for compression savings, higher is also better, e.g `1 − compressed / uncompressed`, which is easy to read as a percentage. This method provides the inverse of that so it makes the desired direction "down", which is awkward for Prometheus or Grafana dashboards. Operators reading a dashboard tile labeled `storeFileCompressionRatio` with a value e.g. of `0.33` will routinely interpret it the opposite of what is intended.

Recommendation: either rename to `storeFileCompressedToUncompressedRatio` so the direction is clearly established by the name of the metric, or reimplement as `uncompressed / compressed`, or `storeFileCompressionSavingsRatio` as (`1 − compressed/uncompressed`).

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
```diff
@@ -533,6 +533,20 @@ public long getStoreFileSize() {
...
+    long uncompressed = aggregate.storeFileUncompressedSize;
+    if (uncompressed == 0) {
```
apurtell, line 544: A value of "1.0" will generally look identical to an operator like somehow compression has stopped working, and will generate false positive alerts.

(Claude calls this a "footgun for dashboards", lol. He means well...)

Recommendation: Return 0.0 instead, and update the description to something like "Returns 0.0 when there is no data."

apurtell (changes requested):

hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerSource.java
```diff
@@ -244,6 +246,10 @@ public interface MetricsRegionServerSource extends BaseSource, JvmPauseMonitorSo
   String STOREFILE_SIZE_DESC = "Size of storefiles being served.";
   String STOREFILE_SIZE_GROWTH_RATE_DESC =
     "Bytes per second by which the size of storefiles being served grows.";
+  String STOREFILE_UNCOMPRESSED_SIZE_DESC = "Total uncompressed size of storefiles being served.";
+  String STOREFILE_COMPRESSION_RATIO_DESC =
+    "Compression ratio of storefiles (compressed/uncompressed). Lower values indicate better"
+      + " compression. Returns 1.0 when there is no data.";
```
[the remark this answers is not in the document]
Umeshkumar9414: changed the way of ratio calculation. Now higher is better.

hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapper.java
```diff
@@ -116,6 +116,17 @@ public interface MetricsRegionServerWrapper {
...
+  /**
+   * Get the total uncompressed size of the store files this region server is serving from.
+   */
+  long getStoreFileUncompressedSize();
+
+  /**
+   * Get the compression ratio of store files on this region server. This is the ratio of compressed
+   * on-disk size to uncompressed data size. Returns 1.0 when there is no data.
+   */
+  double getStoreFileCompressionRatio();
```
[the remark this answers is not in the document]
Umeshkumar9414: changes the ratio calculation way. Now higher is better.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
Umeshkumar9414: reimplemented as uncompressed / compressed as that is more common.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
Umeshkumar9414: done. Thanks

apurtell (commented):
I think the latest copilot suggestions are good

1) MetricsRegionServerWrapperImpl.java:547 `aggregate` is volatile and replaced by the recomputation thread, but this method reads it twice. A swap between those reads can combine the uncompressed size from one snapshot with the compressed size from another, producing an incorrect ratio (including infinity when regions are removed). Capture one `aggregate` snapshot and use it for both operands.

2) MetricsRegionServerWrapperImpl.java:997 The added source test uses a stub that returns the ratio directly, so it does not exercise this aggregation or the production ratio calculation. Extend TestMetricsRegionServerAggregate, which already validates getStoreFileSize(), to assert the aggregated uncompressed size and computed ratio (including the zero-data case).

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperImpl.java
```diff
@@ -533,6 +533,20 @@ public long getStoreFileSize() {
...
   }
 
+  @Override
+  public long getStoreFileUncompressedSize() {
+    return aggregate.storeFileUncompressedSize;
+  }
+
+  @Override
+  public double getStoreFileCompressionRatio() {
+    long uncompressed = aggregate.storeFileUncompressedSize;
```
mnpoonia, line 543: `aggregate` is a volatile snapshot as andrew mentioned and can be replaced between these two reads. That can calculate the ratio from different refreshes, including `nonzero / 0` when regions are removed.  Could we capture `aggregate` once, read both values from that snapshot?

```
RegionMetricAggregate current = aggregate;
if (current.storeFileSize <= 0) {
  return 0.0;
}
return (double) current.storeFileUncompressedSize / current.storeFileSize;
```

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/MetricsRegionServerWrapperStub.java
```diff
@@ -87,6 +87,16 @@ public long getStoreFileSize() {
     return 1900;
   }
 
+  @Override
```
mnpoonia, line 90: This verifies source export from a fixed wrapper stub, but it does not exercise `MetricsRegionServerWrapperImpl` aggregation or ratio calculation. Could we extend `TestMetricsRegionServerAggregate` to assert the summed compressed and uncompressed sizes, along with the resulting `uncompressed / compressed` ratio, including the empty-aggregate case?

Using stores with different individual ratios would also verify that this remains a weighted aggregate rather than an average of per-store ratios.

Basically +1 to Andrew's comment.

PR #8392, opened 2026-06-22
HBASE-30246 MOB compaction does not close MobCell after resolving ref…

liuxiaocs7:
…erence cells

- see: HBASE-30246

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/mob/DefaultMobStoreCompactor.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/mob/DefaultMobStoreCompactor.java
@@ -37,6 +37,7 @@ import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.DoNotRetryIOException;
 import org.apache.hadoop.hbase.ExtendedCell;
 import org.apache.hadoop.hbase.KeyValue;
+import org.apache.hadoop.hbase.KeyValueUtil;
 import org.apache.hadoop.hbase.PrivateCellUtil;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.regionserver.CellSink;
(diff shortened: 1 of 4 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/mob/DefaultMobStoreCompactor.java
```diff
@@ -174,6 +175,29 @@ public DefaultMobStoreCompactor(Configuration conf, HStore store) {
...
+   * The {@link KeyValueUtil#copyToNewKeyValue(ExtendedCell)} call is required by this ownership
+   * model: HFile writers and encoders may retain references to appended cells (e.g.
+   * {@code lastCell}, {@code firstCellInBlock}, and the data block encoder's {@code prevCell})
+   * until {@code beforeShipped()}. Returning the scanner-backed cell directly would let those later
+   * reads access released buffers. Removing this copy would require changing the caller to retain
+   * each {@link MobCell} and close it only after the writers have shipped their retained
+   * references.
+   */
+  protected ExtendedCell resolveMobCell(ExtendedCell reference) throws IOException {
+    try (MobCell mobCell = mobStore.resolve(reference, cacheMobBlocksOnCompaction, false)) {
```
guluo2016, line 196: Maybe `protected ExtendedCell resolveMobCell(ExtendedCell reference, boolean flag)` is better？

I think we can also reuse this method in the future if we need to call `mobStore.resolve(reference, cacheMobBlocksOnCompaction, true)` by this way

liuxiaocs7: Thanks @guluo2016 ! The false (readEmptyValueOnMobCellMiss) is coupled to this path's miss handling (it relies on resolve throwing, while true returns an empty cell needing different handling), so it can't be reused as-is. I'd prefer to add the param with a proper name once a real true caller appears, WDYT?

guluo2016: Agreed, thanks.

guluo2016 (approved):

liuxiaocs7:
Thanks @guluo2016 for reviewing!

PR #8396, opened 2026-06-23
HBASE-30195 Update Thrift gateway to use new hbase-shaded-thrift in thirdparty

apurtell:
On https://github.com/apache/hbase-thirdparty/pull/160 we onboard `libthrift` to `hbase-thirdparty` so we can consume libthrift security fixes while maintaining Java 8 compatibility for HBase branch-2, branch-2.5, and branch-2.6. 

Use the new `hbase-shaded-thrift` in `hbase-thirdparty`. 

`branch-2` companion to #8299


apurtell:
Let me see what happened here locally and solve this before merging the `master` PR.

apurtell:
Ok, rebased on `branch-2` and done properly.

apurtell:
Looking better. There is a flaky test, `TestAsyncProcess`, that is not related to this change but needs follow up. Thrift tests pass.

apurtell:
Carrying over signoffs from `master` version. The only difference here is generation of Thrift gen-sources from the IDL that ships in `branch-2`.

PR #8397, opened 2026-06-23
HBASE-30240 TestPrefetchPersistence fails because of a file is not fu…

Apache9:
…lly cached (#8386)

(cherry picked from commit e2218e0b589bbd9f0a82a8ceeae59fc16f746a6f)


Apache9:
The failed UT is TestSaslTlsIPC, which is not related to the changes here.

Let me merge.

PR #5947, opened 2024-05-26
HBASE-28614 Introduce a field to display whether the snapshot is expired

guluo2016:
Details see: [HBASE-28614](https://issues.apache.org/jira/browse/HBASE-28614)

After applying this commit.

On HBase UI:  
image

image

image

On hbase shell:  
```
hbase:027:0> list_snapshots
SNAPSHOT                                    TABLE + CREATION TIME + TTL(Sec)
 expired_snap01                             t01 (2024-05-26 14:18:16 +0800) 10 (Expired)
 snap01                                     t01 (2024-05-25 22:28:13 +0800) FOREVER
 snap02                                     t01 (2024-05-25 22:28:37 +0800) 100000
3 row(s)
Took 0.0209 seconds
=> ["expired_snap01", "snap01", "snap02"]
```

PR #8403, opened 2026-06-23
HBASE-30252 `hbase snapshot info` display inaccurate snapshot TTL information messages

guluo2016:
Details see: [HBASE-30252](https://issues.apache.org/jira/browse/HBASE-30252)

In addition, I also added information about whether the snapshot has expired in the display.  like PR https://github.com/apache/hbase/pull/5947

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
@@ -17,6 +17,8 @@
  */
 package org.apache.hadoop.hbase.snapshot;
 
+import static org.apache.hadoop.hbase.HConstants.DEFAULT_SNAPSHOT_TTL;
+
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.net.URI;
@@ -70,6 +72,10 @@ import org.apache.hadoop.hbase.shaded.protobuf.generated.SnapshotProtos.Snapshot
 @InterfaceAudience.Public
 public final class SnapshotInfo extends AbstractHBaseTool {
   private static final Logger LOG = LoggerFactory.getLogger(SnapshotInfo.class);
+
+  private static final String EXPIRED_LABEL = "Yes";
+  private static final String NOT_EXPIRED_LABEL = "No";
+  private static final String TTL_FOREVER = "FOREVER";
 
   static final class Options {
     static final Option SNAPSHOT =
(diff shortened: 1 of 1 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
```diff
@@ -391,11 +397,17 @@ public int doWork() throws IOException, InterruptedException {
     // List Available Snapshots
     if (listSnapshots) {
       SimpleDateFormat df = new SimpleDateFormat("yyyy-MM-dd'T'HH:mm:ss");
-      System.out.printf("%-20s | %-20s | %-20s | %s%n", "SNAPSHOT", "CREATION TIME", "TTL IN SEC",
-        "TABLE NAME");
-      for (SnapshotDescription desc : getSnapshotList(conf)) {
-        System.out.printf("%-20s | %20s | %20s | %s%n", desc.getName(),
-          df.format(new Date(desc.getCreationTime())), desc.getTtl(), desc.getTableNameAsString());
+      List<SnapshotDescription> snapshotDescriptions = getSnapshotList(conf);
```
guluo2016, line 400: I update the code here, moved this code before `System.out.printf(...);`
The reason is that `getSnapshotList(conf)` would print logs, breaking the display continuity of the snapshot list.

hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
```diff
@@ -391,11 +397,17 @@ public int doWork() throws IOException, InterruptedException {
...
+      System.out.printf("%-20s | %-20s | %-20s | %-20s | %s%n", "SNAPSHOT", "CREATION TIME",
+        "TTL(Sec)", "EXPIRED", "TABLE NAME");
+      for (SnapshotDescription desc : snapshotDescriptions) {
+        String ttlInfo =
+          desc.getTtl() == DEFAULT_SNAPSHOT_TTL ? TTL_FOREVER : String.valueOf(desc.getTtl());
+        String expired = SnapshotDescriptionUtils.isExpiredSnapshot(desc.getTtl(),
+          desc.getCreationTime(), System.currentTimeMillis()) ? EXPIRED_LABEL : NOT_EXPIRED_LABEL;
```
liuxiaocs7: nit: do we need to move `System.currentTimeMillis()` outside the loop?

liuxiaocs7 (approved):

PDavid (approved):
Nice catch, thanks for fixing and the improvement. 👍

hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotInfo.java
guluo2016: Good catch. Let me move it outside the loop, thanks

PDavid:
I guess we cannot add a unit test for this change as this class writes to stdout. It would require refactor to allow testing. :thinking:

guluo2016:
> I guess we cannot add a unit test for this change as this class writes to stdout. It would require refactor to allow testing. 🤔

Yes, you're right. The` SnapshotInfo`  directly writes to stdout. I've verified the patch manually, and the output is as expected: 
```bash
[root@localhost hbase-4.0.0-alpha-1-SNAPSHOT]# ./bin/hbase snapshot info --list-snapshots
2026-06-25T06:23:55,233 WARN  [main] util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2026-06-25T06:23:55,520 INFO  [main] hbase.TableName: Meta table name: hbase:meta
SNAPSHOT             | CREATION TIME        | TTL(Sec)             | EXPIRED              | TABLE NAME
t01_snap01           |  2026-06-25T06:22:10 |              FOREVER |                   No | t01
t01_snap02           |  2026-06-25T06:22:34 |               100000 |                   No | t01
t01_snap03           |  2026-06-25T06:22:44 |                   10 |                  Yes | t01
[root@localhost hbase-4.0.0-alpha-1-SNAPSHOT]#
[root@localhost hbase-4.0.0-alpha-1-SNAPSHOT]# ./bin/hbase snapshot info --snapshot t01_snap03
2026-06-25T06:24:16,442 WARN  [main] util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2026-06-25T06:24:16,736 INFO  [main] hbase.TableName: Meta table name: hbase:meta
Snapshot Info
-----------------------------------------
    Name: t01_snap03
    Type: FLUSH
   Table: t01
  Format: 2
 Created: 2026-06-25T06:22:44
Ttl(Sec): 10
 Expired: Yes
   Owner:


[root@localhost hbase-4.0.0-alpha-1-SNAPSHOT]# ./bin/hbase snapshot info --snapshot t01_snap02
2026-06-25T06:24:23,404 WARN  [main] util.NativeCodeLoader: Unable to load native-hadoop library for your platform... using builtin-java classes where applicable
2026-06-25T06:24:23,691 INFO  [main] hbase.TableName: Meta table name: hbase:meta
Snapshot Info
-----------------------------------------
    Name: t01_snap02
    Type: FLUSH
   Table: t01
  Format: 2
 Created: 2026-06-25T06:22:34
Ttl(Sec): 100000
 Expired: No
   Owner:

```

guluo2016:
Thanks for reviewing @liuxiaocs7 @PDavid 
Merging it now

PR #8404, opened 2026-06-24
HBASE-30253 Do not add/remove peer between tests in TestBulkLoadRepli…

Apache9:
…cation


hbase-server/src/test/java/org/apache/hadoop/hbase/replication/TestBulkLoadReplication.java
```diff
@@ -162,13 +162,19 @@ private static void startThirdCluster() throws Exception {
...
+    // Removing the peer and adding it back causes previously completed bulk-load jobs to be
+    // resubmitted. Override setUpBase/tearDownBase so we do not add/remove peers between tests;
+    // peers are added once in @BeforeAll.
     BULK_LOADS_COUNT = new AtomicInteger(0);
   }
 
+  @Override
+  public void tearDownBase() throws Exception {
+    // do not remove PEER_ID2
+  }
```
liuxiaocs7, line 178: We’d better add an `@AfterEach` annotation to maintain consistency here, even though it’s equivalent?

liuxiaocs7 (approved):

PR #8412, opened 2026-06-24
HBASE-30256 FuzzyRowFilter mishandles the fuzzy-mask encoding on the …

liuxiaocs7:
…no-unsafe path

- see: HBASE-30256

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFuzzyRowFilter.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFuzzyRowFilter.java
@@ -405,7 +405,7 @@ public class TestFuzzyRowFilter {
     original.filterCell(KeyValueUtil.createFirstOnRow(new byte[] { 1, 50, 3 }));
 
     FuzzyRowFilter parsed = FuzzyRowFilter.parseFrom(original.toByteArray());
-    // A row matching only via the wildcard position must still be INCLUDEEd after the round-trip.
+    // A row matching only via the wildcard position must still be INCLUDED after the round-trip.
     assertEquals(Filter.ReturnCode.INCLUDE,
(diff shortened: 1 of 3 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFuzzyRowFilter.java
```diff
@@ -391,6 +391,43 @@ public void testReverseFilterListSkipsSameRowFuzzyHint() throws IOException {
...
+   */
+  @Test
+  public void testSerializationAfterFilterCellPreservesBehavior() throws Exception {
+    FuzzyRowFilter original =
+      new FuzzyRowFilter(Arrays.asList(new Pair<>(new byte[] { 1, 2, 3 }, new byte[] { 0, 1, 0 })));
+    // Process a cell first to prove serialization is unaffected by scanning.
+    original.filterCell(KeyValueUtil.createFirstOnRow(new byte[] { 1, 50, 3 }));
+
+    FuzzyRowFilter parsed = FuzzyRowFilter.parseFrom(original.toByteArray());
+    // A row matching only via the wildcard position must still be INCLUDEEd after the round-trip.
```
PDavid: Super-minor / nit: typo:

```suggestion
    // A row matching only via the wildcard position must still be INCLUDED after the round-trip.
```

PDavid (approved):
Nice catch and many thanks for fixing this. 👍

hbase-server/src/test/java/org/apache/hadoop/hbase/filter/TestFuzzyRowFilter.java
liuxiaocs7: Thanks @PDavid, committed!

PDavid (approved):
Thanks! 👍

PR #8420, opened 2026-06-26
HBASE-30241 Addendum Fix flaky testClonedTableWithLinksSurvivesSource…

liuxiaocs7:
…Deletion

`testClonedTableWithLinksSurvivesSourceDeletion` fails on branch-2, introduce an addendum to forwardport master and branch-3

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
@@ -121,12 +121,13 @@ public class CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase
     // so the snapshot contains only HFileLinks and no Reference files.
     int numRegions = admin.getRegions(tableName).size();
     admin.split(tableName, SPLIT_KEY);
-    await().atMost(Duration.ofSeconds(60)).untilAsserted(
+    await().atMost(Duration.ofSeconds(30)).untilAsserted(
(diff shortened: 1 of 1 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/client/CloneSnapshotFromClientAfterSplittingRegionWithLinksTestBase.java
```diff
@@ -121,12 +121,13 @@ public void testClonedTableWithLinksSurvivesSourceDeletion() throws Exception {
     // so the snapshot contains only HFileLinks and no Reference files.
     int numRegions = admin.getRegions(tableName).size();
     admin.split(tableName, SPLIT_KEY);
-    await().atMost(Duration.ofSeconds(60)).untilAsserted(
+    await().atMost(Duration.ofSeconds(30)).untilAsserted(
```
guluo2016, line 124: Why was this changed from 60 to 30?

liuxiaocs7: Hi, @guluo2016, thanks for the comments.

The `60s` was just an arbitrary upper bound from the original HBASE-30241 fix. I lowered it to `30s` to match the analogous `CloneSnapshotFromClientAfterSplittingRegionTestBase`, which uses the same `await().atMost(30s)` for the identical split-completion check. `await()` returns as soon as the condition holds, so it's only a ceiling.

It's unrelated to the actual flaky fix though, unify the code in master and branch-3 with the code in branch-2.x

guluo2016 (approved):
Thanks!

PR #8427, opened 2026-06-28
HBASE-29364 Fix FAILED_OPEN state during master restart

SwaraliJoshi:
https://issues.apache.org/jira/browse/HBASE-29364

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedure.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedure.java
@@ -1,0 +1,89 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.master.assignment;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import org.apache.hadoop.hbase.ServerName;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.RegionInfo;
+import org.apache.hadoop.hbase.master.RegionState.State;
+import org.apache.hadoop.hbase.master.procedure.MasterProcedureEnv;
+import org.apache.hadoop.hbase.testclassification.MasterTests;
+import org.apache.hadoop.hbase.testclassification.MediumTests;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import org.apache.hadoop.hbase.shaded.protobuf.generated.RegionServerStatusProtos.RegionStateTransition.TransitionCode;
+
+/**
+ * Coverage for {@link OpenRegionProcedure#restoreSucceedState}, which is invoked while loading
+ * hbase:meta on master failover to re-apply a transition that was reported by the RegionServer but
+ * not yet persisted before the old master went away.
+ * <p>
+ * Regression test for HBASE-29364: a recorded {@code FAILED_OPEN} report must NOT be restored as
+ * {@code OPEN} (doing so could leave the region marked open on a now-dead server); it must be left
+ * in transition for the parent {@link TransitRegionStateProcedure} to retry. A recorded
+ * {@code OPENED} must still be restored as {@code OPEN}.
+ */
+@Tag(MasterTests.TAG)
+@Tag(MediumTests.TAG)
+public class TestOpenRegionProcedure extends TestAssignmentManagerBase {
+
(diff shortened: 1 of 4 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedure.java
```diff
@@ -0,0 +1,89 @@
...
+ * not yet persisted before the old master went away.
+ * <p>
+ * Regression test for HBASE-29364: a recorded {@code FAILED_OPEN} report must NOT be restored as
+ * {@code OPEN} (doing so could leave the region marked open on a now-dead server); it must be left
+ * in transition for the parent {@link TransitRegionStateProcedure} to retry. A recorded
+ * {@code OPENED} must still be restored as {@code OPEN}.
+ */
+@Tag(MasterTests.TAG)
+@Tag(MediumTests.TAG)
+public class TestOpenRegionProcedure extends TestAssignmentManagerBase {
```
Apache9, line 46: Is it possible to reproduce the problem in a real mini cluster with a real TRSP?
We need to confirm that this could happen in real production.

SwaraliJoshi: Yes, I did have a test to confirm this issue before fixing it. Here is the code : 
1. Mini Cluster test - 
[TestFailedOpenRestoredAsOpenOnFailover.java](https://github.com/user-attachments/files/29453546/TestFailedOpenRestoredAsOpenOnFailover.java)

2. Replication of exact bug manually - 
[TestOpenRegionProcedureRestoreFailedOpen.java](https://github.com/user-attachments/files/29453564/TestOpenRegionProcedureRestoreFailedOpen.java)

These two assert the pre-fix (buggy) behavior — that a FAILED_OPEN ends up OPEN. With the fix currently in place, they will fail. They're bug-demonstration artifacts, not regression guards.

Apache9: Let's include them in this PR too.

SwaraliJoshi: But these tests will fail. Or should I instead have `assertEquals(State.OPENING, regionState(newMaster, hri));` ? 

Update : I have added the test to assert the state is marked as OPENING

virajjasani:
Build failure:
```
Banned imports detected:
Reason: Use junit5 instead
	in file: org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
		static org.junit.Assert.assertNotEquals (Line: 20, Matched by: static org.junit.Assert.**)
```

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
```diff
@@ -0,0 +1,245 @@
...
+    // Wait until the reassignment open has failed and the procedure is frozen in REPORT_SUCCEED
+    // (its post-FAILED_OPEN persistToMeta is blocked in the meta coprocessor).
+    LOG.info("Waiting for the FAILED_OPEN procedure to freeze in REPORT_SUCCEED...");
+    org.junit.jupiter.api.Assertions.assertTrue(
+      FailOpenAndBlockMetaCP.reachedBlock.await(120, TimeUnit.SECONDS),
+      "Timed out waiting for the FAILED_OPEN persistToMeta to be reached");
+
+    // The region must NOT be OPEN at this point - the open failed.
+    State beforeFailover = regionState(cluster.getMaster(), hri);
+    LOG.info("State before failover (active master): {}", beforeFailover);
```
virajjasani, line 143: Instead of this log, we can assert the region state with expected value

SwaraliJoshi: @virajjasani Updated sir, could you please check now?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
```diff
@@ -0,0 +1,247 @@
...
+    assertEquals(State.OPENING, regionState(cluster.getMaster(), hri));
+    LOG.info("State before failover (active master): {}", beforeFailover);
+
+    // The active Master "dies"; a backup Master takes over and replays procedures during meta-load.
+    LOG.info("Triggering real master failover...");
+    MasterProcedureTestingUtility.masterFailover(UTIL);
+    HMaster newMaster = cluster.getMaster();
+    LOG.info("New active master: {}", newMaster.getServerName());
+
+    UTIL.waitFor(60_000, 500, () -> regionState(newMaster, hri) == State.OPENING);
```
Apache9, line 151: I think the real harm here is that, the region is recorded as OPEN on the target regionserver but actually it is not? So the end to end test should make sure that the region is finally online and we can read/write it? Or we get the region location in meta, and then check with the region server directly about whether the region is on it.

SwaraliJoshi: This is an inconsistent state issue, where the region is marked as OPEN on hbase:meta while we are still trying to open it (and it failed). Our responsibility with this JIRA is to make sure that the state is correctly gauged so that TRSP can retry the assignment, which may or may not be successful. So we should not be checking if the region is finally online, we should just make sure that it is in OPENING state. 

Similarly, the meta already has the record as OPEN for this region on this particular region server, so re-checking region location to verify the region server is redundant as we are already aware it is incorrect state.

SwaraliJoshi:
@Apache9 Could you please review this PR?

PR #8432, opened 2026-06-29
HBASE-30261 Hide segmented CodeHeap memory pools on process info JSP …

liuxiaocs7:
…pages for JDK 9+

- see: HBASE-30261


PDavid (approved):
Thanks, looks good to me. 👍

PR #8442, opened 2026-06-30
HBASE-30266 Do not drop visibility delete markers as redundant during…

liuxiaocs7:
… minor compaction

- see: HBASE-30266


junegunn (approved):
Thanks!

PR #8456, opened 2026-07-06
HBASE-30275 TestRecoveryPersistentBucketCache#testBucketCacheRecovery…

mnpoonia:
…WithAllocationInconsistencies has a race condition between backingMapValidated and cacheState=ENABLED

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/bucket/TestRecoveryPersistentBucketCache.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/bucket/TestRecoveryPersistentBucketCache.java
@@ -22,10 +22,12 @@ import static org.apache.hadoop.hbase.io.hfile.bucket.BucketCache.ACCEPT_FACTOR_
 import static org.apache.hadoop.hbase.io.hfile.bucket.BucketCache.DEFAULT_ERROR_TOLERATION_DURATION;
 import static org.apache.hadoop.hbase.io.hfile.bucket.BucketCache.EXTRA_FREE_FACTOR_CONFIG_NAME;
 import static org.apache.hadoop.hbase.io.hfile.bucket.BucketCache.MIN_FACTOR_CONFIG_NAME;
+import static org.awaitility.Awaitility.await;
 import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertNull;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.time.Duration;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.HBaseConfiguration;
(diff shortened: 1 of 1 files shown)


wchevreuil (approved):
Thanks for submitting this fix. 

Just a nit comment: should we set pollDelay and pollInterval to 10ms? IIUC, Awaitility sets these to 100ms, by default.

Umeshkumar9414 (approved):

mnpoonia:
Thanks @wchevreuil.    i was thinking the same but then i thought how much difference will pollinterval of 10 ms and 100 ms would make. 

But Now i see that adding it will make sure test behaviour would be same as before. I will add it.

mnpoonia:
@wchevreuil Added the pollDelay and pollInterval of 10ms. Please review.

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/bucket/TestRecoveryPersistentBucketCache.java
```diff
@@ -251,9 +253,12 @@ public void testBucketCacheRecoveryWithAllocationInconsistencies() throws Except
...
       8192, bucketSizes, writeThreads, writerQLen, testDir + "/bucket.persistence",
       DEFAULT_ERROR_TOLERATION_DURATION, conf);
-    while (!newBucketCache.getBackingMapValidated().get()) {
-      Thread.sleep(10);
-    }
+    // backingMapValidated is set inside retrieveFromFile(), but isCacheEnabled() (which
+    // getBlock() checks) is set to true only in the finally block that follows. Waiting on
+    // backingMapValidated alone leaves a race window; await both conditions together.
+    await().atMost(Duration.ofSeconds(30)).pollDelay(Duration.ofMillis(10))
+      .pollInterval(Duration.ofMillis(10)).until(
```
guluo2016: For me is ok.
A small nit: pollDelay would be autimatically set as to the same value as pollInterval. 
Since both are set to 10ms here, maybe the `pollDelay(Duration.ofMillis(10))`can be removed.

Image

guluo2016 (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/io/hfile/bucket/TestRecoveryPersistentBucketCache.java
mnpoonia: You're right — Awaitility's pollDelay defaults to the same value as pollInterval when not explicitly set. So we could have written just:
   
  await().atMost(Duration.ofSeconds(30))
    .pollInterval(Duration.ofMillis(10)).until(...)

  and it would have the same behavior (10ms initial delay, 10ms between polls). Setting both explicitly just makes the intent clearer. What do you think. 
If you want it concise, we can remove pollDelay

guluo2016: > Setting both explicitly just makes the intent clearer

Agreed，thanks

virajjasani (approved):

mnpoonia:
Thank you @wchevreuil for reviewing and backporting this from master down to branch-2.6! Appreciate you shepherding this through all the branches.

PR #8464, opened 2026-07-10
HBASE-29974: Persist filter hints across scan circuit breaks

shubham-roy:
Backport of HBASE-29974 to branch-2.5 (cherry-picked from master 21d3fdfbca6, PR #7882). Adapted for branch-2.5: ExtendedCell → Cell (branch-2.5 predates the ExtendedCell migration), and the nextRowViaHint fast path drops the ScannerContext.setSkippingRow calls (that API — HBASE-27558 — is not on branch-2.5). Tests unchanged (JUnit 5, delete-marker tests retained since HBASE-29039 is present); HBaseTestingUtil → HBaseTestingUtility for the pre-rename branch. No-op by default.

Generated-by: Claude Opus 4.6 (1M context) <noreply@anthropic.com>
(cherry picked from commit 21d3fdfbca653e8effb32a4ccb44ca2ba5f5d1e1)

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/filter/Filter.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/filter/Filter.java
@@ -30,6 +30,10 @@ import org.apache.yetus.audience.InterfaceAudience;
  * <li>{@link #reset()} : reset the filter state before filtering a new row.</li>
  * <li>{@link #filterAllRemaining()}: true means row scan is over; false means keep going.</li>
  * <li>{@link #filterRowKey(Cell)}: true means drop this row; false means include.</li>
+ * <li>{@link #getHintForRejectedRow(Cell)}: if {@code filterRowKey} returned true, optionally
+ * provide a seek hint to skip past the rejected row efficiently.</li>
+ * <li>{@link #getSkipHint(Cell)}: when a cell is structurally skipped (time-range, column, or
+ * version gate) before {@code filterCell} is reached, optionally provide a seek hint.</li>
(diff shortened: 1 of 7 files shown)


Umeshkumar9414 (approved):
LGTM

hbase-client/src/main/java/org/apache/hadoop/hbase/filter/Filter.java
```diff
@@ -219,6 +223,85 @@ public enum ReturnCode {
...
+   * <p>
+   * Contract:
+   * <ul>
+   * <li>May be called for cells that have <strong>never</strong> been passed to
+   * {@link #filterCell(Cell)}.</li>
+   * <li>Implementations <strong>must not</strong> modify any filter state; this method is treated
+   * as logically stateless. Only filters whose hint computation is based solely on immutable
+   * configuration (e.g. a fixed column range or a fuzzy-row pattern) should override this.</li>
+   * <li>The returned {@link Cell}, if non-null, must be an
+   * {@link org.apache.hadoop.hbase.ExtendedCell} because filters are evaluated on the server
```
Umeshkumar9414, line 282: does this comment related to ExtendedCell still apply ?

virajjasani (approved):

hbase-client/src/main/java/org/apache/hadoop/hbase/filter/Filter.java
virajjasani: I think it's fine to keep for backport PR if master and branch-2 already has this comment. We can update them all later as addendum PR if needed.

mnpoonia (approved):

PR #8466, opened 2026-07-11
HBASE-30284 FNFE when viewing MOB Files of a MOB table restored from snapshot in HBase UI

guluo2016:
Details see: HBASE-30284

After applying this patch. 
image


liuxiaocs7 (approved):

guluo2016:
Thanks for reviewing @liuxiaocs7

PR #8439, opened 2026-06-29
HBASE-30265 Fix flaky TestProcDispatcher.testRetryLimitOnConnClosedEr…

liuxiaocs7:
…rors

- see: HBASE-30265
- failed job: https://github.com/apache/hbase/actions/runs/28292988640

PR #8467, opened 2026-07-12
HBASE-30234: Replication shippedBytes metric overflows to negative due to int truncation of batch size

asdf2014:
### Problem
Several size-tracking variables in the replication source pipeline use `int`.
Once a WAL entry batch (or a bulk load's referenced store files) exceeds
`Integer.MAX_VALUE` (~2GB), the value silently overflows: the `shippedBytes`
JMX metric goes negative and the (negative) size is fed to the bandwidth
throttler, breaking throttling.

### Fix
Widen the size to `long` end-to-end (shipper heap size, `tryThrottle` /
`getNextSleepInterval`, `MetricsSource#shipBatch`, `ReplicateContext#size`,
and `ReplicationSourceWALReader#sizeOfStoreFilesIncludeBulkLoad`).

### Testing
- New `TestReplicationSizeOverflow` and a >2GB case in `TestReplicationThrottler`.
- Before/after: reintroducing the `int` cast makes the store-file test fail with
  `expected: <3500000000> but was: <-794967296>`.
- `spotbugs:check` reports 0 bugs; changed lines are checkstyle-clean.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/ReplicationEndpoint.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/replication/ReplicationEndpoint.java
@@ -158,7 +158,7 @@ public interface ReplicationEndpoint extends ReplicationPeerConfigListener {
   @InterfaceAudience.LimitedPrivate(HBaseInterfaceAudience.REPLICATION)
   static class ReplicateContext {
     List<Entry> entries;
-    int size;
+    long size;
     String walGroupId;
     int timeout;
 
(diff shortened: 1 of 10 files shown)


asdf2014:
The `large-wave-2` failure looks unrelated to this change. The only failing test is `TestProcDispatcher.testRetryLimitOnConnClosedErrors` (a `LargeTests` under `hbase.util` covering the remote procedure dispatcher), which failed on a 5s `Waiter.waitFor` timeout — a timing flake rather than anything this patch touches (this PR only widens replication size fields from `int` to `long`). It was the only failure out of 1145 tests in that wave, and it passes locally for me. Every other check is green, including `general-check` (checkstyle/spotbugs/spotless) and `small`/`medium`/`large-wave-1`/`large-wave-3`. Could a committer please re-run the failed job? Thanks!

liuxiaocs7:
@asdf2014 Re-running! And https://github.com/apache/hbase/pull/8439 is for this flaky test.

PR #8468, opened 2026-07-12
HBASE-30248 Validate snapshot region timeout

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30248

This change validates `hbase.snapshot.region.timeout` in `RegionServerSnapshotManager`.

When the configured value is less than or equal to zero, HBase logs a warning and falls back to the documented default value of 300000 milliseconds.

All RegionServer snapshot call sites now use the same validated timeout value. The configuration documentation is also updated to state that the value must be greater than zero.

HBASE-30248 reports that a negative `hbase.snapshot.region.timeout` is passed to `ThreadPoolExecutor` as the keep-alive time during snapshot procedure initialization.

This causes an `IllegalArgumentException` and aborts the RegionServer during startup. A zero value is also invalid when core thread timeout is enabled.

Falling back to the documented default prevents an invalid timeout configuration from aborting the RegionServer.

Added `TestRegionServerSnapshotManager`, covering:

- A negative value falls back to the default.
- Zero falls back to the default.
- A positive value is preserved.

The focused unit test can be run with:

```bash
mvn -pl hbase-server -am -DskipITs -Dtest=TestRegionServerSnapshotManager test

### DIFF
--- a/hbase-common/src/main/resources/hbase-default.xml
+++ b/hbase-common/src/main/resources/hbase-default.xml
@@ -1964,6 +1964,7 @@ possible configurations would overwhelm and obscure the important.
     <value>300000</value>
     <description>
        Timeout for regionservers to keep threads in snapshot request pool waiting.
+       Must be greater than 0.
     </description>
    </property>
    <property>
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/snapshot/RegionServerSnapshotManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/snapshot/RegionServerSnapshotManager.java
@@ -102,8 +102,7 @@ public class RegionServerSnapshotManager extends RegionServerProcedureManager {
   private ProcedureMember member;
 
   static long getSnapshotTimeoutMillis(Configuration conf) {
-    long timeoutMillis =
-      conf.getLong(SNAPSHOT_TIMEOUT_MILLIS_KEY, SNAPSHOT_TIMEOUT_MILLIS_DEFAULT);
+    long timeoutMillis = conf.getLong(SNAPSHOT_TIMEOUT_MILLIS_KEY, SNAPSHOT_TIMEOUT_MILLIS_DEFAULT);
     if (timeoutMillis > 0) {
       return timeoutMillis;
     }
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by Sigma-Ma** [comment]

**Comment by Sigma-Ma** [comment]
Hi @guluo2016, sorry for the direct ping. Could you please help approve the fork workflows for this PR, or suggest an appropriate reviewer, when convenient? The original reproducer has been verified successfully. Thank you!

**Comment by Sigma-Ma** [comment]
Hi @guluo2016, I pushed a small follow-up commit to fix the Spotless formatting failure. The three workflows are awaiting approval again. Could you please approve them when convenient? Thanks!

**Comment by Sigma-Ma** [comment]
Hi @Apache9, sorry for the direct ping. All CI checks are now green, and the reporter has verified the original reproducer with both negative and zero timeout values.
Could you please review this small RegionServer snapshot fix when convenient, and merge it into master if it looks good?
The same unvalidated timeout reads are also present on branch-3, branch-2, branch-2.6, and branch-2.5. I would be happy to prepare separate backport PRs for the branches the maintainers consider appropriate. Please let me know which branches you prefer.
One JIRA bookkeeping note: HBASE-30248 is currently marked as Resolved with Fix Version 2.5.14, which has already been released. The status and Fix Versions may need to be updated to reflect the branches that actually receive the patch.
Thanks!

File: hbase-common/src/main/resources/hbase-default.xml [hunk 1]
```diff
@@ -1964,6 +1964,7 @@ possible configurations would overwhelm and obscure the important.
     <value>300000</value>
     <description>
        Timeout for regionservers to keep threads in snapshot request pool waiting.
+       Must be greater than 0.
```
guluo2016, line 1967: Thanks for your contribution. Sorry for the delayed review.
I think we should add a comment here to clarify that hbase.snapshot.region.timeout must be greater than 0.
https://github.com/apache/hbase/blob/master/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/snapshot/RegionServerSnapshotManager.java#L89-L90

guluo2016 (approved):

hbase-common/src/main/resources/hbase-default.xml
Sigma-Ma: Thanks for the review. I added a comment next to SNAPSHOT_TIMEOUT_MILLIS_KEY to clarify that the configured value must be greater than 0.

guluo2016 (approved):
LGTM, thanks

guluo2016:
Please also raise a PR for branch-2, thanks  @Sigma-Ma

Sigma-Ma:
Thanks @guluo2016. I have opened "https://github.com/apache/hbase/pull/8497" to backport this fix to branch-2.

PR #8469, opened 2026-07-13
HBASE-30285 BucketCache shutdown/disable does not release backingMap-owned BucketEntry references

mwkang:
See [HBASE-30285](https://issues.apache.org/jira/browse/HBASE-30285).

## Problem

`BucketCache.backingMap` owns the initial reference of each `BucketEntry`.

The existing non-persistent shutdown and disable path clears `backingMap` without calling `BucketEntry.markAsEvicted()`. The persistent shutdown path serializes the map but also leaves these owner references held. As a result, their reference counts may never reach zero after the map entries are discarded.

## Changes

- Move backing-map disposal to a one-shot cleanup thread after the cache transitions to `DISABLED`.
- Wait for the persistence retriever and writer threads, and stop the cache persister, before releasing backing-map entries.
- Remove each exact backing-map entry under its offset write lock and call `markAsEvicted()` only after successful conditional removal.
- Preserve references held by in-flight callers until those callers release them.
- For persistent caches, copy each entry into the persistence image and release its owner reference in the same traversal.
- Run fallback index cleanup even when final persistence fails, while leaving the previous committed checkpoint intact.
- Keep non-persistent shutdown and automatic disable cleanup asynchronous to avoid writer self-join and read-to-write lock upgrade deadlocks. Explicit persistent shutdown waits for final persistence and cleanup.
- Handle `FileIOEngine` cached-time mismatches separately from ordinary read failures: `HBaseIOException` evicts the canonical stored entry through `blockEvicted`, while a plain `IOException` keeps the existing IO-error tolerance behavior.
- Reuse the reference-key resolver and the `blocksByHFile` index instead of scanning the complete `backingMap`.
- Keep `blocksByHFile` consistent when a cached entry is replaced.
- Do not add lifecycle or global locking to normal cache admission or successful read paths.

For persistent cleanup, owner-reference release is performed during the existing serialization traversal. A successful serialization empties the map, avoiding a second full backing-map traversal.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
@@ -1390,15 +1390,13 @@ public class BucketCache implements BlockCache, HeapSize {
         blockEvicted(key, previousEntry, false, false);
         return null;
       });
-      bucketEntry.withWriteLock(offsetLock, () -> {
-        if (backingMap.get(key) == bucketEntry) {
-          blocksByHFile.add(key);
-        }
-        return null;
-      });
-    } else {
-      blocksByHFile.add(key);
     }
+    bucketEntry.withWriteLock(offsetLock, () -> {
+      if (backingMap.get(key) == bucketEntry) {
+        blocksByHFile.add(key);
+      }
+      return null;
+    });
   }
 
   /**
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
@@ -92,6 +92,7 @@ import org.apache.hadoop.hbase.util.IdReadWriteLockStrongRef;
 import org.apache.hadoop.hbase.util.IdReadWriteLockWithObjectPool;
 import org.apache.hadoop.hbase.util.IdReadWriteLockWithObjectPool.ReferenceType;
 import org.apache.hadoop.hbase.util.Pair;
+import org.apache.hadoop.hbase.util.Threads;
 import org.apache.hadoop.util.StringUtils;
 import org.apache.yetus.audience.InterfaceAudience;
 import org.slf4j.Logger;
@@ -205,6 +206,12 @@ public class BucketCache implements BlockCache, HeapSize {
    * that Bucket IO exceptions/errors don't bring down the HBase server.
    */
   private volatile CacheState cacheState;
+
+  /** The single cleanup thread shared by disable and explicit shutdown calls. */
+  private volatile Thread cacheCleanupThread;
+
+  /** The thread restoring the persistent cache index during initialization. */
+  private volatile Thread persistenceRetrieverThread;
 
   /**
    * A list of writer queues. We have a queue per {@link WriterThread} we have running. In other
@@ -416,12 +423,17 @@ public class BucketCache implements BlockCache, HeapSize {
           LOG.error("Exception during Bucket Allocation", allocatorException);
         }
       } finally {
-        this.cacheState = CacheState.ENABLED;
-        startWriterThreads();
+        synchronized (BucketCache.this) {
+          if (cacheState == CacheState.INITIALIZING) {
+            cacheState = CacheState.ENABLED;
+            startWriterThreads();
+          }
+        }
       }
     };
-    Thread t = new Thread(persistentCacheRetriever);
-    t.start();
+    persistenceRetrieverThread = new Thread(persistentCacheRetriever,
+      "BucketCachePersistenceRetriever-" + System.identityHashCode(this));
+    persistenceRetrieverThread.start();
   }
 
   private void sanityCheckConfigs() {
@@ -643,19 +655,22 @@ public class BucketCache implements BlockCache, HeapSize {
    *         the passed key doesn't relate to a reference.
    */
   public BucketEntry getBlockForReference(BlockCacheKey key) {
-    BucketEntry foundEntry = null;
-    String referredFileName = null;
-    if (StoreFileInfo.isReference(key.getHfileName())) {
-      referredFileName = StoreFileInfo.getReferredToRegionAndFile(key.getHfileName()).getSecond();
-    }
-    if (referredFileName != null) {
-      // Since we just need this key for a lookup, it's enough to use only name and offset
-      BlockCacheKey convertedCacheKey = new BlockCacheKey(referredFileName, key.getOffset());
-      foundEntry = backingMap.get(convertedCacheKey);
+    BlockCacheKey referredKey = getBlockKeyForReference(key);
+    BucketEntry foundEntry = referredKey != null ? backingMap.get(referredKey) : null;
+    if (referredKey != null) {
       LOG.debug("Got a link/ref: {}. Related cacheKey: {}. Found entry: {}", key.getHfileName(),
-        convertedCacheKey, foundEntry);
+        referredKey, foundEntry);
     }
     return foundEntry;
+  }
+
+  private BlockCacheKey getBlockKeyForReference(BlockCacheKey key) {
+    if (!StoreFileInfo.isReference(key.getHfileName())) {
+      return null;
+    }
+    String referredFileName =
+      StoreFileInfo.getReferredToRegionAndFile(key.getHfileName()).getSecond();
+    return referredFileName != null ? new BlockCacheKey(referredFileName, key.getOffset()) : null;
   }
 
   /**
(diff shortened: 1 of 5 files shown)


liuxiaocs7:
We’d better keep the Jira title and commit message consistent, thanks!

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -722,15 +764,17 @@ public Cacheable getBlock(BlockCacheKey key, boolean caching, boolean repeat,
         // When using file io engine persistent cache,
         // the cache map state might differ from the actual cache. If we reach this block,
         // we should remove the cache key entry from the backing map
-        backingMap.remove(key);
-        fileNotFullyCached(key, bucketEntry);
```
wchevreuil, line 726: Where are we tracking the file as not fully cached now? This can affect the cache aware load balancer proper functioning.

wchevreuil (commented):
Thank you for submitting this fix. I have some concerns about this, however.

Considering we call BucketCache.shutdown normally when stopping a RS process, the reference count leak here wouldn't be a big problem (the jvm process will be over soon). The introduction of locks on a hot code path, though, could have a performance impact. 

Have we done any performance benchmark comparison for validating this change is not introducing a noticeable performance impact?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
mwkang: Good catch. In the revised patch, the failed-read path now releases the offset read lock first, finds the actual key stored in backingMap, and removes the entry under its offset write lock.

After a successful conditional removal, it calls blockEvicted(cacheKey, bucketEntry, true, false), which invokes fileNotFullyCached for persistent IO engines. This keeps the file and region cache metadata used by the cache-aware load balancer consistent.

I also added regression tests for both direct and reference-file lookups.

mwkang:
The benchmark concern was valid for the previous revision, which added lifecycle locking around cacheBlock and writer-queue publication.
That hot-path locking and the queue-polling change have been removed.

Compared with master, cacheBlock and the normal writer-queue path are unchanged.
The successful getBlock path adds no new lock or backingMap scan. In getBlock, the new scan and offset write-lock path runs only after an HBaseIOException. Shutdown/disable cleanup runs after the cache is disabled and acquires per-entry offset write locks there.
Therefore, I did not run a benchmark for this revision.

The leak is also reachable when IOEngine read, write, or sync errors persist beyond the configured tolerance duration: BucketCache disables itself while the RegionServer JVM continues running. It is therefore not limited to process shutdown.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -722,22 +727,48 @@ public Cacheable getBlock(BlockCacheKey key, boolean caching, boolean repeat,
...
+        failedBucketEntryRead = true;
         LOG.debug("Failed to fetch block for cache key: {}.", key, hioex);
       } catch (IOException ioex) {
         LOG.error("Failed reading block " + key + " from bucket cache", ioex);
         checkIOErrorIsTolerated();
       } finally {
         lock.readLock().unlock();
       }
+      if (failedBucketEntryRead) {
+        removeFailedBucketEntry(bucketEntry);
```
wchevreuil: We are modifying behaviour with this. Now, in the event of an io read error, even if checkIOErrorIsTolerated ignores the error, we will evict the block?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -722,22 +727,48 @@ public Cacheable getBlock(BlockCacheKey key, boolean caching, boolean repeat,
...
+    }
+    bucketEntry.withWriteLock(offsetLock, () -> {
+      if (backingMap.remove(cacheKey, bucketEntry)) {
+        blockEvicted(cacheKey, bucketEntry, true, false);
+      }
+      return null;
+    });
+  }
+
+  private BlockCacheKey findBackingMapKey(BucketEntry bucketEntry) {
```
wchevreuil: Rather than iterating the whole backingMap, can we use similar logic from getBlockForReference to resolve the reference key? Please make the common code reusable.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -1798,70 +1829,110 @@ private void checkIOErrorIsTolerated() {
       if (isCacheEnabled() && (now - ioErrorStartTimeTmp) > this.ioErrorsTolerationDuration) {
         LOG.error("IO errors duration time has exceeded " + ioErrorsTolerationDuration
           + "ms, disabling cache, please check your IOEngine");
-        disableCache();
+        disableCache(false);
```
wchevreuil: Why we wouldn't want to persist the backingMap here?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
wchevreuil: Thanks for clarifying.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -1798,70 +1829,110 @@ private void checkIOErrorIsTolerated() {
...
-  }
-
-  @Override
-  public void shutdown() {
-    if (isCacheEnabled()) {
-      disableCache();
-      LOG.info("Shutdown bucket cache: IO persistent=" + ioEngine.isPersistent()
-        + "; path to write=" + persistencePath);
-      if (ioEngine.isPersistent() && persistencePath != null) {
+  private void cleanupCache(boolean persistOnCleanup) {
```
wchevreuil: nit: Let's call this 'cleanupBackingMap' or 'cleanupCacheIndex' to avoid confusion.

wchevreuil:
> The leak is also reachable when IOEngine read, write, or sync errors persist beyond the configured tolerance duration: BucketCache disables itself while the RegionServer JVM continues running. It is therefore not limited to process shutdown.

For file based bucket cache, backingMap can grow quite large and traversing it to release references can be costly. Should we worry about this only when shutting down BC after exceeding IO errors thresholds, and ignoring it when doing RS shutdown?

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
mwkang: Addressed in the latest revision. HBaseIOException and plain IOException now follow separate paths.

Only the cached-time mismatch HBaseIOException marks the entry as inconsistent and removes it after releasing the offset read lock. A plain IOException only calls checkIOErrorIsTolerated(). While the error remains within the configured tolerance, the backing-map entry, blocksByHFile entry, fully-cached metadata, and owner reference are retained.

testPlainIOExceptionKeepsEntryAndCacheMetadata covers this behavior.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
mwkang: I extracted getBlockKeyForReference and now reuse it from both getBlockForReference and getBlock.

The inconsistent-entry path no longer scans backingMap. It uses blocksByHFile.ceiling(lookupKey) to find the actual stored key, including its region and family metadata, verifies the hfile/offset identity, and conditionally removes the matching BucketEntry under its offset write lock.

testHBaseIOExceptionThroughReferenceEvictsStoredEntry covers the reference-file lookup path.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
mwkang: Explicit shutdown and the IO-error threshold path now use the same disable and cleanup flow. When the cache is persistent, cleanup always performs the final backing-map persistence regardless of what triggered the disable.

The retriever, writer threads, and persister are stopped before the final persistence pass. If persistence fails, the remaining owner references are still released while the previous committed checkpoint is left intact.

The empty final checkpoint and final-persistence failure cases are covered by dedicated regression tests.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
mwkang: Addressed. The cleanup orchestration method is now named cleanupCache, the aggregate index cleanup method is cleanupCacheIndex, and the per-entry operation is cleanupBackingMapEntry.

mwkang:
> > The leak is also reachable when IOEngine read, write, or sync errors persist beyond the configured tolerance duration: BucketCache disables itself while the RegionServer JVM continues running. It is therefore not limited to process shutdown.
> 
> For file based bucket cache, backingMap can grow quite large and traversing it to release references can be costly. Should we worry about this only when shutting down BC after exceeding IO errors thresholds, and ignoring it when doing RS shutdown?

I did not special-case normal RegionServer shutdown to skip owner-reference cleanup.

For a persistent file-backed cache, normal shutdown already traverses backingMap to serialize the checkpoint. The current implementation releases each owner reference through a callback in that same serializer pass, after copying the entry into the persistence image. After successful persistence, backingMap is empty, so cleanupCacheIndex does not perform a second full-map iteration. The fallback traversal only visits entries left behind when persistence fails.

This avoids an additional full traversal on the normal persistent shutdown path, while still releasing the references. I have not benchmarked this with a very large backing map, so this conclusion is based on the traversal structure rather than measured results.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
```diff
@@ -1350,13 +1384,20 @@ public void run() {
...
         return null;
       });
+      bucketEntry.withWriteLock(offsetLock, () -> {
+        if (backingMap.get(key) == bucketEntry) {
+          blocksByHFile.add(key);
+        }
+        return null;
+      });
+    } else {
+      blocksByHFile.add(key);
```
wchevreuil: Why do we need this lock when previousEntry != null, and why we don't when it's null?

wchevreuil (commented):
Thanks for working on the suggestions @mwkang. I think we are getting close, I just have an extra question about when we are adding to the backingMap under lock and when aren't under lock.

hbase-server/src/main/java/org/apache/hadoop/hbase/io/hfile/bucket/BucketCache.java
mwkang: Thanks for taking another look and catching this. There was no valid reason for the asymmetry. I moved the conditional index update under `bucketEntry`’s write lock for both the initial-insertion and replacement paths, and added a regression test for this race.

wchevreuil:
@mwkang 
Can you take a look at the latest TestBucketCache UT failure? It seems legit to me, as I could reproduce when building this branch locally.

mwkang:
Thanks for catching this. The failure was legitimate.

The test held the old entry's offset write lock while waiting for the replacement block to be fully flushed.
After `c72699b`, the writer also needs that lock to finish `putIntoBackingMap`, so both sides waited on each other.

I pushed `35db486` to fix the test's lock ordering.
The test now waits only for the replacement to be published while holding the old entry's lock, releases the lock, and then waits for the full flush.
It still verifies that the pending eviction does not remove the replacement entry and that `blocksByHFile` remains consistent.

I ran the full `TestBucketCache` suite locally with JDK 17: 68 tests, 0 failures, and 0 errors.

wchevreuil (approved):
LGTM, +1.

PR #8483, opened 2026-07-19
HBASE-30099 Avoid NPE when GetBootstrapNodes RPC arrives during startup

Sigma-Ma:

Make `HRegionServer#getBootstrapNodes` return an empty iterator when `bootstrapNodeManager` has not been initialized yet.

Declare `bootstrapNodeManager` as volatile to ensure safe publication between the RegionServer startup thread and RPC handler threads.

Add a regression test covering calls to `getBootstrapNodes` before `bootstrapNodeManager` initialization.

The RPC server can start accepting connections before `preRegistrationInitialization()` assigns `bootstrapNodeManager`.

If a `GetBootstrapNodes` RPC arrives during this startup window, `HRegionServer#getBootstrapNodes` dereferences a null `bootstrapNodeManager` and throws a `NullPointerException`.

Returning an empty iterator during this window avoids the startup race and allows callers to continue bootstrap-node discovery normally.

```bash
mvn -pl hbase-server -am \
  -Dtest=TestBootstrapNodeManager \
  -DskipITs \
  test

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java
@@ -473,7 +473,7 @@ public class HRegionServer extends HBaseServerBase<RSRpcServices>
 
   private FileSystemUtilizationChore fsUtilizationChore;
 
-  private BootstrapNodeManager bootstrapNodeManager;
+  private volatile BootstrapNodeManager bootstrapNodeManager;
 
   /**
    * True if this RegionServer is coming up in a cluster where there is no Master; means it needs to
@@ -3729,7 +3729,10 @@ public class HRegionServer extends HBaseServerBase<RSRpcServices>
 
   @Override
   public Iterator<ServerName> getBootstrapNodes() {
-    return bootstrapNodeManager.getBootstrapNodes().iterator();
+    BootstrapNodeManager manager = bootstrapNodeManager;
+    return manager != null
+      ? manager.getBootstrapNodes().iterator()
+      : Collections.<ServerName> emptyList().iterator();
   }
 
   @Override
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java
@@ -473,7 +473,7 @@ public class HRegionServer extends HBaseServerBase<RSRpcServices>
 
   private FileSystemUtilizationChore fsUtilizationChore;
 
-  private BootstrapNodeManager bootstrapNodeManager;
+  private volatile BootstrapNodeManager bootstrapNodeManager;
 
   /**
    * True if this RegionServer is coming up in a cluster where there is no Master; means it needs to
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Sigma-Ma** [comment]

**Comment by Sigma-Ma** [comment]
Hi @liuxiaocs7, sorry for the direct ping. Could you please help approve the fork workflows for this PR, or suggest an appropriate reviewer, when convenient? The original reproducer has been verified successfully. Thank you!

**Comment by Sigma-Ma** [comment]
Hi @wchevreuil and @ndimiduk, could you please take a look at this PR when convenient?

This patch fixes the RegionServer startup race reported in HBASE-30099. During RegionServer initialization, `getBootstrapNodes()` may be called before `BootstrapNodeManager` has been created, which currently results in an NPE.

The proposed fix makes the RPC return an empty bootstrap-node list during this short initialization window. Once `BootstrapNodeManager` is initialized, the existing behavior remains unchanged. A regression test has also been added to cover the RPC call before manager initialization.

@ndimiduk, since you reported HBASE-30099, it would also be helpful if you could confirm whether returning an empty list during startup matches the expected behavior.

The PR is currently mergeable and has not received a review yet. If the forked GitHub Actions workflows still require approval, could a committer please approve them as well?

Thanks!

**Comment by petersomogyi** [comment]
Hi @Sigma-Ma, apologies for the delay on this PR. I don't see an option to enable the Github Action, maybe it was idle for too long. Could you do a force push? Hopefully that will help.

File: hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/HRegionServer.java [hunk 1]
```diff
@@ -3729,7 +3729,10 @@ public List<ServerName> getBackupMasters() {
 
   @Override
   public Iterator<ServerName> getBootstrapNodes() {
-    return bootstrapNodeManager.getBootstrapNodes().iterator();
+    BootstrapNodeManager manager = bootstrapNodeManager;
+    return manager != null
+      ? manager.getBootstrapNodes().iterator()
+      : Collections.<ServerName> emptyList().iterator();
```
petersomogyi:
```suggestion
      : Collections.emptyIterator();
```

Sigma-Ma: Done, thanks. Changed it to Collections.emptyIterator().

Sigma-Ma:
> Hi @Sigma-Ma, apologies for the delay on this PR. I don't see an option to enable the Github Action, maybe it was idle for too long. Could you do a force push? Hopefully that will help.

Thanks @petersomogyi. I applied the suggestion, amended the commit.

Could you please check whether the GitHub Actions approval option is available now?

Sigma-Ma:
Hi @petersomogyi. The remaining failure looks unrelated to this patch. TestSyncReplicationStandbyKillMaster failed in the large-wave-3 shard because the RegionServer still reported the peer as STANDBY after the transition, and the retries then ran into stale mini-cluster state.

The General Check, Compile Check, and the other four unit-test shards all passed. Could you please rerun the failed job when you have a chance?

petersomogyi:
A rerun of wave3 was successful.

petersomogyi (approved):

PR #8491, opened 2026-07-22
HBASE-29696 Fix incorrect TableInputFormat split boundaries

Sigma-Ma:

Restore the original start and end rows after `Bytes.split` calculates the intermediate split keys in `TableInputFormatBase#createNInputSplitsUniform`.

Synthetic finite boundaries are still used internally for calculating the split keys, but they are no longer exposed as the boundaries of the first and last generated `TableSplit`.

A regression test was added to cover both empty and non-empty start rows with an open-ended range. The test also verifies continuity between adjacent splits.

An empty end row represents an unbounded scan in HBase. `createNInputSplitsUniform` temporarily replaces it with a finite all-`0xFF` key so that `Bytes.split` can calculate intermediate keys.

Previously, this temporary key was also used as the end row of the final `TableSplit`. Since scan stop rows are exclusive, rows equal to or lexicographically greater than the synthetic key, such as `0xFFFF`, could be skipped.

Restoring the original empty end row keeps the final split open-ended and preserves the complete input range.

Added `TestTableInputFormatBase#testCreateNInputSplitsUniformPreservesOriginalBoundaries`, covering:

- empty start and empty end rows;
- non-empty start and empty end rows;
- continuity between adjacent splits;
- preservation of the original first and last boundaries.

The regression test was run with:

mvn -pl hbase-mapreduce -am \
  -Dtest=TestTableInputFormatBase#testCreateNInputSplitsUniformPreservesOriginalBoundaries \
  -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/TableInputFormatBase.java
+++ b/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/TableInputFormatBase.java
@@ -407,6 +407,9 @@ public abstract class TableInputFormatBase extends InputFormat<ImmutableBytesWri
 
     // Split Region into n chunks evenly
     byte[][] splitKeys = Bytes.split(startRow, endRow, true, n - 1);
+    // Restore the original boundaries after using synthetic ones to calculate the split keys.
+    splitKeys[0] = ts.getStartRow();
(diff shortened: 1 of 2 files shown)


noslowerdna (approved):

Sigma-Ma:

junegunn (approved):
Thanks for catching this!

PR #8498, opened 2026-07-24
HBASE-30299 SimpleRegionNormalizer should not attempt to merge primary and secondary replicas

SaadASTheDev:
SimpleRegionNormalizer currently tries to merge primary and secondary replicas, which is invalid. While these steps get rejected, it wastes time and can block normalization from completing when secondary replicas are enabled.

Fix: filter out non-primary replicas before planning merge/split operations via a single removeIf on the region list. 

Adds a test covering the case where secondary replicas are present alongside primaries.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/normalizer/TestSimpleRegionNormalizer.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/normalizer/TestSimpleRegionNormalizer.java
@@ -750,4 +750,26 @@ public class TestSimpleRegionNormalizer {
     }
     return ret;
   }
+
+  @Test
+  public void itIgnoresSecondaryReplicasForMergeAndSplitPlanning() {
+    conf.setBoolean(SPLIT_ENABLED_KEY, true);
+    conf.setBoolean(MERGE_ENABLED_KEY, true);
+    conf.setInt(MERGE_MIN_REGION_COUNT_KEY, 1);
+    conf.setInt(MERGE_MIN_REGION_SIZE_MB_KEY, 0);
+
+    final List<RegionInfo> primaryRegions = createRegionInfos(tableName, 5);
+    final List<RegionInfo> allRegions = new ArrayList<>(primaryRegions);
+    for (RegionInfo primary : primaryRegions) {
+      allRegions.add(RegionInfoBuilder.newBuilder(tableName).setStartKey(primary.getStartKey())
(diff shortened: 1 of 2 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/master/normalizer/TestSimpleRegionNormalizer.java
```diff
@@ -750,4 +750,26 @@ private static Map<byte[], Integer> createRegionSizesMap(final List<RegionInfo>
     }
     return ret;
   }
+
+  @Test
+  public void itIgnoresSecondaryReplicasForMergeAndSplitPlanning() {
```
wchevreuil: nit: please follow the other tests naming conventions.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/normalizer/TestSimpleRegionNormalizer.java
```diff
@@ -750,4 +750,26 @@ private static Map<byte[], Integer> createRegionSizesMap(final List<RegionInfo>
...
+      allRegions.add(RegionInfoBuilder.newBuilder(tableName).setStartKey(primary.getStartKey())
+        .setEndKey(primary.getEndKey()).setRegionId(primary.getRegionId()).setReplicaId(1).build());
+    }
+
+    final Map<byte[], Integer> regionSizes = createRegionSizesMap(primaryRegions, 15, 5, 5, 15, 16);
+    setupMocksForNormalizer(regionSizes, allRegions);
+
+    assertThat(normalizer.computePlansForTable(tableDescriptor),
+      contains(new MergeNormalizationPlan.Builder().addTarget(primaryRegions.get(1), 5)
+        .addTarget(primaryRegions.get(2), 5).build()));
```
wchevreuil: Shouldn't we also check that normalizer has not created a merge plan for the replicas?

SaadASTheDev: Yup, added a assertion that verifies normalizer didn't create a merge plan for replicas

hbase-server/src/test/java/org/apache/hadoop/hbase/master/normalizer/TestSimpleRegionNormalizer.java
SaadASTheDev: Done!

charlesconnell (approved):

wchevreuil (approved):

PR #8502, opened 2026-07-24
HBASE-30300 CacheAwareLoadBalancer disproportionally favouring cache ratio over skewness

wchevreuil:


kgeisz (approved):
LGTM

hbase-balancer/src/main/java/org/apache/hadoop/hbase/master/balancer/CacheAwareLoadBalancer.java
```diff
@@ -238,20 +268,38 @@ public long getThrottleDurationMs(RegionPlan plan) {
       LOG.debug("Moving region {} to server {} with cache ratio {}. No throttling needed.",
         plan.getRegionInfo().getEncodedName(), plan.getDestination(), rsRatio.getSecond());
       return 0L;
+    }
+    // Skip throttling for regions with low cache ratio on their source server — there is
+    // negligible cached data to lose, so no warm-up delay is needed on the destination.
+    float cacheRatioOnSource = getRegionCacheRatioOnSource(plan);
+    if (cacheRatioOnSource < lowCacheRatioThreshold) {
```
taklwu, line 275: nit: do we have a test for this case?

taklwu (approved):
LGTM

hbase-balancer/src/main/java/org/apache/hadoop/hbase/master/balancer/CacheAwareLoadBalancer.java
wchevreuil: Yes. See testThrottlingRegionBeyondThreshold and testThrottlingRegionBelowThreshold in TestCacheAwareLoadBalancer.

PR #8512, opened 2026-07-29
HBASE-29984 Support separate old WAL directories in backup

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-29984

This change updates `IncrementalBackupManager` to recursively collect archived WAL files under the old WAL directory instead of assuming that every direct child is a WAL file.

It reuses `BackupUtils.getFiles` so that both the existing flat layout and the per-RegionServer directory layout enabled by `hbase.separate.oldlogdir.by.regionserver` are supported.

A regression test is added to verify that an archived WAL stored under a RegionServer-specific directory is included in the incremental backup file list.

`BackupUtils.parseHostFromOldLog` now derives the host and port from the parent directory when it is a full `ServerName`. It falls back to the existing WAL filename parsing for the flat old WAL layout.

When `hbase.separate.oldlogdir.by.regionserver` is enabled, archived WALs are stored in RegionServer-specific subdirectories under the old WAL directory.

The existing backup code uses `FileSystem.listStatus` and treats every direct child as a WAL file. It can therefore attempt to parse a directory as a WAL and fail during incremental backup log collection.

Recursively collecting files supports both directory layouts without changing the existing WAL filtering behavior.

Without this handling, a path such as `oldWALs/<serverName>/wal.<timestamp>` cannot be associated with its RegionServer because the WAL basename does not contain the host and port. `IncrementalBackupManager` would consequently skip the archived WAL.

Updated `TestIncrementalBackupManager` to create an archived WAL using the following layout: 

```bash
oldWALs/<serverName>/wal.<timestamp>
```

The test verifies that the host is obtained from the parent directory and that the archived WAL is included in the incremental backup file list.

Also ran `TestBackupUtils` to verify that the existing flat WAL filename parsing continues to work.

```bash
mvn -pl hbase-backup -am \
  -DskipITs \
  -Dtest=TestBackupUtils \
  test

mvn -pl hbase-backup -am \
  -DskipITs \
  -Dtest=TestIncrementalBackupManager \
  test
```

Addressed both review comments:

- Added direct unit coverage for the parent-directory fallback in `BackupUtils.parseHostFromOldLog`.
- Added incremental backup coverage for `SEPARATE_OLDLOGDIR=false`, while retaining the existing `SEPARATE_OLDLOGDIR=true` coverage.

### DIFF
--- a/hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/impl/IncrementalBackupManager.java
+++ b/hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/impl/IncrementalBackupManager.java
@@ -184,10 +184,10 @@ public class IncrementalBackupManager extends BackupManager {
     }
 
     // Include the .oldlogs files too.
-    FileStatus[] oldlogs = fs.listStatus(oldLogDir);
-    for (FileStatus oldlog : oldlogs) {
-      p = oldlog.getPath();
-      currentLogFile = p.toString();
+    List<String> oldlogs = BackupUtils.getFiles(fs, oldLogDir, new ArrayList<>(), path -> true);
+    for (String oldlog : oldlogs) {
+      p = new Path(oldlog);
+      currentLogFile = oldlog;
       if (AbstractFSWALProvider.isMetaFile(p)) {
         if (LOG.isDebugEnabled()) {
           LOG.debug("Skip .meta log file: " + currentLogFile);
--- a/hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/util/BackupUtils.java
+++ b/hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/util/BackupUtils.java
@@ -332,6 +332,10 @@ public final class BackupUtils {
     if (p.getName().endsWith(MasterRegionFactory.ARCHIVED_WAL_SUFFIX)) {
       return null;
     }
+    Path parent = p.getParent();
+    if (parent != null && ServerName.isFullServerName(parent.getName())) {
+      return ServerName.valueOf(parent.getName()).getAddress().toString();
+    }
     try {
       String urlDecodedName = URLDecoder.decode(p.getName(), "UTF8");
       Iterable<String> nameSplitsOnComma = Splitter.on(",").split(urlDecodedName);
--- a/hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/TestIncrementalBackupManager.java
+++ b/hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/TestIncrementalBackupManager.java
@@ -1,0 +1,104 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.backup;
+
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+import java.util.List;
+import java.util.Map;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
+import org.apache.hadoop.hbase.HBaseTestingUtil;
+import org.apache.hadoop.hbase.ServerName;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.backup.impl.BackupAdminImpl;
+import org.apache.hadoop.hbase.backup.impl.IncrementalBackupManager;
+import org.apache.hadoop.hbase.backup.util.BackupUtils;
+import org.apache.hadoop.hbase.client.Connection;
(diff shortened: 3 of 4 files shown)


hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/impl/IncrementalBackupManager.java
```diff
@@ -184,10 +184,10 @@ private List<String> getLogFilesForNewBackup(Map<String, Long> olderTimestamps,
     }
 
     // Include the .oldlogs files too.
-    FileStatus[] oldlogs = fs.listStatus(oldLogDir);
-    for (FileStatus oldlog : oldlogs) {
-      p = oldlog.getPath();
-      currentLogFile = p.toString();
+    List<String> oldlogs = BackupUtils.getFiles(fs, oldLogDir, new ArrayList<>(), path -> true);
```
taklwu, line 187: can you double check if `BackupUtils#parseHostFromOldLog` would need to be changed when `hbase.separate.oldlogdir.by.regionserver=true` ? basically I checked the logic that when `hbase.separate.oldlogdir.by.regionserver=true` , there may have archived file without the host:port but the host name is part of the directory , e.g. `oldWALs/10.0.0.1,60030,1700000000000/wal.00000000000000000001`

taklwu (changes requested):

hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/impl/IncrementalBackupManager.java
Sigma-Ma: Good catch. I double-checked the built-in WAL creation and archive paths.

For the standard WAL providers, enabling hbase.separate.oldlogdir.by.regionserver only changes the archive directory. The WAL filename still normally contains the ServerName because WALFactory uses the ServerName as the WAL prefix and archiving preserves the original filename.

However, parseHostFromOldLog currently ignores the parent directory, so a path such as oldWALs/<serverName>/wal.<timestamp> would return null and the archived WAL would be skipped by incremental backup collection.

I agree that we should handle this problem. I will update parseHostFromOldLog to use a valid full ServerName from the parent directory, falling back to the existing filename parsing for the flat oldWAL layout, and adjust the regression test to cover a hostless WAL filename.

Sigma-Ma: BackupUtils.parseHostFromOldLog now checks whether the parent directory is a full ServerName and uses it to derive host:port. The existing filename parsing is retained as the fallback for the flat oldWAL layout.

I also updated TestIncrementalBackupManager to use oldWALs/<serverName>/wal.<timestamp>, so the regression test now covers the hostless WAL filename case directly.

TestBackupUtils and TestIncrementalBackupManager both pass.

Sigma-Ma:
Hi @taklwu, thanks again for the review. I’ve addressed your concern : BackupUtils.parseHostFromOldLog now reads the ServerName from the parent directory for per-RegionServer oldWAL layouts, while preserving the existing filename fallback.

I also added coverage for oldWALs/<serverName>/wal.<timestamp>, and the focused tests pass. When you have time, could you please take another look? Thanks!

hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/util/BackupUtils.java
```diff
@@ -332,6 +332,10 @@ public static String parseHostFromOldLog(Path p) {
     if (p.getName().endsWith(MasterRegionFactory.ARCHIVED_WAL_SUFFIX)) {
       return null;
     }
+    Path parent = p.getParent();
+    if (parent != null && ServerName.isFullServerName(parent.getName())) {
+      return ServerName.valueOf(parent.getName()).getAddress().toString();
+    }
```
taklwu, line 338: Thanks for the update. I see `TestIncrementalBackupManager` creates `oldWALs/<serverName>/wal.<ts>` with SEPARATE_OLDLOGDIR=true, which test the full incremental path end-to-end. 

That’s helpful, but it doesn’t add an unit test for `BackupUtils.parseHostFromOldLog()` with the parent-directory fallback in TestBackupUtils. Could you add such test that asserts host parsing for `oldWALs/<serverName>/wal.*` directly?

hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/TestIncrementalBackupManager.java
```diff
@@ -0,0 +1,90 @@
...
+import org.junit.jupiter.api.Test;
+
+@Tag(LargeTests.TAG)
+public class TestIncrementalBackupManager extends TestBackupBase {
+
+  @BeforeAll
+  public static void setUp() throws Exception {
+    TEST_UTIL = new HBaseTestingUtil();
+    conf1 = TEST_UTIL.getConfiguration();
+    conf1.setBoolean(AbstractFSWALProvider.SEPARATE_OLDLOGDIR, true);
```
taklwu, line 51: nit: is is possible to have one test for `SEPARATE_OLDLOGDIR=false` as well here?

taklwu (changes requested):
LGTM and it's very close, two minor comments and please help to address them

hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/util/BackupUtils.java
Sigma-Ma: I have added direct coverage in TestBackupUtils. The test constructs `oldWALs/<serverName>/wal.<timestamp>` and verifies that `BackupUtils.parseHostFromOldLog` returns the expected host:port from the parent ServerName directory.

hbase-backup/src/test/java/org/apache/hadoop/hbase/backup/TestIncrementalBackupManager.java
Sigma-Ma: I have added `testCollectWALFilesFromFlatOldWALDirectory` for `SEPARATE_OLDLOGDIR=false`.

The flat and per-RegionServer layouts now share the same test helper: the flat layout uses a WAL filename containing the ServerName, while the separate layout uses a hostless filename under the ServerName directory.

Both cases pass.

taklwu (approved):
LGTM

PR #8515, opened 2026-07-30
HBASE-23996 Add metric for the split WAL procedure

Umeshkumar9414:
We already have deprecated HBASE_SPLIT_WAL_COORDINATED_BY_ZK but some Classes only used by that flow was not marked deprecated. I have marked those deprecated in this PR only. I hope that is fine else I will create another PR for that. 

Some screen shots for the old metrics and the new one. 
example 1 
image (11)
example 2 
image (12)
Old one
image (13)

### DIFF
--- a/hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterFilesystemSourceImpl.java
+++ b/hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterFilesystemSourceImpl.java
@@ -21,6 +21,13 @@ import org.apache.hadoop.hbase.metrics.BaseSourceImpl;
 import org.apache.hadoop.metrics2.MetricHistogram;
 import org.apache.yetus.audience.InterfaceAudience;
 
+/**
+ * @deprecated since 2.4.0 and in 3.0.0, to be removed in 4.0.0, replaced by procedure-based
+ *             distributed WAL splitter; see SplitWALManager. Implementation of the deprecated
+ *             {@link MetricsMasterFileSystemSource}, whose metrics are only emitted by the
+ *             ZK-coordinated WAL split path.
+ */
+@Deprecated
 @InterfaceAudience.Private
 public class MetricsMasterFilesystemSourceImpl extends BaseSourceImpl
   implements MetricsMasterFileSystemSource {
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/coordination/ZKSplitLogManagerCoordination.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/coordination/ZKSplitLogManagerCoordination.java
@@ -58,8 +58,11 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 /**
- * ZooKeeper based implementation of {@link SplitLogManagerCoordination}
+ * ZooKeeper based implementation of {@link SplitLogManagerCoordination}.
+ * @deprecated since 2.4.0 and in 3.0.0, to be removed in 4.0.0, replaced by procedure-based
+ *             distributed WAL splitter; see SplitWALManager.
  */
+@Deprecated
 @InterfaceAudience.Private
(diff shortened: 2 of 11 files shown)


apurtell (approved):

mnpoonia:
The new success-path integration test only verifies `splitWALSubmittedCount`. Please also assert that `splitWALTime` receives an entry after the completed procedure. splitWALFailedCount has no corresponding terminal-failure test; if a stable failure path is impractical, please explain why that metric is intentionally left without integration coverage.
@Umeshkumar9414

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestSplitWALManager.java
```diff
@@ -251,6 +261,9 @@ private void splitLogsTestHelper(HBaseTestingUtil testUtil) throws Exception {
     ProcedureTestingUtility.submitAndWait(masterPE, procedures.get(0));
     assertEquals(0, splitWALManager.getWALsToSplit(metaServer, true).size());
     assertEquals(1, splitWALManager.getWALsToSplit(metaServer, false).size());
+    // The meta SplitWALProcedure should also have been counted by the split WAL metric.
+    METRICS_HELPER.assertCounter(MetricsMasterSource.SPLIT_WAL_METRIC_PREFIX + "SubmittedCount",
```
mnpoonia, line 265: This validates `splitWALSubmittedCount`, but the new OperationMetrics also exports `splitWALTime` and `splitWALFailedCount`. Could we assert that the time histogram receives one sample for each successfully completed SplitWALProcedure here?  

I do not see a deterministic terminal-failure path for SplitWALProcedure, so I would not require a FailedCount integration test in this PR. The generic procedure metrics test covers failure accounting; an assertion that FailedCount remains unchanged on this successful/retryable path would be a useful optional addition.

hbase-server/src/main/java/org/apache/hadoop/hbase/coordination/ZKSplitLogManagerCoordination.java
```diff
@@ -58,8 +58,11 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * ZooKeeper based implementation of {@link SplitLogManagerCoordination}
+ * ZooKeeper based implementation of {@link SplitLogManagerCoordination}.
```
mnpoonia, line 61: Could we make the replacement a Javadoc link? SplitWALManager is in org.apache.hadoop.hbase.master, so use {@link org.apache.hadoop.hbase.master.SplitWALManager}. This makes the migration target navigable in generated API docs.

hbase-server/src/main/java/org/apache/hadoop/hbase/coordination/ZkSplitLogWorkerCoordination.java
```diff
@@ -57,9 +57,11 @@
 import org.slf4j.LoggerFactory;
 
 /**
- * ZooKeeper based implementation of {@link SplitLogWorkerCoordination} It listen for changes in
- * ZooKeeper and
+ * ZooKeeper based implementation of {@link SplitLogWorkerCoordination}.
+ * @deprecated since 2.4.0 and in 3.0.0, to be removed in 4.0.0, replaced by procedure-based
```
mnpoonia, line 61: Same javadoc comment applies here

hbase-hadoop-compat/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterFilesystemSourceImpl.java
```diff
@@ -21,6 +21,13 @@
 import org.apache.hadoop.metrics2.MetricHistogram;
 import org.apache.yetus.audience.InterfaceAudience;
 
+/**
+ * @deprecated since 2.4.0 and in 3.0.0, to be removed in 4.0.0, replaced by procedure-based
+ *             distributed WAL splitter; see SplitWALManager. Implementation of the deprecated
```
mnpoonia, line 26: Same javadoc comment applies here

PR #8364, opened 2026-06-16
HBASE-29081: Add HBase Read Replica Cluster feature (#8044)

Kota-SH:
Backporting the Read Replica Feature PR #8044 
(cherry picked from commit 9ba5d3e55a3624193e8c4ab13631db058b419546)

PR #8517, opened 2026-07-30
HBASE-30087: Create a nightly Jenkins test for the Read-Replica feature

kgeisz:
[HBASE-30087: Create a nightly Jenkins test for the Read-Replica feature](https://issues.apache.org/jira/browse/HBASE-30087)

Note: Some code in the pull request was written using Claude Opus 4.6.

# Introduction

- [Apache HBase Read Replica Cluster Feature](https://docs.google.com/document/d/1EI0lsURX1BZhv3DYgMvZCl4EUy-ADJRkHUc1PjzZtj0/edit?usp=sharing)
- [Read Replica Test Plan Document](https://docs.google.com/document/d/1liMDJd8xtgu9ExzfjnY6mY6hYNd4Y_7H-MpVZ5qWZJg/edit?usp=sharing)
- [HBase Website Documentation for Read Replica Cluster](https://hbase.apache.org/docs/read-replica-cluster)

HBase's Read-Replica feature was merged into master in PR #8364. This pull request introduces integration tests for Read-Replica via Jenkins and Docker containers. Jenkins and Docker containers have been useful for testing this feature because it works around the `META_TABLE_NAME` issue mentioned in [HBASE-29691](https://issues.apache.org/jira/browse/HBASE-29691) and PR #7730.

In a Read-Replica setup, clusters share the same storage location and need to have different `META_TABLE_NAME`s. The table names are distinguished using the `hbase.meta.table.suffix` configuration property. However, this still leads to problems when using `MiniHBaseCluster` to test a multi-cluster Read-Replica setup because these clusters run in the same JVM and end up sharing the same `static META_TABLE_NAME` variable.

# How It Works

- A Read-Replica stage has been added to the `dev-support/Jenkinsfile`.
- The `dev-support/read-replica` directory contains various files for building a Docker image, running containers with HBase procecces, and running integration test scripts written in Python. It also contains files and directories used by HBase, such as `conf` directories.
- The `hbase_docker_client.py` file is the most important Python file. Every test script uses this file to communicate with each hbase-docker container. It does so by using `docker exec` to run commands in a container's HBase shell.

The tests are triggered within a stage in HBase's nighly Jenkins run. The stage starts by cloning the current HBase repository to dev-support/read-replica. From here, the repo gets used to build a Docker image designed to run HBase in a Read-Replica setup. After the image has been built, two Docker containers are started, where one starts as the active cluster (read-write mode), and the other container starts as a replica cluster (read-only mode). These clusters share a `data-store` directory containing the `hbase.rootdir`. It is a mounted volume between each container and the local filesystem, and it is created by the Python scripts before container startup. The directory is also given `777` permissions in order to avoid HBase startup failures. There are other volumes as well for easy access, such as each container's `conf` directory, a `utils` directory for bulkloading data, and a `logs` directory for preserving HBase logs from each container.

Once the containers are up and running, a series of Python scripts are run as integration tests. They test expected behavior for a Read-Replica cluster setup, such as verifying valid/invalid startup, blocking writes on replica clusters, being able to refresh meta and HFiles on replica clusters to make them consistent with the active cluster, read-only flag flipping (changing `hbase.global.readonly.enabled` via dynamic configuration), verifying bug fixes, etc. If the Jenkins stage finishes successfully or a failure occurs at any point, then the HBase logs are saved to the Jenkins output directory, the containers are stopped, and the Docker image is deleted.

# Other Information

- `docker-compose.yml` defines and configures two hbase-docker containers capable of running in a read-replica setup.
- `build_images.sh` uses `Dockerfile` to build an hbase-docker image. This script assumes there is an `hbase` repo in the same directory.
- `.env` defines environment variables used by `docker-compose.yml`, `build_images.sh`, and Python scripts.
- `proto_compiler.py` copies `ActiveClusterSuffix.proto` from the hbase repo and compiles it. The generated output is used for verifying the `active.cluster.suffix.id` file.
- `utils` is a directory containing scripts for bulkloading data into HBase.
- The `cluster1` and `cluster2` directories contain cluster-specific conf files and HBase log output directories. They each contain their own `hbase-site.xml`, `log4j2.properties`, and `zoo.cfg` files. The directories are defined as mounted volumes in `docker-compose.yml`. This is especially useful for changing the value of `hbase.global.readonly.enabled` in order to change a cluster's read-only mode. The mounted logs directory is useful because it helps preserve the HBase logs so they can be copied to the Jenkins output directory if a test failure occurs.

### DIFF
--- a/.github/workflows/read-replica-integration-tests.yml
+++ b/.github/workflows/read-replica-integration-tests.yml
@@ -1,0 +1,120 @@
+name: Read-Replica Integration Tests
+
+on:
+  pull_request:
+    types: [labeled, synchronize, opened, reopened, ready_for_review]
+
+# Cancel running tests for previous commits on the same PR to save time/resources
+concurrency:
+  group: read-replica-tests-${{ github.event.pull_request.number }}
+  cancel-in-progress: true
+
+jobs:
+  read-replica-integration-test:
+    # Trigger only if the 'read-replica' label is attached and the PR is not a draft
+    if: ${{ contains(github.event.pull_request.labels.*.name, 'read-replica') && !github.event.pull_request.draft }}
+    env:
+      REPLICA_BASE_DIR: ./dev-support/integration-test/read-replica
+    defaults:
+      run:
+        # Change default working directory from GITHUB_WORKSPACE to this
+        working-directory: ${{ env.REPLICA_BASE_DIR }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout HBase Repository with Pull Request's SHA
+        uses: actions/checkout@v4
+
+      - name: Checkout HBase Respoitory Again Inside read-replica Directory
+        uses: actions/checkout@v4
+        with:
+          # Place repo inside dev-support/integration-test/read-replica
+          # Use this repo to build the hbase-docker image
+          path: ${{ env.REPLICA_BASE_DIR }}/hbase
+
+      - name: Show Working Directory
+        run: |
+          echo "The previous default working directory was: GITHUB_WORKSPACE=${GITHUB_WORKSPACE}"
+          echo "The current and default working directory is now: $(pwd)"
+          echo "The working directory now has an HBase repo:"
+          ls -la
+          echo "Showing the most recent commit in this HBase repo:"
+          cd hbase && git log --oneline | head
+
+      - name: Load .env File into GitHub Actions Environment
+        run: |
+          # Turn on auto-export and source the file
+          set -a
+          source .env
+          set +a
+
+          # Filter out comments, grab the keys, and write the expanded values to GITHUB_ENV
+          grep -E "^[A-Za-z0-9_]+=" .env | cut -d= -f1 | while read -r key; do
+            echo "$key=${!key}"
+            echo "$key=${!key}" >> $GITHUB_ENV
+          done
+
+      - name: Set up Docker Compose
+        uses: docker/setup-compose-action@v2
+
+      - name: Install Python Requirements
+        run: |
+          pip install -r requirements.txt
+
+      - name: Set Python Path
+        run: |
+          echo "Setting PYTHONPATH=$(pwd) for GITHUB_ENV=${GITHUB_ENV}"
+          echo "PYTHONPATH=$(pwd)" >> ${GITHUB_ENV}
+
+      - name: Compile Protobuf
+        run: |
+          echo "Copying ActiveClusterSuffix.proto from HBase repo to have latest version"
+          cp hbase/hbase-protocol-shaded/src/main/protobuf/server/ActiveClusterSuffix.proto \
+             python/proto
+          python3 python/proto/proto_compiler.py
+
+      - name: Build Docker Image
+        run: |
+          echo "Building docker image"
+          ./build-images.sh
+          echo "Listing Docker images"
+          docker image ls
+
+      - name: Verify Can't Start Two Active Clusters at Once
+        run: |
+          # This script automatically prepares the HBASE_DATA_STORE_ROOT defined in .env.
+          # This directory and its sub-directories need 777 permissions to prevent HBase
+          # from failing on startup. The directories are deleted when the containers are stopped
+          python3 python/scripts/test_dual_active_cluster_startup.py --clean-up-containers
+
+      - name: Properly Start HBase Docker Containers
+        run: |
+          # This script also prepares HBASE_DATA_STORE_ROOT, similar to the previous step
+          echo "Starting one active and one replica HBase cluster and waiting for them to initialize..."
+          python3 python/scripts/verify_hbase_start.py
+
+      - name: Test Table Create/Drop Behavior
+        run: python3 python/scripts/test_create_drop_behavior.py --skip-table-cleanup-on-start
+
+      - name: Test Put/Get/Delete Behavior
+        run: python3 python/scripts/test_put_get_delete_behavior.py --skip-table-cleanup-on-start
+
+      - name: Test Read-Only Flag Flipping
+        run: |
+          python3 python/scripts/test_read_only_flag_flipping.py --skip-container-start-or-restart
+
+      - name: Test Cannot Promote Replica to Second Active Cluster
+        run: |
+          python3 python/scripts/test_cannot_promote_second_active_cluster.py --skip-container-start-or-restart
+
+      - name: Test Bulkloaded Data and Region Splitting
+        run: |
+          python3 python/scripts/test_bulkloaded_data_and_region_splits.py --skip-container-start-or-restart
+
+      - name: Clean up
+        run: docker compose -f docker-compose.yml down
+
+      - name: Dump Logs on Failure
+        if: failure()
+        run: |
+          docker logs hbase-docker
+          docker logs hbase-docker-2
--- a/dev-support/integration-test/read-replica/.env
+++ b/dev-support/integration-test/read-replica/.env
@@ -1,0 +1,25 @@
+# The name of the HBase Docker image
+HBASE_IMAGE=kgeisz/hbase-docker:read-replica-gha
+# The name of the HBase docker container
+HBASE_CONTAINER_NAME=hbase-docker
+# The directory within the docker container that contains the config files
+HBASE_CONF_DIR=/opt/hbase/conf
+# The directory containing the 'data-store/<hbase.rootdir>' directory.
+# This directory is mounted with the HBase Docker container.
+HBASE_DATA_STORE_ROOT=${RUNNER_TEMP}/tmp
+# The port for the active cluster's HBase UI (used for cluster readiness)
+ACTIVE_CLUSTER_PORT=16010
+# The port for the replica cluster's HBase UI (used for cluster readiness)
+REPLICA_CLUSTER_PORT=26010
+# Local path to the active cluster's HBase config file.
+# This file is part of a mounted volume, so modifying it
+# locally also modifies it within the container (and vise-versa).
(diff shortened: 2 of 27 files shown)


Apache9:
Better add this to nightly or integration tests. Typically we do not require integration test when merging a PR.

dev-support/read-replica/python/src/hbase_docker_client.py
```diff
@@ -0,0 +1,529 @@
...
+        return self._cluster_name
+
+    def run_docker_exec_command(self, bash_cmd: str, timeout: int | None = None) -> str:
+        """
+        Uses 'docker exec' to run the provided Bash command in the object's Docker container.
+        The command looks like: docker exec <container> bash -c <bash_cmd>
+        Note: In the Terminal, we usually put double quotes around everything after "-c",
+        but doing that with subprocess.run() results in a failure.
+        """
+        cmd = ["docker", "exec", self._container_name, "bash", "-c", f'''{bash_cmd}''']
```
anmolnar: Could it be easier to use `docker` python module to execute commands inside the container instead of subprocessing the docker command?

```
exec_run(cmd, stdout=True, stderr=True, stdin=False, tty=False, privileged=False, user='', detach=False, stream=False, socket=False, environment=None, workdir=None, demux=False)

Run a command inside this container. Similar to docker exec.
```

https://docker-py.readthedocs.io/en/stable/containers.html

anmolnar (approved):
lgtm.

dev-support/read-replica/python/src/hbase_docker_client.py
kgeisz: I updated `run_docker_exec_command()` method in `hbase_docker_client.py` do use Docker's Python library for running `docker exec` commands. Unfortunately, this library does not support `docker compose`, so I am still using Python's `subprocess` to run `docker compose` commands.

kgeisz:
Hi @Apache9, thanks for the feedback! I updated the PR by adding a read-replica stage to `dev-support/Jenkinsfile`.  I removed the GitHub Action workflow file as well.  Unfortunately, I am not an HBase committer, so I cannot test this file in an upstream Jenkins sandbox. I did, however, test the read-replica stage using a local Jenkins Docker container.  The Jenkinsfile I used was a tuncated version of the upstream Jenkinsfile that only has code relevent to the read-replica test stage (I left out Yetus stages, etc.).

Is there any way this updated Jenkinsfile can be tested with upstream Jenkins before merging the PR?
cc. @anmolnar @taklwu @ndimiduk @apurtell

anmolnar (approved):
Still looks good to me.

dev-support/read-replica/Dockerfile
```diff
@@ -0,0 +1,131 @@
+# Stage 0: Cache Maven dependencies
+ARG BASE_IMAGE=registry.access.redhat.com/ubi8/openjdk-17:1.23
+FROM ${BASE_IMAGE} AS cache-stage
```
taklwu, line 4: it still failed the build as below 



```
dev-support/read-replica/Dockerfile:3: DL3006 warning: Always tag the version of an image explicitly
dev-support/read-replica/Dockerfile:27: DL3006 warning: Always tag the version of an image explicitly
dev-support/read-replica/Dockerfile:54: DL3006 warning: Always tag the version of an image explicitly
```

dev-support/read-replica/Dockerfile
```diff
@@ -0,0 +1,131 @@
...
+        git \
+        diffutils \
+        vim \
+        openssh-clients \
+        python3 \
+        procps \
+    && microdnf clean all
+
+# Copy the built HBase binaries from the build-stage
+COPY --from=build-stage /opt/hbase-src/hbase-assembly/target/hbase-4.0.0-alpha-1-SNAPSHOT-bin.tar.gz /opt/
```
taklwu, line 88: nit: it's better to get this version `4.0.0-alpha-1-SNAPSHOT` from pom.xml for future development.

dev-support/read-replica/python/src/hbase_docker_client.py
```diff
@@ -0,0 +1,623 @@
...
+        output = self.run_hbase_shell_command(get_cmd)
+        logger.debug(f"Got data:\n{output}")
+        return output
+
+    def delete(self, table_name: str, row: str, column: str, timestamp: int | None = None,
+               spec_map: str | None = None) -> None:
+        logger.info(f"Deleting data from table '{table_name}' on {self.name}")
+        delete_cmd = f"delete '{table_name}', '{row}', '{column}'"
+        if timestamp:
+            delete_cmd += f", {table_name}"
```
taklwu, line 299:
```suggestion
            delete_cmd += f", {timestamp}"
```

dev-support/hbase_nightly_read_replica_test.sh
```diff
@@ -0,0 +1,117 @@
...
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+#
+# Run the read-replica Docker integration test suite.
+
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPLICA_DIR="${SCRIPT_DIR}/read-replica"
```
taklwu, line 24: nit: please define `OUTPUT_DIR`

```suggestion
REPLICA_DIR="${SCRIPT_DIR}/read-replica"
OUTPUT_DIR="${OUTPUT_DIR:-${REPLICA_DIR}/output}"
```

dev-support/read-replica/python/src/hbase_docker_client.py
```diff
@@ -0,0 +1,623 @@
...
+
+    def disable_table(self, table_name: str) -> None:
+        logger.debug(f"Disabling table '{table_name}' on {self.name}")
+        self.run_hbase_shell_command(f"disable '{table_name}'")
+
+    def drop_table(self, table_name: str) -> None:
+        logger.info(f"Dropping table '{table_name}' on {self.name}")
+        self.run_hbase_shell_command(f"drop '{table_name}'")
+
+    def list_tables(self) -> list:
```
taklwu, line 253: nit: will we hit any empty list case? if so, please handle it

dev-support/read-replica/requirements.txt
```diff
@@ -0,0 +1,13 @@
+certifi==2026.2.25
+charset-normalizer==3.4.7
+docker==7.2.0
+dotenv==0.9.9
```
taklwu, line 4: nit: reviewed by the cursor, `dotenv` should be removed and only need `python-dotenv`

taklwu (changes requested):
@Apache9 @ndimiduk sorry for pinging you guys directly, how do we add a nightly Jenkins job after this PR merge? is it automated?

dev-support/read-replica/Dockerfile
kgeisz: `hadolint` is not giving me this warning when I run it locally. Maybe the linter being used here cannot parse `${BASE_IMAGE}`. I added `hadolint ignore=DL3006` to these `FROM ${BASE_IMAGE}` lines to suppress the warning.

Also, in case you didn't notice, I chose to ignore some other warnings as well. Let me know if you want me to address these.

```
# hadolint ignore=DL3002 to suppress:
./Dockerfile:11 DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
./Dockerfile:35 DL3041 warning: Specify version with `dnf install -y <package>-<version>`.
./Dockerfile:70 DL3041 warning: Specify version with `dnf install -y <package>-<version>`.

# hadolint ignore=DL3002 to suppress:
./Dockerfile:8 DL3002 warning: Last USER should not be root
./Dockerfile:33 DL3002 warning: Last USER should not be root
```

dev-support/read-replica/python/src/hbase_docker_client.py
kgeisz: The empty list case has been handled. The regex grabs the `[...]` text from the HBase shell's `list` output, and the final line in the return statement evaluates it and turns it into Python code. If there are no tables, then we get `[]`, which is evaluated as an empty list.

dev-support/read-replica/python/__init__.py
taklwu, line 1: nit: do we really need this init file ? I saw few of them are empty with just the ASF header

dev-support/read-replica/python/src/hbase_docker_client.py
```diff
@@ -0,0 +1,643 @@
...
+
+class HBaseShellCommandError(DockerExecCommandError):
+    pass
+
+
+class DockerExecCommandTimeoutError(DockerExecCommandError):
+    pass
+
+
+class HBaseDockerClient:
```
taklwu, line 48: nit: this itself could be a single commit to start the container with hbase cluster and client

taklwu (approved):
LGTM

kgeisz:
I made some changes to how the tests are run. The tests are now run with `pytest`, which allows us to generate `xml` and `html` test results for Jenkins. Here are some screenshots from my local Jenkins deployment to give an idea of what the results looks like:

Screenshot 2026-09-16 at 8 06 25 AM
Screenshot 2026-09-16 at 8 06 54 AM

taklwu:
@kgeisz let's learnt from #8601. 

1. create a jenkins feature branch , review and merge this PR
2. create the jenkins job based on the feature branch , keep monitoring for 1 weeks or 2. 
3. create the merge PR into master.

kgeisz:
Changed the destination branch to `HBASE-30087` so the Jenkins changes can be tested before they are merged into `master`.

PR #8519, opened 2026-07-31
HBASE-30303: Fix balancing for regions assigned through RSGroup fallback

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30303

This pull request updates `RSGroupBasedLoadBalancer#correctAssignments` to use the same candidate-server selection logic as region assignment when RSGroup fallback is enabled.

When the target RSGroup has no online RegionServers, `correctAssignments` now validates region locations against the effective fallback candidates.

The existing `TestRSGroupsFallback` test is also extended to verify that no region moves are calculated when regions are correctly hosted on fallback servers.

When RSGroup fallback is enabled and a table's target RSGroup has no online RegionServers, its regions are assigned to the default RSGroup, or to any available RegionServer if the default group is also unavailable.

Previously, `correctAssignments` only validated these regions against the table's original RSGroup. Therefore, regions already hosted on valid fallback servers were incorrectly classified as misplaced during every balance run, resulting in continuous and unnecessary region movement.

The following tests were executed:

```bash
mvn -pl hbase-server \
  -am \
  -Dtest=org.apache.hadoop.hbase.rsgroup.TestRSGroupsFallback \
  test

mvn -pl hbase-server \
  -am \
  -Dtest=org.apache.hadoop.hbase.master.balancer.TestRSGroupBasedLoadBalancer \
  test

```

`TestRSGroupsFallback` verifies that balancing calculates no moves after regions fall back to the default RSGroup and after they fall back to another available RSGroup.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRSGroupsFallback.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRSGroupsFallback.java
@@ -17,6 +17,7 @@
  */
 package org.apache.hadoop.hbase.rsgroup;
 
+import static org.junit.jupiter.api.Assertions.assertEquals;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
 import java.io.IOException;
@@ -25,6 +26,7 @@ import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HBaseTestingUtil;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.BalanceResponse;
 import org.apache.hadoop.hbase.client.ColumnFamilyDescriptorBuilder;
 import org.apache.hadoop.hbase.client.TableDescriptor;
 import org.apache.hadoop.hbase.client.TableDescriptorBuilder;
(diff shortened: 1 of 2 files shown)


hbase-server/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRSGroupsFallback.java
```diff
@@ -93,10 +94,12 @@ public void testFallback() throws Exception {
     // server of test group crash, regions move to default group
     crashRsInGroup(groupName);
     assertRegionsInGroup(tableName, RSGroupInfo.DEFAULT_GROUP);
+    assertEquals(0, MASTER.balance().getMovesCalculated());
 
     // server of default group crash, regions move to any other group
     crashRsInGroup(RSGroupInfo.DEFAULT_GROUP);
     assertRegionsInGroup(tableName, FALLBACK_GROUP);
+    assertEquals(0, MASTER.balance().getMovesCalculated());
```
chaijunjie0101: maybe we need trigger balancer once before we check the balance plan，and make sure balancer is doing to generate plan, because the default balance is scheduled

chaijunjie0101:
please confirm all UT test passed @Sigma-Ma

hbase-server/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRSGroupsFallback.java
Sigma-Ma: Thanks, good catch. `movesCalculated` is also zero when the balancer is skipped, so the previous assertion could pass without `balanceCluster` actually running.

I updated the test to capture the `BalanceResponse` from the same balance invocation, assert `isBalancerRan()` first, and then assert that `getMovesCalculated()` is zero for both fallback cases.

Sigma-Ma:
> please confirm all UT test passed @Sigma-Ma

I ran the unit test suite with:

```bash
mvn \
  -pl hbase-server \
  -am \
  -Dtest=org.apache.hadoop.hbase.rsgroup.TestRSGroupsFallback \
  test

mvn \
  -pl hbase-server \
  -am \
  -Dtest=org.apache.hadoop.hbase.master.balancer.TestRSGroupBasedLoadBalancer \
  test

mvn \
  -pl hbase-server \
  -am \
  -P runAllTests \
  test
```

The HBASE-30303 related tests passed, but the full run encountered one unrelated failure: TestExecutorService hit an existing synchronization race under high system load. I also reproduced the TestExecutorService failure in isolation.

chaijunjie0101 (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRSGroupsFallback.java
chaijunjie0101: > Thanks, good catch. `movesCalculated` is also zero when the balancer is skipped, so the previous assertion could pass without `balanceCluster` actually running.
> 
> I updated the test to capture the `BalanceResponse` from the same balance invocation, assert `isBalancerRan()` first, and then assert that `getMovesCalculated()` is zero for both fallback cases.

great！need HBase's committer to help review and approve, if you have time, please check and review this PR @guluo2016

Sigma-Ma:
Hi @guluo2016, could you please take a look at this PR when you have time?

@chaijunjie0101 has reviewed and approved it, and I have addressed the review feedback by ensuring that the balancer actually runs before asserting that no balance plan is generated. The focused RSGroup tests passed, and the unrelated failures encountered during the full UT run have been documented above.

If the change looks good, could you please approve it and help merge it?

@Apache9, since this change touches RSGroupBasedLoadBalancer, your review would also be greatly appreciated. Thanks!

PR #8531, opened 2026-08-06
HBASE-30317 Fix synchronization race in TestExecutorService

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30317

Update `TestExecutorService.testExecutorService` to poll fresh `ExecutorStatus` snapshots with `Waiter.waitFor` until the status dump shows a handler waiting on the `AtomicBoolean` monitor.

The existing dump check now returns a boolean used as the wait predicate. No fixed sleep is introduced.

The test previously waited only for the shared counter to reach five. However, each handler increments the counter before performing its initial logging and entering `lock.wait()`.

Therefore, `counter == 5` proves only that all handlers have started. Under resource pressure, the status dump can be captured while all handlers are still RUNNABLE in `StringConcatFactory.makeConcatWithConstants`, causing the expected waiting-state assertion to fail.

Waiting for the expected state in the actual executor status dump removes this synchronization race.

## How was this patch tested?

```bash
mvn \
  -pl hbase-server -am \
  -Dtest=org.apache.hadoop.hbase.executor.TestExecutorService \
  test
```

The resulting executor status dump showed all five handlers in the `WAITING` state on the `AtomicBoolean` monitor.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/executor/TestExecutorService.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/executor/TestExecutorService.java
@@ -98,7 +98,8 @@ public class TestExecutorService {
     ExecutorStatus status = executor.getStatus();
     assertTrue(status.queuedEvents.isEmpty());
     assertEquals(5, status.running.size());
-    checkStatusDump(status);
+    Waiter.waitFor(mockedServer.getConfiguration(), 10000,
+      () -> checkStatusDump(executor.getStatus()));
 
     // Now interrupt the running Executor
     synchronized (lock) {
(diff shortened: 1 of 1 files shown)


Sigma-Ma:
Hi @Apache9 @virajjasani, could one of you please approve the pending workflows and review this small test-only fix when convenient?

HBASE-30317 fixes a synchronization race in `TestExecutorService`. Reaching `counter == 5` does not guarantee that the handlers have entered `lock.wait()`, so the status dump can still capture RUNNABLE threads under load. The patch uses `Waiter.waitFor` with fresh `ExecutorStatus` snapshots instead of adding a fixed sleep.

virajjasani:
The changes look good, let's wait for the github workflow results. Thank you @Sigma-Ma

Sigma-Ma:
Thanks for your review! @virajjasani

virajjasani (approved):

PR #8535, opened 2026-08-07
Implement GzipByteBuffDecompressor and GzipHFileDecompressionContext …

SaadASTheDev:
## Why

HBase's `ByteBuffDecompressor` interface enables block decompression directly from/to
`ByteBuffer` objects, avoiding intermediate byte array copies. GZIP had no implementation
of this interface — all GZIP decompression went through a byte array round-trip regardless
of whether the underlying buffers were on-heap or off-heap.

## What

Adds `GzipByteBuffDecompressor`, a `ByteBuffDecompressor` for GZIP that handles two cases:

- **Off-heap (direct ByteBuffers):** delegates to Hadoop's native `ZlibDirectDecompressor`
  with `GZIP_FORMAT`. Requires native zlib to be loaded; `canDecompress()` returns false
  if it isn't.
- **On-heap (heap ByteBuffers):** uses Java's `Inflater` in raw DEFLATE (`nowrap`) mode,
  skipping the 10-byte GZIP header manually and verifying the CRC32 and ISIZE trailer
  fields after inflation. The native `ZlibDirectDecompressor` requires a stable native
  memory address, which heap `ByteBuffer`s do not provide, making this fallback necessary.

Also adds:

- `GzipHFileDecompressionContext` — carries per-context configuration (e.g. whether
  ByteBuff decompression is allowed), wired in via `reinit()`
- Input-length validation — rejects inputs shorter than the minimum valid GZIP member
  (header + trailer = 18 bytes)

## Testing

- Off-heap → off-heap decompression
- On-heap → on-heap decompression
- CRC32 mismatch detection
- ISIZE mismatch detection
- Input too short
- Output buffer too small
- `canDecompress()` guard logic

### DIFF
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
@@ -66,7 +66,7 @@ public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
     if (inputDirect && outputDirect) {
       return decompressor != null;
     }
-    return true;
+    return !inputDirect && !outputDirect;
   }
 
   @Override
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
@@ -1,0 +1,129 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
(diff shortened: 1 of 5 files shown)


hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,218 @@
...
+ * Glue for ByteBuffDecompressor on top of Hadoop's native
+ * {@link ZlibDecompressor.ZlibDirectDecompressor}.
+ */
+@InterfaceAudience.Private
+public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
+
+  private static final int GZIP_HEADER_LENGTH = 10;
+  private static final int GZIP_TRAILER_LENGTH = 8;
+
+  @Nullable
```
SaadASTheDev, line 41: These are the Singleton objects reused across decompress calls — one per decompression path, Allocated once per buffer, to avoid GC overhead

PR #8541, opened 2026-08-07
HBASE-30321: Implement GzipByteBuffDecompressor with on-heap and off-heap decompression paths

SaadASTheDev:
## Why

HBase's `ByteBuffDecompressor` interface enables block decompression directly from/to
`ByteBuffer` objects, avoiding intermediate byte array copies. GZIP had no implementation
of this interface — all GZIP decompression went through a byte array round-trip regardless
of whether the underlying buffers were on-heap or off-heap.

## What

Adds `GzipByteBuffDecompressor`, a `ByteBuffDecompressor` for GZIP that handles Direct Buffer cases:

- **Off-heap (direct ByteBuffers):** delegates to Hadoop's native `ZlibDirectDecompressor`
  with `GZIP_FORMAT`. Requires native zlib to be loaded; `canDecompress()` returns false
  if it isn't.

Also adds:

- `GzipHFileDecompressionContext` — carries per-context configuration (e.g. whether
  ByteBuff decompression is allowed), wired in via `reinit()`
- Input-length validation — rejects inputs shorter than the minimum valid GZIP member
  (header + trailer = 18 bytes)

## Testing

- Off-heap → off-heap decompression
- CRC32 mismatch detection
- ISIZE mismatch detection
- Input too short
- Output buffer too small
- `canDecompress()` guard logic
- Multiple Member payloads

### DIFF
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
@@ -27,7 +27,8 @@ import org.apache.yetus.audience.InterfaceAudience;
 
 /**
  * Glue for ByteBuffDecompressor on top of Hadoop's native
- * {@link ZlibDecompressor.ZlibDirectDecompressor}.
+ * {@link ZlibDecompressor.ZlibDirectDecompressor}. Only direct-to-direct decompression is
+ * supported, which is zero-copy; callers with on-heap buffers fall back to the stream path.
  */
 @InterfaceAudience.Private
 public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
@@ -35,14 +36,8 @@ public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
   private static final int GZIP_HEADER_LENGTH = 10;
   private static final int GZIP_TRAILER_LENGTH = 8;
 
-  // Size of the on-heap ZlibDecompressor's internal staging buffers
-  private static final int HEAP_DECOMPRESSOR_BUFFER_SIZE = 64 * 1024;
-
   @Nullable
   private final ZlibDecompressor.ZlibDirectDecompressor decompressor;
-
-  @Nullable
-  private final ZlibDecompressor heapDecompressor;
 
   private boolean allowByteBuffDecompression;
 
@@ -50,10 +45,6 @@ public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
     decompressor = nativeZlibLoaded
       ? new ZlibDecompressor.ZlibDirectDecompressor(ZlibDecompressor.CompressionHeader.GZIP_FORMAT,
         0)
-      : null;
-    heapDecompressor = nativeZlibLoaded
-      ? new ZlibDecompressor(ZlibDecompressor.CompressionHeader.GZIP_FORMAT,
-        HEAP_DECOMPRESSOR_BUFFER_SIZE)
       : null;
     allowByteBuffDecompression = true;
   }
@@ -66,15 +57,9 @@ public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
     if (!(output instanceof SingleByteBuff) || !(input instanceof SingleByteBuff)) {
       return false;
     }
-    boolean inputDirect = input.nioByteBuffers()[0].isDirect();
-    boolean outputDirect = output.nioByteBuffers()[0].isDirect();
-    if (inputDirect && outputDirect) {
-      return decompressor != null;
-    }
-    if (inputDirect != outputDirect) {
-      return false;
-    }
-    return !inputDirect && !outputDirect && heapDecompressor != null;
+    // Only direct-to-direct decompression is supported.
+    return input.nioByteBuffers()[0].isDirect() && output.nioByteBuffers()[0].isDirect()
+      && decompressor != null;
   }
 
   @Override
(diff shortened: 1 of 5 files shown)


hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,220 @@
...
+ * Glue for ByteBuffDecompressor on top of Hadoop's native
+ * {@link ZlibDecompressor.ZlibDirectDecompressor}.
+ */
+@InterfaceAudience.Private
+public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
+
+  private static final int GZIP_HEADER_LENGTH = 10;
+  private static final int GZIP_TRAILER_LENGTH = 8;
+
+  @Nullable
```
SaadASTheDev, line 39: These are the Singleton objects reused across decompress calls, Allocated once per buffer, to avoid GC overhead

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,220 @@
...
+
+  GzipByteBuffDecompressor(boolean nativeZlibLoaded) {
+    decompressor = nativeZlibLoaded
+      ? new ZlibDecompressor.ZlibDirectDecompressor(ZlibDecompressor.CompressionHeader.GZIP_FORMAT,
+        0)
+      : null;
+    allowByteBuffDecompression = true;
+  }
+
+  @Override
```
SaadASTheDev, line 52: The flow for wether we can decompress is 
1: verify config via context to ensure we havent disabled this feature before we try to decompress via bytebuff
2. we only support singleByteBuff decompression for now
3. Since Zlib is a JNI we need to ensure that its properly instantiated if and only if the buffers we are trying to decompress are direct, so if both are direct we ensure the decompressor is loaded
4. then we return if both I/O are not direct, if true it uses the inflater pathway

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,220 @@
...
+        throw new IllegalStateException(
+          "GzipByteBuffDecompressor#decompress() was called with direct buffers but Hadoop's "
+            + "native zlib library is not loaded, this should never happen since "
+            + "canDecompress() would have returned false");
+      }
+      return decompressOffHeap(nioInput, nioOutput, inputLen);
+    }
+    return decompressOnHeap(nioInput, nioOutput, inputLen);
+  }
+
```
SaadASTheDev, line 84: Respective off heap decompress which uses the zlib library and verifies the trailer on its own

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,220 @@
...
+    }
+  }
+
+  private static long readLittleEndianUInt32(byte[] data, int offset) {
+    return (data[offset] & 0xFFL) | ((data[offset + 1] & 0xFFL) << 8)
+      | ((data[offset + 2] & 0xFFL) << 16) | ((data[offset + 3] & 0xFFL) << 24);
+  }
+
+  @Override
+  public void reinit(@Nullable Compression.HFileDecompressionContext newHFileDecompressionContext) {
```
SaadASTheDev, line 115: We load in a new context at every new Hfile, it would be too expensive to do it at every block

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,220 @@
...
+@InterfaceAudience.Private
+public class GzipByteBuffDecompressor implements ByteBuffDecompressor {
+
+  private static final int GZIP_HEADER_LENGTH = 10;
+  private static final int GZIP_TRAILER_LENGTH = 8;
+
+  @Nullable
+  private final ZlibDecompressor.ZlibDirectDecompressor decompressor;
+
+  private final Inflater inflater = new Inflater(true);
```
charlesconnell: For consistency, I suggest you use the top-level ZlibDecompressor to handle on-heap

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,220 @@
...
+      } catch (IOException e) {
+        throw new IOException("Invalid gzip stream: " + e.getMessage(), e);
+      }
+      if (nioOutput.remaining() == outputRemainingBefore && !decompressor.finished()) {
+        if (!nioOutput.hasRemaining()) {
+          throw new IOException("Output buffer is too small for the decompressed gzip stream");
+        }
+        throw new IOException("Unexpected end of gzip stream");
+      }
+    }
```
charlesconnell: Is there a reason this needs to loop? Why would the decompressor need multiple attempts?

hbase-common/src/test/java/org/apache/hadoop/hbase/io/compress/TestGzipByteBuffDecompressor.java
```diff
@@ -0,0 +1,511 @@
...
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+@Tag(SmallTests.TAG)
+public class TestGzipByteBuffDecompressor {
+
+  /*
+   * "HBase is fun to use and very fast" compressed as a single gzip member via GZIPOutputStream,
+   * matching the framing that ReusableStreamGzipCodec produces on the compression side.
+   */
```
charlesconnell: Are there are situations where a gzip input payload would contain multiple members?

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
SaadASTheDev: Yea It doesnt need to loop. I was under the impression the zlib decompressor only follows a streaming, chunk-based pattern, where a block gets decompressed in chunks. Thats only true if we don't pre size the output buffer, going to fix this, Thanks

hbase-common/src/main/java/org/apache/hadoop/hbase/io/compress/GzipByteBuffDecompressor.java
SaadASTheDev: Will do, wasn't familiar with the top level decompressor but that simplifies things, Thanks!

hbase-common/src/test/java/org/apache/hadoop/hbase/io/compress/TestGzipByteBuffDecompressor.java
SaadASTheDev: There are situations, the payload will contain multiple members but we enclose over ours via: InputStart + InputLen, but we should test to make sure we did it correctly

charlesconnell:
I looked into how ZlibDecompressor supports decompression of on-heap byte arrays. It's not very efficient because it copies the bytes into a ByteBuffer. I think for this PR let's focus on where we can achieve zero-copy, so please skip support for on-heap.

charlesconnell (approved):

PR #8547, opened 2026-08-11
HBASE-30323 [RSGroup] Forward-port HBASE-22658 to branch-2

Umeshkumar9414:
RegionMover.unloadRegions() previously picked destination servers from all online RegionServers regardless of RSGroup membership, potentially trying to move regions out of their assigned group during server decommission, althoug HMaster prevents it.

This is a branch-2-compatible port of HBASE-22740. On master/branch-3 the fix used admin.getRSGroup() which is integrated directly into the Admin interface (via HBASE-22971). On branch-2 RSGroup remains an optional coprocessor module (hbase-rsgroup), so we:

1. Check whether RSGroups are enabled via RSGroupTableAccessor.isRSGroupsEnabled(conn) (hbase-client) to skip the RSGroup path on clusters that do not use RSGroups.
2. Look up the server's RSGroupInfo by reading the hbase:rsgroup system table directly via RSGroupTableAccessor.getAllRSGroupInfo(conn), this avoids a circular dependency between hbase-server and hbase-rsgroup.
3. Filter the destination server list to only servers in the same RSGroup (filterRSGroupServers); filtering is always applied by membership — the DEFAULT_GROUP short-circuit that could leak regions across groups is removed.
4. If RSGroups are enabled but the server cannot be matched to any group, throw IOException rather than falling back to treating every online server as a valid destination — that fallback would defeat RSGroup isolation. This can happen from hostname-vs-IP address-form mismatches (HBASE-27304).

Tests:
- TestRegionMoverWithRSGroupEnable (hbase-rsgroup): 5-node mini cluster integration test with RSGroupAdminEndpoint enabled. Verifies that unloading a non-default-group server places all regions exclusively on the remaining server in that group (positive assertion) and that no default-group server receives any of those regions (isolation assertion). The decommission target is chosen by checking which server actually hosts a region, since moveTableRegionsToGroup() places regions via randomAssignment(). Also tests the guard path: unloading a default-group server succeeds end-to-end when RSGroups are enabled, with destinations filtered to the default group.
- TestRegionMoverFilterRSGroupServers (hbase-server): unit tests for filterRSGroupServers() — default group returns full server list, non-default group filters to members only, no-match group returns empty.


mnpoonia (approved):

virajjasani (approved):

hbase-rsgroup/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRegionMoverWithRSGroupEnable.java
```diff
@@ -0,0 +1,239 @@
...
+        LOG.info("Unloading default-group server {}", decommission.getHostname());
+        rm.unload();
+      }
+
+      // After unload, the decommissioned server must hold no regions of the default table.
+      HRegionServer decommRS = TEST_UTIL.getMiniHBaseCluster().getRegionServerThreads().stream()
+        .map(JVMClusterUtil.RegionServerThread::getRegionServer)
+        .filter(rs -> rs.getServerName().equals(defaultSN)).findFirst().get();
+      assertEquals(0, decommRS.getRegions(defaultTable).size(),
+        "Decommissioned default-group server must hold no regions after unload");
```
sanjeet006py, line 257: Can we also please add assertion that default group regions don't land on `test` RS group servers?

hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMoverFilterRSGroupServers.java
```diff
@@ -0,0 +1,119 @@
...
+
+      Collection<ServerName> result = rm.filterRSGroupServers(defaultGroup, allServers);
+      assertEquals(allServers.size(), result.size());
+      assertTrue(result.containsAll(allServers));
+    }
+  }
+
+  /** A non-default group with one member must return only that member. */
+  @Test
+  public void testNonDefaultGroupFiltersToMembers() throws Exception {
```
sanjeet006py, line 114: Can we add a test case where there is default RS group and some test RS group. And, assertion is that output of filter for default RS group should not have any server of test RS group. Please correct me if my understanding is wrong.

sanjeet006py (commented):
How about adding an IT test case for scenario when all servers in test RS group go down and fallback config is enabled (`hbase.rsgroup.fallback.enable = true`), then regions from test RS group do move to default RS group servers on calling RegionMover.unload().

Umeshkumar9414:
> How about adding an IT test case for scenario when all servers in test RS group go down and fallback config is enabled (`hbase.rsgroup.fallback.enable = true`), then regions from test RS group do move to default RS group servers on calling RegionMover.unload().

There is already test written for that, 'TestRSGroupsFallback'.

hbase-rsgroup/src/test/java/org/apache/hadoop/hbase/rsgroup/TestRegionMoverWithRSGroupEnable.java
Umeshkumar9414: added the assertion in testUnloadDefaultGroupServerWithRSGroupEnabled

hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMoverFilterRSGroupServers.java
Umeshkumar9414: added in testNonDefaultGroupFiltersToMembers

sanjeet006py (approved):

virajjasani:
The PR is ready to merge. We just need to co-ordinate when to merge this, I think it will be better to merge this after your fix on master branch is first merged and backported to branch-2 and others. Please let me know if there is any concern, I am fine either way.

apurtell:
@virajjasani Copilot found at least one issue (the first one) that should be resolved before merge.

apurtell:
The move of `RSGroupAdmin.proto` from `hbase-rsgroup` to `hbase-protocol` is not necessary and the way you've left the state of the POM in `hbase-rsgroup` the protobuf compilation still happens there but now there's no input files. I guess it works?? but is messy. 

We move RSGroups out of the separate module into core in HBase 3 and up, so there's that, but the way we hedged RSGroups in branch-2 puts everything else in `hbase-rsgroup`. It's fine and even customary to have protos outside of `hbase-protos`. We have the protos for REST broken out into `hbase-rest`, for example, so this is maybe a smell. 

This is not a dealbreaker, but there is an alternative that does not require it. Consider just using `RSGroupTableAccessor` from `hbase-client`. Skip the coprocessor RPC and read the `hbase:rsgroup` system table directly via the accessor that already exists in `hbase-client`. Something like:

```java
private RSGroupInfo getRSGroupInfo(String host, int port) throws IOException {
  if (!RSGroupTableAccessor.isRSGroupsEnabled(conn)) {
    return null;
  }
  Address addr = Address.fromParts(host, port);
  for (RSGroupInfo info : RSGroupTableAccessor.getAllRSGroupInfo(conn)) {
    if (info.containsServer(addr)) {
      return info;
    }
  }
  return null;
}
```

I also like this approach better because the invocation of the coprocessor is brittle. It works today but any subclass or wrapper would break it, if someone is extending rsgroups and implementing their own enhanced coprocessors. Unlikely but not impossible.

Umeshkumar9414:
Pushed 3dcff721ab addressing the remaining review feedback:

**@apurtell's suggestion** — switched `getRSGroupInfo()` to read `hbase:rsgroup` directly via `RSGroupTableAccessor` (from `hbase-client`) instead of the `RSGroupAdminEndpoint` coprocessor RPC. This let me move `RSGroupAdmin.proto` back to `hbase-rsgroup`, so `hbase-protocol` no longer carries it — the messy proto relocation is gone.

**Copilot findings:**
- `getRSGroupInfo()` now throws `IOException` instead of returning `null` when RSGroups are enabled but the server can't be matched to any group. Previously that case fell through to treating every online server as a valid destination, which defeats RSGroup isolation during decommission.
- `testUnloadRegionsRespectsRSGroup` now picks the decommission target by checking which `rsservers` member actually hosts a `TABLE_NAME` region, rather than assuming `rsservers.get(0)`. `moveTableRegionsToGroup()` places regions via `randomAssignment()`, so the old assumption could let the test pass without exercising the move/filter path at all.
- Fixed the Javadoc on `testUnloadDefaultGroupServerWithRSGroupEnabled` — it said regions "spread across all available servers (not filtered)" but the test actually asserts destinations are filtered to the default group.

Verified: `TestRegionMoverFilterRSGroupServers` (3/3) and `TestRegionMoverWithRSGroupEnable` (2/2) pass, checkstyle and spotbugs clean on both `hbase-server` and `hbase-rsgroup`.

Umeshkumar9414:
> The move of `RSGroupAdmin.proto` from `hbase-rsgroup` to `hbase-protocol` is not necessary and the way you've left the state of the POM in `hbase-rsgroup` the protobuf compilation still happens there but now there's no input files. I guess it works?? but is messy.
> 
> We move RSGroups out of the separate module into core in HBase 3 and up, so there's that, but the way we hedged RSGroups in branch-2 puts everything else in `hbase-rsgroup`. It's fine and even customary to have protos outside of `hbase-protos`. We have the protos for REST broken out into `hbase-rest`, for example, so this is maybe a smell.
> 
> This is not a dealbreaker, but there is an alternative that does not require it. Consider just using `RSGroupTableAccessor` from `hbase-client`. Skip the coprocessor RPC and read the `hbase:rsgroup` system table directly via the accessor that already exists in `hbase-client`. Something like:
> 
> ```java
> private RSGroupInfo getRSGroupInfo(String host, int port) throws IOException {
>   if (!RSGroupTableAccessor.isRSGroupsEnabled(conn)) {
>     return null;
>   }
>   Address addr = Address.fromParts(host, port);
>   for (RSGroupInfo info : RSGroupTableAccessor.getAllRSGroupInfo(conn)) {
>     if (info.containsServer(addr)) {
>       return info;
>     }
>   }
>   return null;
> }
> ```
> 
> I also like this approach better because the invocation of the coprocessor is brittle. It works today but any subclass or wrapper would break it, if someone is extending rsgroups and implementing their own enhanced coprocessors. Unlikely but not impossible.

Thanks a lot. @apurtell . I missed this.

Umeshkumar9414:
I checked both are clean cherry-pick to 2.5 and 2.6.

virajjasani:
Triggered github workflow, please keep an eye if we get any non-flaky test failures

virajjasani (approved):

PR #8552, opened 2026-08-17
HBASE-30331 Remove RegionMover.filterRSGroupServers's short-circuit on default RSGroup

Umeshkumar9414:
## Problem

`RegionMover.filterRSGroupServers(RSGroupInfo, Collection<ServerName>)` short-circuited whenever the RSGroup being filtered was named `"default"`:

```java
if (rsgroup.getName().equals(RSGroupInfo.DEFAULT_GROUP)) {
  return onlineServers;
}
```

`RSGroupInfoManagerImpl` computes "default" group membership dynamically as *online servers minus servers claimed by other groups* — it is never a fixed, all-inclusive set. Because of the short-circuit, `RegionMover` treated **every online server as a valid destination** whenever it filtered against a group literally named `default`, regardless of whether those servers actually belonged to that group. This broke RSGroup isolation: unloading a server from the default group (or any group whose real name happened to be `default`) could try to move regions onto servers that belong to unrelated, non-default RSGroups.

## Fix

Remove the short-circuit so `filterRSGroupServers` always filters `onlineServers` down to the group's real, master-reported membership (`rsgroup.containsServer(...)`), regardless of the group's name.

## Testing

- **New unit tests** — `TestRegionMoverFilterRSGroupServers`: exercises `filterRSGroupServers` directly against a real mini-cluster. All `RSGroupInfo` instances are read back from the master via `admin.getRSGroup(...)` after real `moveServersToRSGroup` calls — none are hand-constructed — so the "default" group's membership genuinely shrinks once a server is moved out, the same way it does in production.
  - `testDefaultGroupFiltersToActualMembers` — reproduces the bug: after moving a server out of `default` into a custom group, filtering against the master-computed `default` group must return only the server that remains, not every online server.
  - `testNonDefaultGroupFiltersToMembers` — sanity check for an ordinary custom group.
  - `testGroupMemberAbsentFromOnlineServersReturnsEmpty` — a group's member that isn't present in the `onlineServers` snapshot must not be returned.
- **Strengthened `TestRegionMoverWithRSGroupEnable`** (existing end-to-end `RegionMover` + RSGroup test) to match the coverage of the equivalent test on branch-2 (`hbase-rsgroup`'s `TestRegionMoverWithRSGroupEnable`, see #8547), ported to this branch's `Admin`-based RSGroup API:
  - `testUnloadRegions` now also asserts that **no default-group server** ends up holding any region of the unloaded table, not just that the surviving test-group server received all of them.
  - Added `testUnloadDefaultGroupServerWithRSGroupEnabled`: unloads a server that belongs to the *default* RSGroup (with RSGroups enabled) and asserts both that the decommissioned server ends up empty and that no test-group server receives any of its regions.
  - Added `@AfterEach` cleanup (remove the test RSGroup, move its servers back to default) since `@BeforeEach` now runs before two `@Test` methods per class instance.
- Verified end-to-end against a real local multi-RegionServer cluster with a custom RSGroup: reproduced the bug (a `RegionMover` unload got a region stuck and eventually failed when servers from another group were treated as valid destinations) with the short-circuit present, then confirmed correct behavior with the fix.
- `mvn test -pl hbase-server -Dtest=TestRegionMoverFilterRSGroupServers,TestRegionMoverWithRSGroupEnable` passes.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/util/RegionMover.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/util/RegionMover.java
@@ -527,9 +527,6 @@ public class RegionMover extends AbstractHBaseTool implements Closeable {
   @InterfaceAudience.Private
   Collection<ServerName> filterRSGroupServers(RSGroupInfo rsgroup,
     Collection<ServerName> onlineServers) {
-    if (rsgroup.getName().equals(RSGroupInfo.DEFAULT_GROUP)) {
-      return onlineServers;
-    }
     List<ServerName> serverLists = new ArrayList<>(rsgroup.getServers().size());
     for (ServerName server : onlineServers) {
       Address address = Address.fromParts(server.getHostname(), server.getPort());
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMoverFilterRSGroupServers.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/util/TestRegionMoverFilterRSGroupServers.java
@@ -1,0 +1,189 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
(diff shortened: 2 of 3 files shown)


sanjeet006py (approved):

virajjasani (approved):

Umeshkumar9414:
Only needed for master and 3.x Not for 2.x and 2.x release. For 2.x and 2.x release branches this is done by https://github.com/apache/hbase/pull/8547.

PR #8555, opened 2026-08-20
HBASE-30332 Preserve QueryMetrics for empty results

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30332

This pull request preserves `QueryMetrics` when an empty `Result` is converted to and from protobuf.

The changes:

- Preserve metrics in both `ProtobufUtil.toResult` and `ProtobufUtil.toResultNoData` when the result contains no cells.
- Continue reusing the cached empty protobuf instances when metrics are not present.
- Create a per-response empty `Result` when deserializing a protobuf containing metrics, instead of attaching metrics to the shared cached `EMPTY_RESULT`.
- Add a focused regression test covering both serialization and both deserialization paths.

No protobuf schema change is required.

When a `Get` with query metrics enabled requests a row that does not exist, the RegionServer still computes and attaches `QueryMetrics`, including `blockBytesScanned`, to the empty `Result`.

However, the protobuf conversion methods returned cached empty protobuf instances before copying the metrics. As a result, the client received `result.getMetrics() == null`.

Simply preserving the metrics during serialization would also expose a second problem: the deserialization path could attach metrics to a shared cached empty `Result`, potentially leaking those metrics into later requests where metrics were not enabled.

These changes preserve the metrics while ensuring that metrics-bearing empty results are not shared between requests.

The following focused test was run:

```bash
mvn -pl hbase-client -am \
  -Dtest=TestProtobufUtil \
  test

### DIFF
--- a/hbase-client/src/main/java/org/apache/hadoop/hbase/shaded/protobuf/ProtobufUtil.java
+++ b/hbase-client/src/main/java/org/apache/hadoop/hbase/shaded/protobuf/ProtobufUtil.java
@@ -1428,7 +1428,10 @@ public final class ProtobufUtil {
 
     ExtendedCell[] cells = ClientInternalHelper.getExtendedRawCells(result);
     if (cells == null || cells.length == 0) {
-      return result.isStale() ? EMPTY_RESULT_PB_STALE : EMPTY_RESULT_PB;
+      ClientProtos.Result emptyResult = result.isStale() ? EMPTY_RESULT_PB_STALE : EMPTY_RESULT_PB;
+      return result.getMetrics() == null
+        ? emptyResult
+        : emptyResult.toBuilder().setMetrics(toQueryMetrics(result.getMetrics())).build();
     }
 
     ClientProtos.Result.Builder builder = ClientProtos.Result.newBuilder();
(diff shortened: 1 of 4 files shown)


charlesconnell (approved):

hbase-client/src/main/java/org/apache/hadoop/hbase/shaded/protobuf/ProtobufUtil.java
guluo2016: The metrics maybe lost if return here for this case. Do we need to handle it here as well?

Sigma-Ma: Good catch. You're right — an existence-only Get can produce a Result with both exists and QueryMetrics, and this early return drops the metrics.

I checked the related paths and found the same problem in toResultNoData, both protobuf-to-Result overloads when hasExists() is set, and the standalone Get response path in RSRpcServices, which currently serializes only the existence boolean.

I'll update these paths and add an existence-only round-trip test. Table#exists itself will remain unchanged since that API returns only a boolean. 

Thanks for catching this.

Sigma-Ma: Hi, I fixed the same issue in toResultNoData, both protobuf-to-Result paths, and the standalone Get response in RSRpcServices. 

I also added coverage for exists=true/false round trips and single/multi existence-only Gets.

guluo2016 (approved):
Thanks a lot

guluo2016:
Please raise a PR for branch-2, thanks @Sigma-Ma

Sigma-Ma:
> Please raise a PR for branch-2, thanks @Sigma-Ma

Sure, I’ll prepare a separate PR for branch-2 and share the link here once it’s ready. Thanks!

Sigma-Ma:
I’ve opened #8565 for branch-2. Thanks for the reminder @guluo2016.

PR #8557, opened 2026-08-20
HBASE-30288 Fix WALPlayer bulk output table mappings

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30288

This PR makes WALPlayer honor source-to-target table mappings when generating bulk-output HFiles.

The bulk mapper now builds the mapping from `TABLES_KEY` and `TABLE_MAP_KEY`, keeps source table names for WAL filtering, and emits mapped target table names in multi-table composite keys. The bulk job is also configured with target table descriptors and region locators.

The existing bulk-load test now uses distinct source and target tables, with the target table pre-split using a different region layout, and verifies that the generated HFiles can be loaded into the mapped target table.

Before this change, the bulk-output path ignored the target table mappings. `WALKeyValueMapper` emitted source table names, and `createSubmittableJob` configured `HFileOutputFormat` with source table metadata.

As a result, HFiles could be generated under the wrong table path and with region boundaries or table descriptors belonging to the source table, preventing them from being correctly bulk-loaded into the intended target table.

```bash
mvn -pl hbase-mapreduce -am \
  '-Dtest=TestWALPlayer#testWALPlayerBulkLoadWithTableMappingAndOverriddenTimestamps' \
  test

mvn -pl hbase-mapreduce -am -Dtest=TestWALPlayer -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALPlayer.java
+++ b/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALPlayer.java
@@ -21,11 +21,8 @@ import java.io.IOException;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.TreeMap;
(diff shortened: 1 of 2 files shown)


Sigma-Ma:
Hi @hgromer, when you have time, could you please take a look at this PR? Thanks!

PR #8558, opened 2026-08-21
HBASE-30336 An extra empty MOB region dir is created under the table dir when restoring a MOB table from snapshot

guluo2016:
Details see: HBASE-30336


liuxiaocs7 (approved):
Thanks for the fix!

guluo2016:
Thanks for reviewing @liuxiaocs7

PR #8560, opened 2026-08-22
HBASE-30337 Miscellaneous improvements on rpc checks

Apache9:


virajjasani:
The changes look good, but github workflows have failures. Is it yetus related issue?

mnpoonia (approved):
This has really hardened the auth of RPC

very small nit
``` 
Adding throws IOException is binary-compatible but source-incompatible for coprocessor callers that invoke these hooks or MasterObserver.super.
```
Maybe a follow up in phoenix( 😄  @virajjasani ) to handle it, ofcourse which is out of scope of this PR

Apache9:
Add IOException in method signature usually does not affect CP users, as you are free to not adding the throws signature in your implementation.

But yes, this is a incompatible change, will add to the release note.

Thanks for reviewing! @mnpoonia

virajjasani (approved):

hbase-server/src/main/java/org/apache/hadoop/hbase/coprocessor/MasterObserver.java
```diff
@@ -1834,12 +1834,34 @@ default void postRevoke(ObserverContext<MasterCoprocessorEnvironment> ctx,
    * @param family    the table column family, null if don't get table family permission
    * @param qualifier the table column qualifier, null if don't get table qualifier permission
    * @throws IOException if something went wrong
+   * @deprecated Since 2.5.17,2.6.8, 2.7.0, 3.0.1 and 3.10, will be removed in 4.0.0. Use
```
liuxiaocs7: `3.10` should be `3.1.0`?

liuxiaocs7 (approved):

PR #8563, opened 2026-08-23
HBASE-29640 Add metric for oldest procedure age

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-29640

This pull request adds an `oldestProcedureAge` gauge to the existing `Master,sub=Procedure` metrics source. The gauge reports the age in milliseconds of the oldest unfinished master procedure, calculated from its submitted time.

The metric returns zero when there are no active procedures or when the master procedure executor is not yet available. Finished procedures are ignored, and negative values caused by clock adjustments are clamped to zero.

Operators currently have no direct metric for identifying how long the oldest master procedure has been active. Exposing this value makes it easier to detect stuck or unexpectedly long-running procedures and to configure monitoring and alerting around procedure execution.

A new `TestOldestProcedureMetrics` test verifies:

- Selection of the oldest unfinished procedure.
- Exclusion of finished procedures.
- Export of the `oldestProcedureAge` gauge.
- The zero value when there are no active procedures.
- The zero value when the procedure executor is unavailable.
- Protection against negative values after clock adjustments.

```bash
mvn -pl hbase-server -am \
  -Dtest=TestOldestProcedureMetrics \
  -Dsurefire.failIfNoSpecifiedTests=false test

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterWrapperImpl.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterWrapperImpl.java
@@ -29,8 +29,11 @@ import org.apache.hadoop.hbase.ServerName;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.RegionInfo;
 import org.apache.hadoop.hbase.client.TableDescriptor;
+import org.apache.hadoop.hbase.master.procedure.MasterProcedureEnv;
+import org.apache.hadoop.hbase.procedure2.ProcedureExecutor;
 import org.apache.hadoop.hbase.quotas.QuotaObserverChore;
 import org.apache.hadoop.hbase.quotas.SpaceQuotaSnapshot;
+import org.apache.hadoop.hbase.util.EnvironmentEdgeManager;
 import org.apache.hadoop.hbase.util.PairOfSameType;
 import org.apache.hadoop.hbase.zookeeper.ZKWatcher;
 import org.apache.yetus.audience.InterfaceAudience;
@@ -178,6 +181,18 @@ public class MetricsMasterWrapperImpl implements MetricsMasterWrapper {
   @Override
   public long getNumWALFiles() {
     return master.getNumWALFiles();
+  }
+
+  @Override
+  public long getOldestProcedureAge() {
+    ProcedureExecutor<MasterProcedureEnv> procedureExecutor = master.getMasterProcedureExecutor();
+    if (procedureExecutor == null) {
+      return 0L;
+    }
+    long now = EnvironmentEdgeManager.currentTime();
+    return procedureExecutor.getActiveProceduresNoCopy().stream()
+      .filter(procedure -> !procedure.isFinished())
+      .mapToLong(procedure -> Math.max(0L, now - procedure.getSubmittedTime())).max().orElse(0L);
   }
 
   @Override
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterWrapperImpl.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterWrapperImpl.java
@@ -30,6 +30,7 @@ import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.RegionInfo;
 import org.apache.hadoop.hbase.client.TableDescriptor;
 import org.apache.hadoop.hbase.master.procedure.MasterProcedureEnv;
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

File: hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterWrapperImpl.java [hunk 1]
```diff
@@ -180,6 +183,18 @@ public long getNumWALFiles() {
     return master.getNumWALFiles();
   }
 
+  @Override
+  public long getOldestProcedureAge() {
```
Umeshkumar9414: OldestAge from all the procedures doesn't give much info IMHO. As there are procedure like LockProcedure that might run for a longer time compared to other procedures (In other words avarage time taken differs based on the type of procedure.). It would be great if we can fetch or expose metric/log for oldestAge prcedure of each type.

Sigma-Ma: Good point, thanks. A LockProcedure can naturally live much longer, so the global max may hide a stuck procedure of another type.

I updated the patch to calculate the oldest age per procedure class and expose gauges such as `oldestProcedureAge_LockProcedure` and `oldestProcedureAge_ServerCrashProcedure`. I kept `oldestProcedureAge` as the overall max for the original use case.

The test now covers two procedure types, multiple procedures of the same type, and finished procedures.

mnpoonia (approved):
This is a much needed metrics. Thanks @Sigma-Ma

Sigma-Ma:
@Umeshkumar9414 I've pushed the per-type changes from your review. Could you take another look when you have a chance?

@apurtell, I'd also appreciate your review since this touches the master procedure metrics path. Thanks.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/MetricsMasterWrapperImpl.java
```diff
@@ -180,6 +184,23 @@ public long getNumWALFiles() {
...
 
+  @Override
+  public Map<String, Long> getOldestProcedureAgeByType() {
+    ProcedureExecutor<MasterProcedureEnv> procedureExecutor = master.getMasterProcedureExecutor();
+    if (procedureExecutor == null) {
+      return Collections.emptyMap();
+    }
+    long now = EnvironmentEdgeManager.currentTime();
+    Map<String, Long> oldestProcedureAgeByType = new HashMap<>();
+    for (Procedure<MasterProcedureEnv> procedure : procedureExecutor.getActiveProceduresNoCopy()) {
```
Umeshkumar9414: Unsafe API used on a steady-state path (MetricsMasterWrapperImpl.java:195) getOldestProcedureAgeByType() iterates ProcedureExecutor.getActiveProceduresNoCopy(). That method's own javadoc: "Should only be used when starting up, where the procedure workers have not been started... Use getProcedures() ... for most cases."

Sigma-Ma: Thanks for catching this. 

The metrics collection path runs after the procedure workers have started, so getActiveProceduresNoCopy() isn't appropriate here. I've switched it to getProcedures() and kept the isFinished() filter because the copied list also contains retained completed procedures.

Umeshkumar9414 (approved):

virajjasani (approved):

virajjasani:
Do we have branch-2 PR?

Sigma-Ma:
> Do we have branch-2 PR?

Not yet, but I’m happy to open a branch-2 PR if needed.

virajjasani:
Yes please, at least branch-2 is always needed for majority PRs because branch-2 supports JDK 8 also whereas master/branch-3 only support JDK 17.

virajjasani:
Let me know whenever branch-2 PR is ready

Sigma-Ma:
> Let me know whenever branch-2 PR is ready

Sure. The usual workflow is to merge the master PR first, then open a separate backport PR for branch-2. I’ll let you know when that PR is ready.

Sigma-Ma:
> Let me know whenever branch-2 PR is ready

If you’d prefer to have the branch-2 PR ready before this one is merged, I can prepare it tonight.

virajjasani:
If you can prepare the PR, that would be great so that I can merge all together.

Sigma-Ma:
> If you can prepare the PR, that would be great so that I can merge all together.

@virajjasani The branch-2 PR is ready: #8689. The metrics tests pass on JDK 8, and Spotless passes too. Thanks for reviewing both.

virajjasani:
Thanks @Sigma-Ma, I will merge the PR soon

PR #8564, opened 2026-08-24
HBASE-30339 MobFileCleanupUtil loads store files from the wrong regio…

liuxiaocs7:
…n directory

- see: HBASE-30339


guluo2016 (approved):
Good catch. Thanks a lot!

PR #8566, opened 2026-08-25
HBASE-30327 Implement adaptive parallel seek in StoreScanner

ichsansaid:
## Summary

Introduces an adaptive parallel seek strategy in `StoreScanner` that gracefully handles thread pool saturation by dynamically switching between parallel and sequential execution based on available capacity.

JIRA: https://issues.apache.org/jira/browse/HBASE-30327

## Problem

When parallel seek is enabled, all StoreFileScanner seek operations are submitted to the `RS_PARALLEL_SEEK` thread pool with an unbounded queue. Under high concurrency, this can lead to:
- Queue buildup while scanner threads wait for their tasks
- Increased latency from blocking on full queues
- Thread starvation

## Solution

Adaptive parallel seek checks pool capacity before submission using a conservative approach (`queue.isEmpty() && activeCount < poolSize`). When the pool is saturated:
- Falls back to sequential seek on the calling thread
- Re-checks capacity after each sequential seek
- Opportunistically parallelizes when slots become available

## Changes

- Add private `getAvailableParallelSeekCapacity()` in `StoreScanner` (keeps change local)
- Add `adaptiveParallelSeek()` method with proper error handling:
  - Inline seek IOException waits for submitted handlers before propagating
  - Uses direct `scanner.seek()` for inline path (not ParallelSeekHandler)
  - Capacity counted for StoreFileScanners only
- Add configuration `hbase.storescanner.adaptive.parallel.seek.enable` (default: false)
- Add `adaptiveParallelSeekEnabled` field to `ScanInfo`

## Configuration

| Property | Default | Description |
|----------|---------|-------------|
| `hbase.storescanner.adaptive.parallel.seek.enable` | `false` | Enable adaptive parallel seek (requires `parallel.seek.enable=true`) |

## Testing

- Unit tests for adaptive behavior (sequential fallback, mixed mode, error propagation)
- Existing parallel seek tests unaffected

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
@@ -27,6 +27,7 @@ import java.util.NavigableSet;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.ThreadPoolExecutor;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.locks.ReentrantLock;
 import java.util.function.IntConsumer;
@@ -46,6 +47,7 @@ import org.apache.hadoop.hbase.client.IsolationLevel;
 import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.conf.ConfigKey;
 import org.apache.hadoop.hbase.executor.ExecutorService;
+import org.apache.hadoop.hbase.executor.ExecutorType;
 import org.apache.hadoop.hbase.filter.Filter;
 import org.apache.hadoop.hbase.ipc.RpcCall;
 import org.apache.hadoop.hbase.ipc.RpcServer;
@@ -95,6 +97,7 @@ public class StoreScanner extends NonReversedNonLazyKeyValueScanner
    * A flag that enables StoreFileScanner parallel-seeking
    */
   private boolean parallelSeekEnabled = false;
+  private boolean adaptiveParallelSeekEnabled = false;
   private ExecutorService executor;
   private final Scan scan;
   private final long oldestUnexpiredTS;
@@ -129,6 +132,8 @@ public class StoreScanner extends NonReversedNonLazyKeyValueScanner
   static final boolean LAZY_SEEK_ENABLED_BY_DEFAULT = true;
   public static final String STORESCANNER_PARALLEL_SEEK_ENABLE =
     "hbase.storescanner.parallel.seek.enable";
+  public static final String STORESCANNER_ADAPTIVE_PARALLEL_SEEK_ENABLE =
+    "hbase.storescanner.adaptive.parallel.seek.enable";
 
   /** Used during unit testing to ensure that lazy seek does save seek ops */
   private static boolean lazySeekEnabledGlobally = LAZY_SEEK_ENABLED_BY_DEFAULT;
@@ -232,6 +237,8 @@ public class StoreScanner extends NonReversedNonLazyKeyValueScanner
       RegionServerServices rsService = store.getHRegion().getRegionServerServices();
       if (rsService != null && scanInfo.isParallelSeekEnabled()) {
         this.parallelSeekEnabled = true;
+        this.adaptiveParallelSeekEnabled =
+          this.parallelSeekEnabled && scanInfo.isAdaptiveParallelSeekEnabled();
         this.executor = rsService.getExecutorService();
       }
     }
@@ -441,6 +448,8 @@ public class StoreScanner extends NonReversedNonLazyKeyValueScanner
             totalScannersSoughtBytes += PrivateCellUtil.estimatedSerializedSizeOf(c);
           }
         }
+      } else if (adaptiveParallelSeekEnabled) {
+        adaptiveParallelSeek(scanners, seekKey);
       } else {
         parallelSeek(scanners, seekKey);
       }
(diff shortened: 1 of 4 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
```diff
@@ -1271,6 +1277,135 @@ private void parallelSeek(final List<? extends KeyValueScanner> scanners, final
...
+      if (scanners.get(i) instanceof StoreFileScanner) {
+        latch.countDown();
+      }
+    }
+
+    try {
+      latch.await();
+    } catch (InterruptedException ie) {
+      throw (InterruptedIOException) new InterruptedIOException().initCause(ie);
+    }
```
Sigma-Ma: Could we avoid returning while submitted handlers are still using the scanners? On interruption, this method throws immediately, and the constructor cleanup may close the scanners before the worker seek finishes. The new interruption test currently demonstrates this by joining the caller before releasing the worker. I think we should finish waiting, restore the interrupt status, and then throw InterruptedIOException.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
```diff
@@ -1271,6 +1277,135 @@ private void parallelSeek(final List<? extends KeyValueScanner> scanners, final
...
+          index++;
+          break;
+        }
+        if (scanner instanceof StoreFileScanner) {
+          latch.countDown();
+        }
+        index++;
+      } else {
+        // Opportunistic parallel: submit batch up to available capacity
+        int batchEnd = Math.min(index + capacity, scannerCount);
```
Sigma-Ma: capacity represents available worker slots, but batchEnd advances over all scanners, including MemStore scanners that run inline. For example, with capacity 2 and two MemStore scanners first, no StoreFile scanner is submitted in that batch. Could we decrement capacity only when a StoreFile scanner is actually submitted?

ichsansaid: You're right. The current batchEnd = index + capacity counts all scanners including MemStore ones that run inline, so we can end up submitting fewer StoreFile tasks than the available capacity allows.

Ill do a changes for this

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
ichsansaid: Good point. The fix is to loop on latch.await() and catch InterruptedException without throwing immediately. This works because CountDownLatch.await() clears the interrupt flag when it throws InterruptedException ([Java docs](https://docs.oracle.com/en/java/javase/21/docs/api/java.base/java/util/concurrent/CountDownLatch.html)), so the next call will block normally until all submitted handlers finish. 

Once the latch reaches zero, we restore the interrupt status via Thread.currentThread().interrupt() and then throw InterruptedIOException. This ensures no scanner is closed while a worker thread is still using it.

One thing I noticed: 
The existing parallelSeek has the same behavior, it throws InterruptedIOException immediately on interruption without waiting for submitted handlers to finish. Should we fix that in this PR as well, or would you prefer to track it as a separate JIRA to keep this patch focused?

Sigma-Ma (changes requested):
Thanks for the update. The capacity accounting now handles MemStore scanners correctly, and the adaptive path waits for submitted handlers before propagating an interruption.

I still found an issue with latch accounting in the inline StoreFileScanner failure path, and the updated interruption test currently fails. I left the details inline.

Please also run mvn spotless:apply, as spotless:check still reports violations in StoreScanner.java and TestStoreScanner.java.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
```diff
@@ -1257,18 +1263,155 @@ private void parallelSeek(final List<? extends KeyValueScanner> scanners, final
...
+      int capacity = getAvailableParallelSeekCapacity();
+
+      if (capacity == 0) {
+        // Sequential fallback: process one scanner on calling thread
+        KeyValueScanner scanner = scanners.get(index);
+        try {
+          scanner.seek(kv);
+        } catch (IOException e) {
+          // Must wait for already-submitted handlers before propagating error
+          inlineSeekError = e;
```
Sigma-Ma, line 1336: Could we advance index before breaking here? This StoreFileScanner has already decremented the latch, but the cleanup loop starts from the same index and decrements it again. If an earlier submitted handler is still running, the latch may reach zero early and the caller can return while that handler is still using its scanner.

A test with a submitted blocking handler followed by an inline StoreFileScanner failure would cover this path.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
```diff
@@ -1257,18 +1263,155 @@ private void parallelSeek(final List<? extends KeyValueScanner> scanners, final
...
+    InterruptedIOException interruptedException = null;
+    while(true) {
+      try {
+        latch.await();
+        break;
+      } catch (InterruptedException ie) {
+        interruptedException = (InterruptedIOException) new InterruptedIOException().initCause(ie);
+      }
+    }
+    if (interruptedException != null) {
```
Sigma-Ma: Since parallelSeek now waits for handlers after interruption, could we restore the interrupt status before throwing, as the adaptive path does? CountDownLatch.await() clears it.

Alternatively, we could leave the existing parallelSeek behavior unchanged here and handle it in a separate JIRA.

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestStoreScanner.java
```diff
@@ -1303,4 +1314,543 @@ public void testGetFilesReadOnInitializationFailure() throws Exception {
...
+    });
+    seekThread.start();
+
+    // Give the thread time to submit the handler and start awaiting the latch
+    Waiter.waitFor(HBaseConfiguration.create(), 5000, () -> seekThread.getState() == Thread.State.WAITING
+      || seekThread.getState() == Thread.State.TIMED_WAITING);
+
+    // Interrupt the seeking thread
+    seekThread.interrupt();
+    seekThread.join(5000);
```
Sigma-Ma: Could we release scannerBlockLatch before the final join, then join and assert that the thread has exited? With the new wait-for-handler behavior, this join times out by design, and the test reads caughtError immediately after releasing the worker. I can reproduce this test failing with caughtError == null.

It would also be useful to capture and verify the seek thread's interrupt status before it exits.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
ichsansaid: I'd like to open a discussion on the broader behavior. With the new wait-for-handler approach, an interrupted parallelSeek now blocks until all background handlers finish before throwing. This raises a concern around cancellation latency, I'm not certain what triggers the interrupt in practice (client cancellation, RPC timeout, region close?), but if it is used as a stop signal, waiting for handlers doing disk/HDFS I/O could delay the response noticeably. Could you clarify what the typical interrupt sources are here?

The motivation for waiting is resource safety, throwing immediately while handlers still hold references to the scanner risks use-after-free or resource leaks if the caller closes the scanner after catching the exception.

Worth noting that this concern already exists in adaptiveParallelSeek, which has the same wait-for-latch pattern. So a proper fix (e.g. interrupting handler threads before waiting, or adding a bounded timeout) would need to be applied to both paths consistently.

A possible middle ground would be to interrupt the handler threads before waiting on the latch, so they can bail out quickly, or add a bounded timeout to the wait. Alternatively, as you suggested, we could leave parallelSeek behavior unchanged and handle this in a separate JIRA, which might actually be the cleaner option since any fix should cover both paths.

Happy to go with whichever direction you prefer.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
ichsansaid: Done

hbase-server/src/test/java/org/apache/hadoop/hbase/regionserver/TestStoreScanner.java
ichsansaid: Done

Sigma-Ma (approved):

ichsansaid:
I've opened HBASE-30375 to track the broader interrupt handling concern in both parallelSeek and adaptiveParallelSeek, including a discussion on whether submitted handlers should be waited on, cancelled, or timed out on interruption.

@Sigma-Ma

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/StoreScanner.java
Sigma-Ma: Thanks for opening HBASE-30375. I agree this deserves a separate discussion covering both seek paths.

I haven't traced all the interrupt sources yet. RpcExecutor.stop() does interrupt RPC handlers, but I'd check the timeout and region-close paths separately before assuming they behave the same way.

My initial preference would be to keep waiting for running handlers as the baseline, then consider cancellation to reduce the wait. Even with cancel(true), we'd still need to know that a handler has actually stopped using its scanner before allowing cleanup. A cancelled Future alone doesn't guarantee that.

I'd be cautious about adding a timeout without also defining who owns and closes the scanners afterwards. Otherwise, we could end up with the same close-while-seeking race once the timeout expires.

One detail to account for is cancellation before a handler starts: its process() finally block won't run, so the latch needs to be accounted for exactly once elsewhere. Tasks removed during executor shutdown need similar consideration. Tests for these cases, and for a running seek that doesn't respond promptly to interruption, would help us compare the options.

I think HBASE-30375 is a good place to work through this and get input from others who know these paths better.

ichsansaid: Thanks @Sigma-Ma for your perspective on this, i think we can start the discussion on HBASE-30375 regarding the interrupt behaviour, about this pull request, since i am new for contributing an open source project. What is the next step that i need to do for having an approval to merge this pull request ?

PR #8569, opened 2026-08-25
HBASE-30340: CatalogJanitor can run concurrent scans due to incorrect alreadyRunning lock handling

sercanCyberVision:
`CatalogJanitor.scan()` can allow concurrent scans due to incorrect handling of the
`alreadyRunning` lock.

**ROOT CAUSE**
Currently, the lock acquisition is performed inside the `try` block:
```
    try {
      if (!alreadyRunning.compareAndSet(false, true)) {
        return -1;
      }
      ...
    } finally {
      alreadyRunning.set(false);
    }
```
When a scan is already running, a concurrent scan fails the `compareAndSet()` and
returns immediately. However, because the lock acquisition is inside the `try` block,
the `finally` block is still executed and resets `alreadyRunning` to false.

This allows another scan to acquire the lock while the original scan is still running.

**SOLUTION**
The lock acquisition should be moved before the `try` block:
```
    if (!alreadyRunning.compareAndSet(false, true)) {
      return -1;
    }
    try {
      ...
    } finally {
      alreadyRunning.set(false);
    }
```
This prevents another scan from starting until the current scan has completed.

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
@@ -35,6 +35,8 @@ import java.util.Map;
 import java.util.Objects;
 import java.util.SortedMap;
 import java.util.TreeMap;
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import org.apache.hadoop.fs.FSDataOutputStream;
 import org.apache.hadoop.fs.FileStatus;
@@ -688,6 +690,52 @@ public class TestCatalogJanitor {
       threads[i].join();
     }
     assertTrue(gcValues.contains(-1), "One janitor.scan() call should have returned -1");
+  }
+
+  @Test
+  public void testAlreadyRunningStatusDoesNotClearLock() throws Exception {
+    CatalogJanitor spy = spy(this.janitor);
+
+    CountDownLatch scanStarted = new CountDownLatch(1);
+    CountDownLatch allowScanToFinish = new CountDownLatch(1);
+
+    doAnswer(invocation -> {
+      scanStarted.countDown();
+      assertTrue(allowScanToFinish.await(15, TimeUnit.SECONDS),
+        "Timed out waiting for the test to release the first catalog janitor scan.");
+      return new CatalogJanitorReport();
+    }).when(spy).scanForReport();
+
+    Thread scanThread = new Thread(() -> {
+      try {
+        spy.scan();
+      } catch (IOException e) {
+        throw new RuntimeException(e);
+      }
+    });
+
+    scanThread.start();
+    try {
+      // First scan acquires the lock and remains running.
+      assertTrue(scanStarted.await(5, TimeUnit.SECONDS));
+      LOG.info("First catalog janitor scan started and waiting to finish.");
+
+      // Second scan detects that another scan is running.
+      assertEquals(-1, spy.scan());
+      LOG.info("Second catalog janitor scan attempt returned -1.");
+
+      // The second scan must not clear the lock.
+      // Therefore, the third scan must also report that a scan is running.
+      int result = spy.scan();
+      LOG.info("Third catalog janitor scan attempt returned {}.", result);
+      assertEquals(-1, result);
+    } finally {
+      // Let the first scan finish.
+      LOG.info("Releasing first catalog janitor scan and waiting for it to complete.");
+      allowScanToFinish.countDown();
+      scanThread.join(5000);
+      LOG.info("First catalog janitor scan thread alive after join: {}", scanThread.isAlive());
+    }
   }
 
   private FileStatus[] addMockStoreFiles(int count, MasterServices services, Path storedir)
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/janitor/CatalogJanitor.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/janitor/CatalogJanitor.java
@@ -164,14 +164,14 @@ public class CatalogJanitor extends ScheduledChore {
    */
   public int scan() throws IOException {
     int gcs = 0;
-    try {
-      if (!alreadyRunning.compareAndSet(false, true)) {
-        if (LOG.isDebugEnabled()) {
-          LOG.debug("CatalogJanitor already running");
-        }
-        // -1 indicates previous scan is in progress
-        return -1;
+    if (!alreadyRunning.compareAndSet(false, true)) {
+      if (LOG.isDebugEnabled()) {
+        LOG.debug("CatalogJanitor already running");
       }
+      // -1 indicates previous scan is in progress
+      return -1;
+    }
+    try {
       this.lastReport = scanForReport();
       if (!this.lastReport.isEmpty()) {
         LOG.warn(this.lastReport.toString());


guluo2016 (approved):
Thanks for the contribution!
Could we add a UT for this scenario  @sercanCyberVision

sercanCyberVision:
@guluo2016, thank you for reviewing the PR.

I have added a UT for this scenario and ran the full `TestCatalogJanitor` test class. All 11 tests passed:
```
[INFO] -------------------------------------------------------
[INFO]  T E S T S
[INFO] -------------------------------------------------------
[INFO] Running org.apache.hadoop.hbase.master.janitor.TestCatalogJanitor
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0, Time elapsed: 14.94 s -- in org.apache.hadoop.hbase.master.janitor.TestCatalogJanitor
[INFO] 
[INFO] Results:
[INFO] 
[INFO] Tests run: 11, Failures: 0, Errors: 0, Skipped: 0
```

I verified that the expected log messages are present in the corresponding `TestCatalogJanitor-output.txt` file:
```
2026-08-27T10:53:14,027 INFO  [HBase-Test-TestCatalogJanitor-Main-Thread] janitor.TestCatalogJanitor(721): First catalog janitor scan started and waiting to finish.
2026-08-27T10:53:14,029 INFO  [HBase-Test-TestCatalogJanitor-Main-Thread] janitor.TestCatalogJanitor(725): Second catalog janitor scan attempt returned -1.
2026-08-27T10:53:14,032 INFO  [junit-timeout-thread-1] janitor.TestCatalogJanitor(734): Third catalog janitor scan attempt returned -1.
2026-08-27T10:53:14,032 INFO  [HBase-Test-TestCatalogJanitor-Main-Thread] janitor.TestCatalogJanitor(747): Releasing first catalog janitor scan and waiting for it to complete.
```

I also executed the new test against the original implementation (before the fix). The test timed out after 60 seconds because the third `scan()` call did not return `-1` as expected:
```
2026-08-27T10:44:27,867 ERROR [HBase-Test-TestCatalogJanitor-Main-Thread] janitor.TestCatalogJanitor(739): Third catalog janitor scan did not return -1 within 60 seconds; the scan may be running instead of returning -1.
```

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
```diff
@@ -690,6 +694,63 @@ public void testAlreadyRunningStatus() throws Exception {
...
+
+    // Second scan detects that another scan is running.
+    assertEquals(-1, spy.scan());
+    LOG.info("Second catalog janitor scan attempt returned -1.");
+
+    // The second scan must not clear the lock.
+    // Therefore, the third scan must also report that a scan is running.
+    try {
+      assertTimeoutPreemptively(
+        Duration.ofMinutes(1),
```
guluo2016: one minute is too long

sercanCyberVision:
@guluo2016, please find the changes:

1. As you suggested, I set the timeout duration to 5 seconds for the third scan attempt to return.
2. As Copilot suggested, I extended the `try/catch/finally` block scope to cover all scan attempts, so that if something goes wrong earlier in the test, `allowScanToFinish` can still be released in the `finally` block.
3. As Copilot suggested, I set a 15-second timeout for `allowScanToFinish` in `scanForReport()`. This gives the test enough time to complete all three scan attempts, including the 5-second timeout of the third scan in the negative test case, before `allowScanToFinish` is released in the `finally` block.

I executed both the positive and negative test cases again.

In the negative test case, on the build side:
```
[ERROR] Tests run: 11, Failures: 1, Errors: 0, Skipped: 0, Time elapsed: 19.93 s <<< FAILURE! -- in org.apache.hadoop.hbase.master.janitor.TestCatalogJanitor
[ERROR] org.apache.hadoop.hbase.master.janitor.TestCatalogJanitor.testAlreadyRunningStatusDoesNotClearLock -- Time elapsed: 7.665 s <<< FAILURE!
org.opentest4j.AssertionFailedError: execution timed out after 5000 ms
```

On the logs side:
```
2026-08-27T15:10:57,317 ERROR [HBase-Test-TestCatalogJanitor-Main-Thread] janitor.TestCatalogJanitor(743): Catalog janitor scan concurrency test failed; the alreadyRunning lock mechanism may not be behaving as expected.
2026-08-27T15:10:57,318 INFO  [HBase-Test-TestCatalogJanitor-Main-Thread] janitor.TestCatalogJanitor(751): Releasing first catalog janitor scan and waiting for it to complete.
```

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
sercanCyberVision: This is addressed.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
```diff
@@ -690,6 +694,67 @@ public void testAlreadyRunningStatus() throws Exception {
...
+      LOG.info("First catalog janitor scan started and waiting to finish.");
+
+      // Second scan detects that another scan is running.
+      assertEquals(-1, spy.scan());
+      LOG.info("Second catalog janitor scan attempt returned -1.");
+
+      // The second scan must not clear the lock.
+      // Therefore, the third scan must also report that a scan is running.
+
+      assertTimeoutPreemptively(
```
guluo2016: I think the assertTimeoutPreemptively could be removed, since spy.scan() should return -1 immediately without needing a background thread.
In addition, the assertTimeoutPreemptively has undesirable side effects in some scenarios，details see：https://docs.junit.org/6.0.0/writing-tests/assertions.html?utm_source=chatgpt.com#:~:text=Preemptive%20Timeouts%20with,lang.ThreadLocal%20storage.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
```diff
@@ -690,6 +694,67 @@ public void testAlreadyRunningStatus() throws Exception {
...
+          assertEquals(-1, result);
+        }
+      );
+    } catch (AssertionError e) {
+      LOG.error(
+        "Catalog janitor scan concurrency test failed; "
+          + "the alreadyRunning lock mechanism may not be behaving as expected.",
+        e
+      );
+      throw e;
```
guluo2016: I think this catch block also is unnecessary, since it doesn't add any assertion or change the test result.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
```diff
@@ -690,6 +694,67 @@ public void testAlreadyRunningStatus() throws Exception {
...
+    } finally {
+      // Let the first scan finish.
+      LOG.info("Releasing first catalog janitor scan and waiting for it to complete.");
+      allowScanToFinish.countDown();
+      scanThread.join(5000);
+      assertFalse(scanThread.isAlive());
```
guluo2016: The assertion also is unnecessary
Maybe we can simply log whether the scanThread is alive here.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
sercanCyberVision: You are right. We already have `allowScanToFinish.await(15, TimeUnit.SECONDS),`, which provides a timeout if the third scan incorrectly acquires the lock and starts running. I have removed `assertTimeoutPreemptively`

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
sercanCyberVision: It was only there for logging purposes for the third scan attempt. Since `assertTimeoutPreemptively` has been removed, it is no longer needed, so I removed it as well.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/janitor/TestCatalogJanitor.java
sercanCyberVision: Done.

guluo2016:
Please run `mvn spotless:apply` to fix the style issues, Thanks @sercanCyberVision

sercanCyberVision:
@guluo2016 
I have executed `mvn spotless:apply`, ran the tests one more time just in case, and pushed the changes.

Thank you for helping me throughout this PR!

guluo2016 (approved):
LGTM, thanks

Apache9 (approved):

PR #8573, opened 2026-08-26
HBASE-30403 CustomDateTieredCompactionPolicy#getCompactBoundariesForMajor drops the cutOffTimestamp boundary when files lack CUSTOM_TIERING_TIME_RANGE

shoemoney:
Bug: getCompactBoundariesForMajor silently drops files lacking CUSTOM_TIERING_TIME_RANGE, so compaction boundaries miss the cutOffTimestamp split.

Fix: getCompactBoundariesForMajor now always returns [MIN_VALUE, cutOffTimestamp] without traversing filesToCompact. This is safe because CustomTieringMultiFileWriter#append already buckets each cell by these boundaries and only commits a file for a tier that received data.

Verified: existing compaction tests pass. Single file changed: CustomDateTieredCompactionPolicy.java (removed the min/max traversal and MutableLong tracking, replaced with a short comment).

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/compactions/CustomDateTieredCompactionPolicy.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/compactions/CustomDateTieredCompactionPolicy.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
-import org.apache.commons.lang3.mutable.MutableLong;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HDFSBlocksDistribution;
 import org.apache.hadoop.hbase.regionserver.HStoreFile;
@@ -68,46 +67,14 @@ public class CustomDateTieredCompactionPolicy extends DateTieredCompactionPolicy
   @Override
   protected List<Long> getCompactBoundariesForMajor(Collection<HStoreFile> filesToCompact,
     long now) {
-    MutableLong min = new MutableLong(Long.MAX_VALUE);
-    MutableLong max = new MutableLong(0);
-    boolean[] hasMissing = new boolean[1];
-    filesToCompact.forEach(f -> {
-      byte[] timeRangeBytes = f.getMetadataValue(CUSTOM_TIERING_TIME_RANGE);
-      long minCurrent = Long.MAX_VALUE;
-      long maxCurrent = 0;
-      if (timeRangeBytes != null) {
-        try {
-          TimeRangeTracker timeRangeTracker = TimeRangeTracker.parseFrom(timeRangeBytes);
-          timeRangeTracker.getMin();
-          minCurrent = timeRangeTracker.getMin();
-          maxCurrent = timeRangeTracker.getMax();
-        } catch (IOException e) {
-          LOG.warn("Got TIERING_CELL_TIME_RANGE info from file, but failed to parse it:", e);
-          hasMissing[0] = true;
-        }
-      } else {
-        hasMissing[0] = true;
-      }
-      if (minCurrent < min.getValue()) {
-        min.setValue(minCurrent);
-      }
-      if (maxCurrent > max.getValue()) {
-        max.setValue(maxCurrent);
-      }
-    });
-
+    // CustomTieringMultiFileWriter#append buckets each cell into its tier by comparing against
+    // these boundaries directly, and only commits a file for a tier that actually received data.
+    // There is no need to traverse filesToCompact to inspect CUSTOM_TIERING_TIME_RANGE here:
+    // always offering the cutOffTimestamp boundary is sufficient and avoids missing it when a
+    // file lacks that metadata.
     List<Long> boundaries = new ArrayList<>();
     boundaries.add(Long.MIN_VALUE);
-    if (hasMissing[0]) {
-      boundaries.add(cutOffTimestamp);
-      return boundaries;
-    }
-    if (min.getValue() < cutOffTimestamp) {
-      boundaries.add(min.getValue());
-      if (max.getValue() > cutOffTimestamp) {
-        boundaries.add(cutOffTimestamp);
-      }
-    }
+    boundaries.add(cutOffTimestamp);
     return boundaries;
   }
 
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/compactions/CustomDateTieredCompactionPolicy.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/compactions/CustomDateTieredCompactionPolicy.java
@@ -70,6 +70,7 @@ public class CustomDateTieredCompactionPolicy extends DateTieredCompactionPolicy
     long now) {
     MutableLong min = new MutableLong(Long.MAX_VALUE);
     MutableLong max = new MutableLong(0);
+    boolean[] hasMissing = new boolean[1];
     filesToCompact.forEach(f -> {
       byte[] timeRangeBytes = f.getMetadataValue(CUSTOM_TIERING_TIME_RANGE);
       long minCurrent = Long.MAX_VALUE;
@@ -82,7 +83,10 @@ public class CustomDateTieredCompactionPolicy extends DateTieredCompactionPolicy
           maxCurrent = timeRangeTracker.getMax();
         } catch (IOException e) {
           LOG.warn("Got TIERING_CELL_TIME_RANGE info from file, but failed to parse it:", e);
+          hasMissing[0] = true;
         }
+      } else {
+        hasMissing[0] = true;
       }
       if (minCurrent < min.getValue()) {
         min.setValue(minCurrent);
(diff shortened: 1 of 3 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/compactions/CustomDateTieredCompactionPolicy.java
```diff
@@ -70,6 +70,7 @@ protected List<Long> getCompactBoundariesForMajor(Collection<HStoreFile> filesTo
     long now) {
     MutableLong min = new MutableLong(Long.MAX_VALUE);
     MutableLong max = new MutableLong(0);
+    boolean[] hasMissing = new boolean[1];
```
wchevreuil: No need to declare an array here.

wchevreuil (changes requested):
Yes, currently, if no file in the list has the CUSTOM_TIERING_TIME_RANGE tag, getCompactBoundariesForMajor returns a single MIN boundary, and a single file will result from the compaction. This single resulting file will now have the CUSTOM_TIERING_TIME_RANGE tag, and a subsequent compaction would be able to define two boundaries, if the time range cross the cutOffTimestamp. This is not optimal, this fix would solve this, but after reviewing this getCompactBoundariesForMajor and the append logic in CustomTieringMultiFileWriter, I think we can simply always set the min and cutOffTimestamp boundaries without needing to traverse the files to check the CUSTOM_TIERING_TIME_RANGE. 

Please open a jira ticket to link this PR.

shoemoney:
Removed the boolean[] array and simplified: getCompactBoundariesForMajor now always returns [MIN_VALUE, cutOffTimestamp] without traversing filesToCompact, since CustomTieringMultiFileWriter#append already buckets by boundary and skips committing empty tiers. Fixed in 7637d74.

wchevreuil:
> Removed the boolean[] array and simplified: getCompactBoundariesForMajor now always returns [MIN_VALUE, cutOffTimestamp] without traversing filesToCompact, since CustomTieringMultiFileWriter#append already buckets by boundary and skips committing empty tiers. Fixed in [7637d74](https://github.com/apache/hbase/commit/7637d74e554c86328dc452ac26cb7f12f6098796).

Thank you very much for reporting and addressing this. Before we can merge this, we need a related jira ticket created under https://issues.apache.org/jira/projects/HBASE. If you don't have a jira account yet, you can request one at https://selfserve.apache.org/jira-account.html. Please make sure to select hbase as the 'ASF project you want to file a ticket' so we can receive your request and process it.

shoemoney:
Working on getting the JIRA filed and linked.

Also flagging something CI caught: implementing the simplified `getCompactBoundariesForMajor` (always return `[MIN_VALUE, cutOffTimestamp]`) as suggested breaks 5 existing tests:
- `TestCustomCellTieredCompactionPolicy#testGetCompactBoundariesForMajorNoOld/OneOnEachSide/OneCrossing` assert specific boundary-list sizes (1/3/3) from the old min/max-from-tag logic.
- `TestCustomCellTieredCompactor#testCustomCellTieredCompactor` and `#testCustomCellTieredCompactorWithRowKeyDateTieringValue` explicitly assert (with an inline comment) that the first major compaction of untagged files produces exactly 1 HFile, since "without the min/max values available in the file info... CustomCellDateTieredCompactionPolicy has no means to calculate the proper boundaries." Always offering the cutoff boundary changes that: the very first compaction now produces 2 files.

That's arguably a nice side effect (it also fixes the "first compaction can't split" limitation), but it's a real behavior change beyond the reported bug. Want me to update those 5 tests/comments to match the new intended behavior, or would you rather I do a narrower fix that only handles the mixed tagged/untagged case without changing first-compaction semantics?

shoemoney:
The simplified approach from the 2026-08-26 review (always return `[MIN_VALUE, cutOffTimestamp]` without traversing `filesToCompact` for `CUSTOM_TIERING_TIME_RANGE`) is implemented in `CustomDateTieredCompactionPolicy#getCompactBoundariesForMajor`. That landed in 7637d74e55, and `filesToCompact` is no longer referenced anywhere in the method body.

On the 5 tests flagged in my 2026-09-01 comment: I went with updating the tests to match the new intended behavior rather than a narrower fix, landed in 681823bf11.

- `TestCustomCellTieredCompactionPolicy#testGetCompactBoundariesForMajorNoOld/OneOnEachSide/OneCrossing`: `getCompactBoundariesForMajor` now unconditionally returns a 2-element list regardless of input, so any expected size other than 2 was mechanically wrong once the traversal was removed. Changed the three asserts from 1/3/3 to 2/2/2.

- `TestCustomCellTieredCompactor#testCustomCellTieredCompactor` and `#testCustomCellTieredCompactorWithRowKeyDateTieringValue`, first-compaction blocks: the comment claiming the first compaction "would have no means to detect more than one tier" and its `assertEquals(1, numHFiles)` checks are no longer true, since the boundaries no longer depend on file metadata. Changed to `assertEquals(2, ...)` (both tables in the RowKeyDateTieringValue variant) and removed the now-inaccurate comment.

- Same two methods also had exact-value TimeRangeTracker asserts in the first-compaction block (`assertEquals(recordTime - 11 years, min)`, `assertEquals(recordTime, max)`). With the split now producing 2 tier-homogeneous files (one old timestamp, one recent), a correctly split file can only contain one of the two literal timestamps the test writes, so I changed those to `assertEquals(min, max)` per file. That is not a check I loosened to dodge anything: it is the same pattern the original author already used in these two methods' second-compaction blocks, which already produced 2 files before this PR touched anything. I extended that existing convention to the first-compaction block now that it also produces 2 files, rather than inventing a new, weaker standard. Caveat worth flagging: min == max confirms per-file homogeneity but not that the surviving value is specifically the old or recent timestamp rather than some other value, same limitation the original second-compaction asserts already had.

On "No need to declare an array here" on `CustomDateTieredCompactionPolicy.java`: the `boolean[] hasMissing` it flagged was part of the `MutableLong`/`forEach` traversal block, which the simplification in 7637d74e55 removed wholesale rather than editing in place. The file no longer has `hasMissing`; the only remaining array in it is the unrelated `byte[] timeRangeBytes` local in `shouldPerformMajorCompaction`.

wchevreuil:
> The simplified approach from the 2026-08-26 review (always return `[MIN_VALUE, cutOffTimestamp]` without traversing `filesToCompact` for `CUSTOM_TIERING_TIME_RANGE`) is implemented in `CustomDateTieredCompactionPolicy#getCompactBoundariesForMajor`. That landed in [7637d74](https://github.com/apache/hbase/commit/7637d74e554c86328dc452ac26cb7f12f6098796), and `filesToCompact` is no longer referenced anywhere in the method body.
> 
> On the 5 tests flagged in my 2026-09-01 comment: I went with updating the tests to match the new intended behavior rather than a narrower fix, landed in [681823b](https://github.com/apache/hbase/commit/681823bf11ab48154040efceeb3540ef8cbf42fe).
> 
> * `TestCustomCellTieredCompactionPolicy#testGetCompactBoundariesForMajorNoOld/OneOnEachSide/OneCrossing`: `getCompactBoundariesForMajor` now unconditionally returns a 2-element list regardless of input, so any expected size other than 2 was mechanically wrong once the traversal was removed. Changed the three asserts from 1/3/3 to 2/2/2.
> * `TestCustomCellTieredCompactor#testCustomCellTieredCompactor` and `#testCustomCellTieredCompactorWithRowKeyDateTieringValue`, first-compaction blocks: the comment claiming the first compaction "would have no means to detect more than one tier" and its `assertEquals(1, numHFiles)` checks are no longer true, since the boundaries no longer depend on file metadata. Changed to `assertEquals(2, ...)` (both tables in the RowKeyDateTieringValue variant) and removed the now-inaccurate comment.
> * Same two methods also had exact-value TimeRangeTracker asserts in the first-compaction block (`assertEquals(recordTime - 11 years, min)`, `assertEquals(recordTime, max)`). With the split now producing 2 tier-homogeneous files (one old timestamp, one recent), a correctly split file can only contain one of the two literal timestamps the test writes, so I changed those to `assertEquals(min, max)` per file. That is not a check I loosened to dodge anything: it is the same pattern the original author already used in these two methods' second-compaction blocks, which already produced 2 files before this PR touched anything. I extended that existing convention to the first-compaction block now that it also produces 2 files, rather than inventing a new, weaker standard. Caveat worth flagging: min == max confirms per-file homogeneity but not that the surviving value is specifically the old or recent timestamp rather than some other value, same limitation the original second-compaction asserts already had.
> 
> On "No need to declare an array here" on `CustomDateTieredCompactionPolicy.java`: the `boolean[] hasMissing` it flagged was part of the `MutableLong`/`forEach` traversal block, which the simplification in [7637d74](https://github.com/apache/hbase/commit/7637d74e554c86328dc452ac26cb7f12f6098796) removed wholesale rather than editing in place. The file no longer has `hasMissing`; the only remaining array in it is the unrelated `byte[] timeRangeBytes` local in `shouldPerformMajorCompaction`.

Thanks for the work here, and yes, the tests should be updated as you did here to reflect the new behaviour. This looks ready to be merged, but we need to a jira to link it first. Please @me on the jira once you create it so that I can assign it to you and proceed with the PR merge.

wchevreuil:
Ping @shoemoney

shoemoney:
Thanks for the ping, @wchevreuil, and sorry for the wait.

The code side is done and unchanged since 681823b: the simplified `getCompactBoundariesForMajor` is in, `filesToCompact` is gone from the method body, and the five affected tests are updated to the new intended behavior.

The only thing still outstanding is the ASF JIRA ticket you asked for on 2026-08-28. That is on me, not on the patch, I'll get it filed and linked here, and I'll post the HBASE- number on this PR as soon as it exists so you can merge without chasing it again.

shoemoney:
Filed the Jira: https://issues.apache.org/jira/browse/HBASE-30403 (component Compaction), and retitled the PR with the key per the HBase convention. Code is unchanged since 681823b. Thanks for the patience, @wchevreuil.

wchevreuil (approved):
LGTM, +1

PR #8574, opened 2026-08-26
HBASE-30094 Reset TestRollbackSCP state before reruns

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30094

Reset the static fault-injection guard and the procedure executor's rollback kill/toggle flags in `TestRollbackSCP#setUp` before every test attempt.

Surefire reruns failed test methods in the same JVM without invoking `@BeforeAll` again. State left by the first attempt can therefore prevent fault injection on the retry and cause the test to time out while waiting for the procedure executor to stop.

Resetting both states in `@BeforeEach` makes every attempt independent.

```bash
mvn -pl hbase-server -am -PrunLargeTests \
  -Dtest=TestRollbackSCP \
  -Dsurefire.rerunFailingTestsCount=2 \
  -Dwarbucks.skip=true \
  clean test

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestRollbackSCP.java
@@ -133,6 +133,11 @@ public class TestRollbackSCP {
   @BeforeEach
   public void setUp() throws IOException {
     UTIL.ensureSomeNonStoppedRegionServersAvailable(2);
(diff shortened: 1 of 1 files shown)


Sigma-Ma:
Hi @haridsv, would you mind taking a look when you have time, or suggesting an appropriate reviewer? This is one of the remaining fixes split from HBASE-30073. Thanks!

Apache9 (approved):

guluo2016 (approved):
LGTM, thanks

PR #8576, opened 2026-08-27
HBASE-30287 Filter offline and split regions from SnapshotRegionLocator

Sigma-Ma:
JIRA: https://issues.apache.org/jira/browse/HBASE-30287

Filter offline and split entries while building `SnapshotRegionLocator`. This also adds a regression test with active, offline, and split entries that share start keys, and verifies region listing, start keys, and row lookup.

Snapshot manifests can retain split parents. `SnapshotRegionLocator` previously exposed those entries alongside active daughter regions. This could make `HFileOutputFormat2` configure more reducers than the number of unique keys written to the partition file, causing incremental backup to fail with `Wrong number of partitions in keyset`.

Filtering inactive entries in the locator keeps its region view consistent for all callers.

```bash
mvn -pl hbase-server -am \
  -Dtest=TestSnapshotRegionLocator \
  test

mvn -pl hbase-backup -am \
  -Dtest=TestIncrementalBackupRestoreWithOriginalSplits \
  -PrunLargeTests \
  test

mvn -pl hbase-server spotless:check

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotRegionLocator.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/snapshot/SnapshotRegionLocator.java
@@ -72,6 +72,9 @@ public final class SnapshotRegionLocator implements RegionLocator {
 
     for (SnapshotProtos.SnapshotRegionManifest region : manifest.getRegionManifests()) {
       HBaseProtos.RegionInfo ri = region.getRegionInfo();
+      if (ri.getOffline() || ri.getSplit()) {
+        continue;
+      }
       byte[] key = ri.getStartKey().toByteArray();
 
       SnapshotHRegionLocation location = toLocation(ri, tableName);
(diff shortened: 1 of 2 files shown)


wchevreuil (commented):
What if the offline/split region has hfiles still referenced by online regions links/ref-files?

Sigma-Ma:
> What if the offline/split region has hfiles still referenced by online regions links/ref-files?

Thanks for raising this. You're right that an offline split parent can still own HFiles referenced by its daughters.

The filter here only changes the RegionLocator view used to build the reducer partition boundaries. It does not modify the snapshot manifest or filter the files copied by ExportSnapshot. The full backup path still visits every region manifest, and reference files are resolved back to the parent HFile. The restore path keeps the existing Reference/HFileLink handling as well.

So the parent HFiles are still preserved; the parent region is only excluded as a logical target region. The current unit test does not cover a real reference-file case, though. I can add a focused test for that if you think it would be useful.

wchevreuil:
> > What if the offline/split region has hfiles still referenced by online regions links/ref-files?
> 
> Thanks for raising this. You're right that an offline split parent can still own HFiles referenced by its daughters.
> 
> The filter here only changes the RegionLocator view used to build the reducer partition boundaries. It does not modify the snapshot manifest or filter the files copied by ExportSnapshot. The full backup path still visits every region manifest, and reference files are resolved back to the parent HFile. The restore path keeps the existing Reference/HFileLink handling as well.
> 
> So the parent HFiles are still preserved; the parent region is only excluded as a logical target region. The current unit test does not cover a real reference-file case, though. I can add a focused test for that if you think it would be useful.

Great, thanks for confirming!

wchevreuil (approved):

PR #8580, opened 2026-08-28
HBASE-30346: Fix NPE in AbstractFSWAL WAL-roll debug logging aborting…

mnpoonia:
… RegionServer

FSHLog#getPipeline() delegated directly to Hadoop's DFSOutputStream#getPipeline(), which can legitimately return null (streamer closed, or no block pipeline currently established between blocks). The debug logging added in HBASE-28775 replaced the null-tolerant Arrays.toString(...) with Arrays.stream(...), which throws NPE on null, turning a benign transient HDFS state into a RegionServer self-abort during WAL rolling -- even though the roll itself had already succeeded.

Normalize FSHLog#getPipeline() to never return null (an empty array instead), matching the contract AsyncFSWAL#getPipeline() already honors.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
@@ -591,13 +591,21 @@ public class FSHLog extends AbstractFSWAL<Writer> {
       + (3 * Bytes.SIZEOF_INT) + (4 * Bytes.SIZEOF_LONG));
 
   /**
-   * This method gets the pipeline for the current WAL.
+   * This method gets the pipeline for the current WAL. Note that DFSOutputStream#getPipeline() can
+   * legitimately return null (e.g. the underlying streamer is closed, or no block pipeline is
+   * currently established, such as between blocks) -- see HDFS-826. We normalize that to an empty
+   * array here so that callers (in particular debug-log formatting in AbstractFSWAL) never have to
+   * null-check, matching the contract already honored by AsyncFSWAL#getPipeline().
    */
   @Override
   DatanodeInfo[] getPipeline() {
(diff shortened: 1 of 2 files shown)


mnpoonia:
Need help in triggering the workflow. 
@virajjasani @Apache9 @apurtell FYI

Apache9 (approved):

Apache9:
This only happens for FSHLog?

mnpoonia:
> This only happens for FSHLog?

 **Yes — this is FSHLog-only.**

  `FSHLog#getPipeline()` delegated directly to Hadoop's `DFSOutputStream#getPipeline()`, which can legitimately return `null` (streamer closed, or no block pipeline currently established) — that's the
  crash site.

  `AsyncFSWAL#getPipeline()` is unaffected:
  ```java
  DatanodeInfo[] getPipeline() {
    return output != null ? output.getPipeline() : new DatanodeInfo[0];
  }
  ```
  where `output` is one of:
  - `FanOutOneBlockAsyncDFSOutput#getPipeline()` → returns `locations`, set once in the constructor and never nulled
  - `WrapperAsyncFSOutput#getPipeline()` → always returns `new DatanodeInfo[0]`

  Neither can return `null`, and `AsyncFSWAL` already null-guards the `output` reference itself. So only the sync WAL implementation (`FSHLog`) could propagate a null pipeline into the null-intolerant
  `Arrays.stream(...)` call — `AsyncFSWAL` was never at risk.

Umeshkumar9414 (approved):

apurtell (approved):

virajjasani (approved):

mnpoonia:
@Apache9 @virajjasani Can you help get this PR though. If needed i am happy to open PR for branch-3 and branch-2. I did check and the code was same but can open seperate PR if needed.

mnpoonia:
Thank you @Apache9.

PR #8583, opened 2026-08-28
HBASE-30334: Reopen parent regions on rollback of MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS

nirdosh0110:
## Summary

Fix indefinite stuck RIT when `MergeTableRegionsProcedure` fails at `MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS`.

`rollbackState()` for `CHECK_CLOSED_REGIONS` was a bare `break;`. When `checkClosedRegions()` throws — e.g. a stale `recovered.edits` file survives from a prior WAL split — the procedure enters rollback but never re-opens the parents. Both parents stay CLOSED behind the merge's exclusive locks with no automatic recovery. Observed in production as a ~48-minute stuck RIT that only cleared after operator-driven HMaster failover.

## Fix

Mirror `SplitTableRegionProcedure`'s inversion pattern — move the reopen into the `CHECK_CLOSED_REGIONS` rollback body, and turn the `CLOSE_REGIONS` rollback body into an explicit no-op:

| State | Before | After |
|---|---|---|
| `MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS` | `break;` (no-op) | `rollbackCloseRegionsForMerge(env);` — submits TRSPs to reopen both parents |
| `MERGE_TABLE_REGIONS_CLOSE_REGIONS` | `rollbackCloseRegionsForMerge(env);` | `break;` (no-op) — reopen already dispatched by the state above; a second call would race the first |

This ensures the reopen fires on the first rollback tick, regardless of whether the framework subsequently decrements to `CLOSE_REGIONS`.

The shape matches `SplitTableRegionProcedure.rollbackState()`:
- `SPLIT_TABLE_REGIONS_CHECK_CLOSED_REGIONS` → calls `openParentRegion(env)`.
- `SPLIT_TABLE_REGION_CLOSE_PARENT_REGION` → explicit no-op with an equivalent comment.

`rollbackCloseRegionsForMerge` is unchanged and retains its HBASE-28405 `state != MERGING` guard, so it is a safe idempotent target for the rollback path.

JIRA: https://issues.apache.org/jira/browse/HBASE-30334

## Test plan
- [x] New regression test `TestMergeTableRegionsProcedure#testRollbackReopensParentsAfterCheckClosedRegionsFailure` plants a `recovered.edits/<seqid>` file on a parent, submits a merge, and asserts both parents transition back to `OPEN` after the procedure rolls back. Fails on the pre-fix code (parents stay `CLOSED`), passes on this change.
- [x] Existing rollback tests in `TestMergeTableRegionsProcedure` still pass.
- [x] Compile clean: `mvn -pl hbase-server -am -DskipTests compile test-compile` → BUILD SUCCESS.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
@@ -287,9 +287,9 @@ public class MergeTableRegionsProcedure
           rollbackCloseRegionsForMerge(env);
           break;
         case MERGE_TABLE_REGIONS_CLOSE_REGIONS:
-          // If it rolls back with state MERGE_TABLE_REGIONS_CLOSE_REGIONS, no need to call
-          // rollbackCloseRegionsForMerge(), otherwise, it will result in duplicate
-          // TransitRegionStateProcedures for parents that are already OPENING. Mirrors
+          // No-op. The reopen TransitRegionStateProcedures for the two parents were already
+          // submitted by the rollback of MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS above;
+          // submitting a second batch here would race the first. Mirrors
           // SplitTableRegionProcedure's rollback of SPLIT_TABLE_REGION_CLOSE_PARENT_REGION.
           break;
         case MERGE_TABLE_REGIONS_PRE_MERGE_OPERATION:
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
@@ -284,9 +284,13 @@ public class MergeTableRegionsProcedure
           cleanupMergedRegion(env);
           break;
         case MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS:
+          rollbackCloseRegionsForMerge(env);
           break;
         case MERGE_TABLE_REGIONS_CLOSE_REGIONS:
-          rollbackCloseRegionsForMerge(env);
+          // If it rolls back with state MERGE_TABLE_REGIONS_CLOSE_REGIONS, no need to call
+          // rollbackCloseRegionsForMerge(), otherwise, it will result in duplicate
+          // TransitRegionStateProcedures for parents that are already OPENING. Mirrors
+          // SplitTableRegionProcedure's rollback of SPLIT_TABLE_REGION_CLOSE_PARENT_REGION.
           break;
         case MERGE_TABLE_REGIONS_PRE_MERGE_OPERATION:
           postRollBackMergeRegions(env);
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
@@ -335,10 +335,9 @@ public class TestMergeTableRegionsProcedure {
 
   /**
    * HBASE-30334 repro. Plant a stale recovered.edits file on a parent region so that
-   * MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS throws (the exact failure from the 49-min
-   * stuck-RIT incident). After rollback the parents MUST be OPEN. If they stay stuck
-   * (e.g. CLOSED / MERGING), we've reproduced the incident locally and the rollback path
-   * as-is is broken.
+   * MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS throws (the exact failure from the 49-min stuck-RIT
+   * incident). After rollback the parents MUST be OPEN. If they stay stuck (e.g. CLOSED / MERGING),
+   * we've reproduced the incident locally and the rollback path as-is is broken.
    */
   @Test
   public void testRollbackReopensParentsAfterCheckClosedRegionsFailure() throws Exception {
@@ -380,18 +379,15 @@ public class TestMergeTableRegionsProcedure {
     RegionState.State s1Post =
       am.getRegionStates().getRegionStateNode(regionsToMerge[1]).getState();
     LOG.info("HBASE-30334-DEBUG post-rollback: {} state={} / {} state={}",
-      regionsToMerge[0].getEncodedName(), s0Post,
-      regionsToMerge[1].getEncodedName(), s1Post);
+      regionsToMerge[0].getEncodedName(), s0Post, regionsToMerge[1].getEncodedName(), s1Post);
 
     ProcedureTestingUtility.assertProcFailed(procExec, procId);
 
     fs.delete(staleFile, false);
 
     UTIL.waitFor(30_000, 500, () -> {
-      RegionState.State a =
-        am.getRegionStates().getRegionStateNode(regionsToMerge[0]).getState();
-      RegionState.State b =
-        am.getRegionStates().getRegionStateNode(regionsToMerge[1]).getState();
+      RegionState.State a = am.getRegionStates().getRegionStateNode(regionsToMerge[0]).getState();
+      RegionState.State b = am.getRegionStates().getRegionStateNode(regionsToMerge[1]).getState();
       return a == RegionState.State.OPEN && b == RegionState.State.OPEN;
     });
   }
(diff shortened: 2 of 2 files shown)


apurtell (approved):

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
```diff
@@ -284,7 +284,6 @@ protected void rollbackState(final MasterProcedureEnv env, final MergeTableRegio
           cleanupMergedRegion(env);
           break;
         case MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS:
-          break;
```
Apache9, line 287: Why we need this change? In general, when rolling back a procedure, finally we will arrive the MERGE_TABLE_REGIONS_CLOSE_REGIONS state and reopen the parent regions?

Apache9 (changes requested):

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
```diff
@@ -284,9 +284,13 @@ protected void rollbackState(final MasterProcedureEnv env, final MergeTableRegio
           cleanupMergedRegion(env);
           break;
         case MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS:
+          rollbackCloseRegionsForMerge(env);
           break;
         case MERGE_TABLE_REGIONS_CLOSE_REGIONS:
-          rollbackCloseRegionsForMerge(env);
+          // If it rolls back with state MERGE_TABLE_REGIONS_CLOSE_REGIONS, no need to call
+          // rollbackCloseRegionsForMerge(), otherwise, it will result in duplicate
+          // TransitRegionStateProcedures for parents that are already OPENING. Mirrors
```
Apache9: Already opened?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
```diff
@@ -327,6 +333,80 @@ public void testRollbackAndDoubleExecution() throws Exception {
...
+      am.getRegionStates().getRegionStateNode(regionsToMerge[1]).getState();
+    LOG.info("HBASE-30334-DEBUG post-rollback: {} state={} / {} state={}",
+      regionsToMerge[0].getEncodedName(), s0Post,
+      regionsToMerge[1].getEncodedName(), s1Post);
+
+    ProcedureTestingUtility.assertProcFailed(procExec, procId);
+
+    fs.delete(staleFile, false);
+
+    long elapsed = org.apache.hadoop.hbase.Waiter.waitFor(conf, 30_000, 500, false,
```
Apache9: Just use UTIL.waitFor?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
```diff
@@ -327,6 +333,80 @@ public void testRollbackAndDoubleExecution() throws Exception {
...
+      new org.apache.hadoop.hbase.Waiter.Predicate<Exception>() {
```
Apache9: Use lambda?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
```diff
@@ -327,6 +333,80 @@ public void testRollbackAndDoubleExecution() throws Exception {
...
+        public boolean evaluate() {
+          RegionState.State a =
+            am.getRegionStates().getRegionStateNode(regionsToMerge[0]).getState();
+          RegionState.State b =
+            am.getRegionStates().getRegionStateNode(regionsToMerge[1]).getState();
+          return a == RegionState.State.OPEN && b == RegionState.State.OPEN;
+        }
+      });
+
+    assertTrue(elapsed > 0,
```
Apache9: What is the intention for asserting elapsed > 0?

Apache9 (commented):
I think the fix here is correct but the descrption of this PR as well as the jira does not reflection what we fix here.

And also the UT does not actually test what we fix here. According to your comment, what we fix is the MTRP fails in MERGE_TABLE_REGIONS_CLOSE_REGIONS state, but the test will make the MTRP fails at MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS state?

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
nirdosh0110: Good catch — I've dropped this change from the test case. It looks like it was leftover from local testing and got pushed inadvertently. Thanks for flagging!

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
nirdosh0110: Addressed in the latest commit — please take another look when you get a chance. Thanks!

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestMergeTableRegionsProcedure.java
nirdosh0110: Addressed in the latest commit — please take another look when you get a chance. Thanks!

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
nirdosh0110: Thanks for catching. Updated the comment.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/MergeTableRegionsProcedure.java
nirdosh0110: Not applicable now.

nirdosh0110:
^ bumping this up for reviews.

virajjasani:
What is the order for PR merges we are following for the relevant work @Apache9 @apurtell?

- https://github.com/apache/hbase/pull/8583
- https://github.com/apache/hbase/pull/8584
- https://github.com/apache/hbase/pull/8599

virajjasani:
> What is the order for PR merges we are following for the relevant work @Apache9 @apurtell?
> 
> * [HBASE-30334: Reopen parent regions on rollback of MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS #8583](https://github.com/apache/hbase/pull/8583)
> * [HBASE-30335 Seed master flushedSequenceIdByRegion with openSeqNum on region OPEN #8584](https://github.com/apache/hbase/pull/8584)
> * [HBASE-30352: Tolerate stale recovered.edits below durable seqid in split/merge #8599](https://github.com/apache/hbase/pull/8599)

@Apache9 we have run into this issue in prod again. Let's move forward with these PRs?
@nirdosh0110 what is the sequence of PRs to check-in one by one? I can take it forward accordingly.

nirdosh0110:
> > What is the order for PR merges we are following for the relevant work @Apache9 @apurtell?
> > 
> > * [HBASE-30334: Reopen parent regions on rollback of MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS #8583](https://github.com/apache/hbase/pull/8583)
> > * [HBASE-30335 Seed master flushedSequenceIdByRegion with openSeqNum on region OPEN #8584](https://github.com/apache/hbase/pull/8584)
> > * [HBASE-30352: Tolerate stale recovered.edits below durable seqid in split/merge #8599](https://github.com/apache/hbase/pull/8599)
> 
@nirdosh0110 what is the sequence of PRs to check-in one by one?

Hi @virajjasani,
- We'd need one review approval on #8599 before merging it in.
- There are no compile/test dependencies between the three PRs — they touch different files and different subsystems. However, there is a clear safety ordering we should follow. Suggested merge order:

1. #8583 — Rollback fix; prevents stuck RITs.
2. #8599 — Tolerate stale recovered.edits; provides immediate prod relief. Blocker: resolve Apache9's HFile metadata request first.
3. #8584 — Seed durability state on OPEN; addresses the root cause. This is the largest change, so it should be merged last.

Let me know your thoughts. Thanks!

virajjasani (approved):

PR #8584, opened 2026-08-28
HBASE-30335 Seed master flushedSequenceIdByRegion with openSeqNum on region OPEN

nirdosh0110:
## Summary
- The master's `flushedSequenceIdByRegion` is only updated via periodic RegionServer heartbeats. On a fresh region OPEN, the entry is absent, so `ServerManager.getLastFlushedSequenceId` returns `NO_SEQNUM` and `WALSplitter` treats every WAL edit as un-flushed. If the source RS of a drain-move crashes before its next heartbeat, this produces orphaned `recovered.edits` files containing already-durable edits, which then surface as false-positive "data loss" warnings during subsequent merge/split operations and leave regions stuck in RIT.
- Add `ServerManager.reportRegionOpen(regionInfo, openSeqNum)` and call it from `AssignmentManager.regionOpenedWithoutPersistingToMeta` so the master seeds its flushed-sequence cache synchronously with the `openSeqNum` reported on the OPEN transition.
- `putIfAbsent` is used so a heartbeat-supplied value (which may reflect flushes after open) is never regressed.

JIRA: https://issues.apache.org/jira/browse/HBASE-30335

## Test plan
- [ ] Existing `TestAssignmentManager` and `TestServerManager` suites pass.
- [ ] Reproduce: drain-move a region, kill source RS before next heartbeat, verify no `recovered.edits` are produced for already-flushed sequence numbers and no downstream stuck RIT on a follow-up merge.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
@@ -1098,14 +1098,16 @@ public class ServerManager implements ConfigurationObserver {
    * heartbeat, so {@link #getLastFlushedSequenceId} returns {@link HConstants#NO_SEQNUM} and
    * WALSplitter conservatively treats already-durable edits as unflushed - producing orphaned
    * recovered.edits when the source server crashes soon after a drain-move. Uses
-   * {@code putIfAbsent} so a heartbeat-supplied value (which may reflect flushes after open) is
-   * never regressed. See HBASE-30335.
+   * {@code merge} with {@link Math#max} so a heartbeat-supplied value (which may reflect flushes
+   * after open) is never regressed - and, unlike {@code putIfAbsent}, a stale-low prior value is
+   * lifted to {@code openSeqNum}. Safe because at OPEN a region cannot have flushed past its own
+   * {@code openSeqNum}. See HBASE-30335.
    */
   public void reportRegionOpen(final RegionInfo regionInfo, final long openSeqNum) {
-    if (openSeqNum == HConstants.NO_SEQNUM || openSeqNum < 0) {
+    if (openSeqNum < 0) { // NO_SEQNUM == -1
       return;
     }
-    flushedSequenceIdByRegion.putIfAbsent(regionInfo.getEncodedNameAsBytes(), openSeqNum);
+    flushedSequenceIdByRegion.merge(regionInfo.getEncodedNameAsBytes(), openSeqNum, Math::max);
   }
 
   public boolean isRegionInServerManagerStates(final RegionInfo hri) {
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
@@ -1097,10 +1097,10 @@ public class ServerManager implements ConfigurationObserver {
    * {@code openSeqNum}. Without this, the entry stays absent until the hosting server's next
    * heartbeat, so {@link #getLastFlushedSequenceId} returns {@link HConstants#NO_SEQNUM} and
    * WALSplitter conservatively treats already-durable edits as unflushed - producing orphaned
-   * recovered.edits when the source server crashes soon after a drain-move. Uses
-   * {@code merge} with {@link Math#max} so a heartbeat-supplied value (which may reflect flushes
-   * after open) is never regressed - and, unlike {@code putIfAbsent}, a stale-low prior value is
-   * lifted to {@code openSeqNum}. Safe because at OPEN a region cannot have flushed past its own
+   * recovered.edits when the source server crashes soon after a drain-move. Uses {@code merge} with
+   * {@link Math#max} so a heartbeat-supplied value (which may reflect flushes after open) is never
+   * regressed - and, unlike {@code putIfAbsent}, a stale-low prior value is lifted to
+   * {@code openSeqNum}. Safe because at OPEN a region cannot have flushed past its own
    * {@code openSeqNum}. See HBASE-30335.
    */
   public void reportRegionOpen(final RegionInfo regionInfo, final long openSeqNum) {
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
@@ -1092,6 +1092,22 @@ public class ServerManager implements ConfigurationObserver {
     flushedSequenceIdByRegion.remove(encodedName);
   }
 
+  /**
+   * Called on region OPEN to seed {@link #flushedSequenceIdByRegion} with the region's
+   * {@code openSeqNum}. Without this, the entry stays absent until the hosting server's next
+   * heartbeat, so {@link #getLastFlushedSequenceId} returns {@link HConstants#NO_SEQNUM} and
+   * WALSplitter conservatively treats already-durable edits as unflushed - producing orphaned
+   * recovered.edits when the source server crashes soon after a drain-move. Uses
+   * {@code putIfAbsent} so a heartbeat-supplied value (which may reflect flushes after open) is
+   * never regressed. See HBASE-30335.
+   */
+  public void reportRegionOpen(final RegionInfo regionInfo, final long openSeqNum) {
+    if (openSeqNum == HConstants.NO_SEQNUM || openSeqNum < 0) {
+      return;
+    }
+    flushedSequenceIdByRegion.putIfAbsent(regionInfo.getEncodedNameAsBytes(), openSeqNum);
+  }
+
   public boolean isRegionInServerManagerStates(final RegionInfo hri) {
     final byte[] encodedName = hri.getEncodedNameAsBytes();
     return (storeFlushedSequenceIdsByRegion.containsKey(encodedName)
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
@@ -2287,6 +2287,10 @@ public class AssignmentManager {
     RegionInfo regionInfo = regionNode.getRegionInfo();
     regionStates.addRegionToServer(regionNode);
     regionStates.removeFromFailedOpen(regionInfo);
+    // HBASE-30335: seed the master's flushed sequence cache with openSeqNum so a subsequent
+    // WAL split (e.g. source RS crashes after drain-move) recognizes already-durable edits
+    // instead of writing orphaned recovered.edits.
+    master.getServerManager().reportRegionOpen(regionInfo, regionNode.getOpenSeqNum());
   }
 
   // should be called under the RegionStateNode lock
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestGetLastFlushedSequenceId.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestGetLastFlushedSequenceId.java
@@ -31,6 +31,7 @@ import org.apache.hadoop.hbase.SingleProcessHBaseCluster;
 import org.apache.hadoop.hbase.TableName;
 import org.apache.hadoop.hbase.client.Put;
 import org.apache.hadoop.hbase.client.Table;
+import org.apache.hadoop.hbase.regionserver.HRegion;
 import org.apache.hadoop.hbase.regionserver.HRegionServer;
 import org.apache.hadoop.hbase.regionserver.Region;
 import org.apache.hadoop.hbase.testclassification.MediumTests;
@@ -106,5 +107,42 @@ public class TestGetLastFlushedSequenceId {
       ids.getLastFlushedSequenceId() + " > " + storeSequenceId);
     assertEquals(ids.getLastFlushedSequenceId(), ids.getStoreSequenceId(0).getSequenceId());
     table.close();
+  }
+
+  /**
+   * HBASE-30335: after a region is opened - and before any user write or flush - the master's
+   * flushedSequenceIdByRegion must already contain the region's openSeqNum. Otherwise a
+   * subsequent WAL split (e.g. the hosting RS crashes before its first flush heartbeat) would
+   * treat already-durable edits as unflushed and produce orphaned recovered.edits.
+   */
+  @Test
+  public void testFlushedSequenceIdSeededOnRegionOpen() throws IOException, InterruptedException {
+    TableName freshTable = TableName.valueOf(getClass().getSimpleName(), "openseed");
+    testUtil.getAdmin()
+      .createNamespace(NamespaceDescriptor.create(freshTable.getNamespaceAsString()).build());
+    Table table = testUtil.createTable(freshTable, families);
+    try {
+      SingleProcessHBaseCluster cluster = testUtil.getMiniHBaseCluster();
+      HRegion region = null;
+      for (JVMClusterUtil.RegionServerThread rst : cluster.getRegionServerThreads()) {
+        for (HRegion r : rst.getRegionServer().getRegions(freshTable)) {
+          region = r;
+          break;
+        }
+        if (region != null) {
+          break;
+        }
+      }
+      assertNotNull(region);
+      long openSeqNum = region.getOpenSeqNum();
+      RegionStoreSequenceIds ids = testUtil.getHBaseCluster().getMaster().getServerManager()
+        .getLastFlushedSequenceId(region.getRegionInfo().getEncodedNameAsBytes());
+      assertNotEquals(HConstants.NO_SEQNUM, ids.getLastFlushedSequenceId(),
+        "flushedSequenceIdByRegion should be seeded on region OPEN (HBASE-30335)");
+      assertEquals(openSeqNum, ids.getLastFlushedSequenceId(),
+        "seeded value must equal the region's openSeqNum");
+    } finally {
+      table.close();
+    }
   }
 }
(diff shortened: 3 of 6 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
```diff
@@ -2287,6 +2287,10 @@ void regionOpenedWithoutPersistingToMeta(RegionStateNode regionNode)
     RegionInfo regionInfo = regionNode.getRegionInfo();
     regionStates.addRegionToServer(regionNode);
     regionStates.removeFromFailedOpen(regionInfo);
+    // HBASE-30335: seed the master's flushed sequence cache with openSeqNum so a subsequent
+    // WAL split (e.g. source RS crashes after drain-move) recognizes already-durable edits
+    // instead of writing orphaned recovered.edits.
```
Umeshkumar9414, line 2292: It would be good to add tests.

nirdosh0110: Ack

nirdosh0110: @Umeshkumar9414 Added a unit test.

Apache9:
I think this is an optimization for not splitting unnecessary wal edits, but why it will lead to merge stuck? After opening a region, we should have removed all the recovered.edits files?

nirdosh0110:
> I think this is an optimization for not splitting unnecessary wal edits, but why it will lead to merge stuck? After opening a region, we should have removed all the recovered.edits files?
 
Thanks @Apache9 for looking into it. Below is incident brief explanation. I've also raised [PR8583](https://github.com/apache/hbase/pull/8583)  to reopen parent region on rollback of MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS linked to this incident. 

What Happened: 
1. Region 112d9f08 was gracefully moved from rs-132 → rs-81 at 16:36:34 UTC.
2. The close on rs-132 had already flushed the region's edits, and rs-81 opened the region with openSeqNum=4997750282, establishing that the edits up to that point were durable.
3. ~25 seconds later, rs-132 was declared dead as part of the broader graceful RS drain.
4. The WAL split worker (rs-34) subsequently created a recovered.edits file for this region containing edit seqId=4997750280.
5. This edit was already durable in the HFiles. The recovered edit file was effectively stale/orphaned, but its existence was not recognized as harmless.


Where it got stuck:

At 18:32:33 UTC, MergeTableRegionsProcedure (pid=46990253) unassigned the region successfully.
During MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS, the procedure saw the recovered.edits file and failed the check.
The merge procedure retained the region lock and did not recover automatically, leaving the region in CLOSED/RIT for ~48m 56s.
The RIT only cleared after the HMaster failover caused the procedure to replay and fresh top-level ASSIGN procedures reopened the regions.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
```diff
@@ -1092,6 +1092,22 @@ public void removeRegion(final RegionInfo regionInfo) {
...
+   * WALSplitter conservatively treats already-durable edits as unflushed - producing orphaned
+   * recovered.edits when the source server crashes soon after a drain-move. Uses
+   * {@code putIfAbsent} so a heartbeat-supplied value (which may reflect flushes after open) is
+   * never regressed. See HBASE-30335.
+   */
+  public void reportRegionOpen(final RegionInfo regionInfo, final long openSeqNum) {
+    if (openSeqNum == HConstants.NO_SEQNUM || openSeqNum < 0) {
+      return;
+    }
+    flushedSequenceIdByRegion.putIfAbsent(regionInfo.getEncodedNameAsBytes(), openSeqNum);
```
apurtell: A max merge is more correct and equally safe, because at OPEN time a region cannot have flushed past its own `openSeqNum`.

```java
public void reportRegionOpen(final RegionInfo regionInfo, final long openSeqNum) {
  if (openSeqNum < 0) { // NO_SEQNUM == -1
    return;
  }
  flushedSequenceIdByRegion.merge(regionInfo.getEncodedNameAsBytes(), openSeqNum, Math::max);
}
```

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestGetLastFlushedSequenceId.java
```diff
@@ -89,10 +91,14 @@ public void test() throws IOException, InterruptedException {
...
       .getLastFlushedSequenceId(region.getRegionInfo().getEncodedNameAsBytes());
-    assertEquals(HConstants.NO_SEQNUM, ids.getLastFlushedSequenceId());
     // This will be the sequenceid just before that of the earliest edit in memstore.
     long storeSequenceId = ids.getStoreSequenceId(0).getSequenceId();
     assertTrue(storeSequenceId > 0);
+    // HBASE-30335: openSeqNum is now seeded on region OPEN, so lastFlushedSequenceId is no
+    // longer NO_SEQNUM before the first flush - it is the region's openSeqNum, which must
+    // still be strictly less than the memstore's earliest unflushed edit.
+    assertNotEquals(HConstants.NO_SEQNUM, ids.getLastFlushedSequenceId());
+    assertTrue(ids.getLastFlushedSequenceId() < storeSequenceId);
```
apurtell: This assertion is fragile. 
It is technically off by one, but because the region open marker consumes one seq id, the test will currently pass.

Maybe `assertNotEquals(NO_SEQNUM, ...)` instead?

apurtell (approved):
Approved, with two suggestions

Apache9:
> > I think this is an optimization for not splitting unnecessary wal edits, but why it will lead to merge stuck? After opening a region, we should have removed all the recovered.edits files?
> 
> Thanks @Apache9 for looking into it. Below is incident brief explanation. I've also raised [PR8583](https://github.com/apache/hbase/pull/8583) to reopen parent region on rollback of MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS linked to this incident.
> 
> What Happened:
> 
> 1. Region 112d9f08 was gracefully moved from rs-132 → rs-81 at 16:36:34 UTC.
> 2. The close on rs-132 had already flushed the region's edits, and rs-81 opened the region with openSeqNum=4997750282, establishing that the edits up to that point were durable.
> 3. ~25 seconds later, rs-132 was declared dead as part of the broader graceful RS drain.
> 4. The WAL split worker (rs-34) subsequently created a recovered.edits file for this region containing edit seqId=4997750280.
> 5. This edit was already durable in the HFiles. The recovered edit file was effectively stale/orphaned, but its existence was not recognized as harmless.
> 
> Where it got stuck:
> 
> At 18:32:33 UTC, MergeTableRegionsProcedure (pid=46990253) unassigned the region successfully. During MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS, the procedure saw the recovered.edits file and failed the check. The merge procedure retained the region lock and did not recover automatically, leaving the region in CLOSED/RIT for ~48m 56s. The RIT only cleared after the HMaster failover caused the procedure to replay and fresh top-level ASSIGN procedures reopened the regions.

Then basically there are two problems.
On 5, we should remove the recovered.edits when opening the region, of course a failure of removing should not be considered as a critical issue.
And in MergeTableRegionsProcedure, when we have a recovered.edits file, we should check if the edits are all below the persistent seqNum, if so we are OK to remove the directory and go on.

We can do these changes in a separated issue.

Thanks.

Umeshkumar9414:
> Then basically there are two problems.
> On 5, we should remove the recovered.edits when opening the region, of course a failure of removing should not be considered as a critical issue.

We should check why the recovered edit files created after region open. BTW I am aware of onc case that I faced (But it looks like you faced another one) when two split worker (one zombie server and another active one) is simultaneously splitting the wal. Active one complete and we even open the region. Now zombie server can try to create one recovered-edit file from its memory. 

> And in MergeTableRegionsProcedure, when we have a recovered.edits file, we should check if the edits are all below the persistent seqNum, if so we are OK to remove the directory and go on.

I might not be right here but what I understand from the code is having recovered.edits files once region is open is unacceptable, and I think that was right. Even if we check that the edits are all below the persistent seqNum, we might not be sure that we didn't miss some edits. But we might not have any other way here, anyway rollback and open will do the same.

nirdosh0110:
> MergeTableRegionsProcedure, when we have a recovered.edits file, we should check if the edits are all below the persistent seqNum, if so we are OK to remove the directory and go on.

@Apache9 Totally agree — we should have tolerance built into the merge and split procedures. I've filed a separate bug to track the issue pointed 
- https://issues.apache.org/jira/browse/HBASE-30352
- https://github.com/apache/hbase/pull/8599

hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
nirdosh0110: Addressed this comment in latest commit.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/TestGetLastFlushedSequenceId.java
nirdosh0110: Addressed this comment in latest commit.

mnpoonia (approved):

Umeshkumar9414:
Can we do this while closing of the region instread of opening the region? 

Looks like SCP just after the region closing is creating these recovered.edits files as RS report the flushed seq id in it heartbeats. Corrent me if I am wrong.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/ServerManager.java
```diff
@@ -1092,6 +1092,24 @@ public void removeRegion(final RegionInfo regionInfo) {
...
+   * {@link Math#max} so a heartbeat-supplied value (which may reflect flushes after open) is never
+   * regressed - and, unlike {@code putIfAbsent}, a stale-low prior value is lifted to
+   * {@code openSeqNum}. Safe because at OPEN a region cannot have flushed past its own
+   * {@code openSeqNum}. See HBASE-30335.
+   */
+  public void reportRegionOpen(final RegionInfo regionInfo, final long openSeqNum) {
+    if (openSeqNum < 0) { // NO_SEQNUM == -1
+      return;
+    }
+    flushedSequenceIdByRegion.merge(regionInfo.getEncodedNameAsBytes(), openSeqNum, Math::max);
```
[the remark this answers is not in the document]
nirdosh0110: Good catch — the race is real. `updateLastFlushedSequenceIds` reads `flushedSequenceIdByRegion` at line 292 and does a conditional `put` at line 300 without holding any lock, so the following interleaving is possible:

1. Stale heartbeat thread reads `existingValue = null`.
2. `reportRegionOpen(regionInfo, openSeqNum=10)` executes its atomic `merge(Math::max)` → map now has `10`.
3. Stale heartbeat thread resumes and writes `completedSeqId = 3` → map regresses to `3`.

The read-then-put pattern predates this PR, but HBASE-30335 sharpens the exposure. Before this change every writer to the map was the same heartbeat path carrying monotonically-nondecreasing values from a single RS — a lost update was self-healing on the next heartbeat. With the new OPEN-time seed, the writer values are heterogeneous (a fresh `openSeqNum` after reopen can be strictly greater than any stale RS's `completedSequenceId` observed during graceful drain / failover), and a regression on this path is not self-healing because the OPEN seed happens once per region open, not periodically.

Filing this as a follow-up JIRA to keep this PR's diff focused on the seed installation and to make bisection precise if either change regresses. The follow-up will convert both the region-level and per-store read-then-put pairs to atomic `compute(...)` and add a concurrency test that races `reportRegionOpen(high)` against a heartbeat-shaped `regionServerReport(low)` and asserts the final map holds `high`. Will link the follow-up PR here once it's up.

nirdosh0110:
> Can we do this while closing of the region instread of opening the region?
> 
> Looks like SCP just after the region closing is creating these recovered.edits files as RS report the flushed seq id in it heartbeats. Corrent me if I am wrong.

CLOSE-time seeding alone doesn't cover the crash path — if the RS dies, there's no close report to seed from, so the watermark stays stale and SCP walks with the wrong fence.

Doing it at OPEN time covers both cases: a graceful close-then-reopen and a crash-then-reopen-elsewhere both go through OPEN, and `openSeqNum` at that point is already bumped past the close marker, so it's the strongest fence we can install.

Umeshkumar9414:
> CLOSE-time seeding alone doesn't cover the crash path — if the RS dies, there's no close report to seed from, so the watermark stays stale and SCP walks with the wrong fence.

If there is no close report then master won't mvoe ahead with open so we are good in that case. We are only having a problem if edits are present after open.

nirdosh0110:
Fair — for a pure server crash we'd have stale data anyway, bounded by `msginterval`. RS already sends `completedSequenceId` for every region on every heartbeat, so master's map is continuously refreshed on the happy path, and pure-crash isn't really the differentiator between CLOSE-time and OPEN-time.

The scenario this PR is targeting is *graceful move + later crash of the source RS*:

1. `t=0` — source RS cleanly closes R (flush to close-marker). Target RS opens R with `openSeqNum` bumped past close-marker. But master's `flushedSequenceIdByRegion[R]` still reflects the last heartbeat from the source RS, which was written **before** the close-flush landed — so it's stale.
2. `t=25s` — source RS dies. SCP walks its WAL and filters using `flushedSequenceIdByRegion[R]` — still the stale pre-close value. Recovered.edits are written for edits that are already durable in HFiles.
3. Later split/merge on R fails `checkClosedRegion` on those stale recovered.edits.

Both CLOSE-time and OPEN-time seeding would fix this. Reasons I kept OPEN-time:
- `openSeqNum` ≥ close-marker, so it's an at-least-as-tight fence.
- Fires on every OPEN (reassignment after SCP, master restart, RS restart), not only clean closes.
- Reuses the existing `reportRegionOpen` payload that already carries `openSeqNum` — no new field on the close report.

apurtell:
Copilot's review (2026-09-02, reiterated 2026-09-16) highlights a real race. This has to be fixed now, not deferred to another followup, otherwise what is the point. A stale in-flight heartbeat from the (soon to be dead) source RS can read null/a low value, race past the OPEN-seed, and then overwrite it with its lower value, reintroducing the bug this PR is meant to fix, just in a narrower window.

nirdosh0110:
> Copilot's review (2026-09-02, reiterated 2026-09-16) highlights a real race. This has to be fixed now, not deferred to another followup, otherwise what is the point. A stale in-flight heartbeat from the (soon to be dead) source RS can read null/a low value, race past the OPEN-seed, and then overwrite it with its lower value, reintroducing the bug this PR is meant to fix, just in a narrower window.

@apurtell Got it. The race was in updateLastFlushedSequenceIds: it did a non-atomic get-then-put, so a stale in-flight heartbeat could read null/low, race past the reportRegionOpen seed, and put its lower value on top.

Both the region- and store-level updates now go through an atomic compute that keeps the existing "never lower the watermark" rule. With the merge(Math::max) seed on the OPEN side, every writer on the live serving path is now an atomic max-merge — the stored value is monotonic non-decreasing, so a concurrent seed can no longer be clobbered. Whatever the interleaving: heartbeat-before-seed → the seed lifts it to openSeqNum; heartbeat-after-seed → compute refuses to lower it.

Added TestServerManager#testConcurrentStaleHeartbeatDoesNotClobberOpenSeed to test the scenerio.

Umeshkumar9414:
> Both CLOSE-time and OPEN-time seeding would fix this. Reasons I kept OPEN-time:
> 
> * `openSeqNum` ≥ close-marker, so it's an at-least-as-tight fence.
> * Fires on every OPEN (reassignment after SCP, master restart, RS restart), not only clean closes.
> * Reuses the existing `reportRegionOpen` payload that already carries `openSeqNum` — no new field on the close report.

But this will still leave a small gap where Close is completed and OPEN is yet to happen. And SCP of source is already started and we are about to split the wal. This is the reason why I wanted to do it with `reportRegionClose` . Because flush is something related to close only not open. So it would make sense if we do this while closing.

nirdosh0110:
> > Both CLOSE-time and OPEN-time seeding would fix this. Reasons I kept OPEN-time:
> > 
> > * `openSeqNum` ≥ close-marker, so it's an at-least-as-tight fence.
> > * Fires on every OPEN (reassignment after SCP, master restart, RS restart), not only clean closes.
> > * Reuses the existing `reportRegionOpen` payload that already carries `openSeqNum` — no new field on the close report.
> 
> But this will still leave a small gap where Close is completed and OPEN is yet to happen. And SCP of source is already started and we are about to split the wal. This is the reason why I wanted to do it with `reportRegionClose` . Because flush is something related to close only not open. So it would make sense if we do this while closing.

Good point @Umeshkumar9414, and I agree it's worth doing. The OPEN-time seed here is the primary fence — openSeqNum ≥ close-marker, and it fires on every reopen path (reassignment after SCP, master restart, RS restart), including the pure-crash case where there's no close report at all. But you're right that it leaves the narrow window you describe: close done on the source, source dies, its SCP splits the WAL before the target finishes OPEN.

Seeding on CLOSE tightens exactly that window for the graceful-close case, and it composes cleanly with the OPEN seed since both go through the same monotonic merge(Math::max). The plumbing is already there — the CLOSED transition is routed through the same handler that receives a seqId, the RS can source it from HRegion.getMaxFlushedSeqId(), and RegionStateTransition already has the optional openSeqNum field, so no proto change is needed.

I'd prefer to keep this PR's diff focused on the OPEN-time seed (it's approved and the heartbeat race is now fixed) and do the close-time seed as a small follow-up — same way we split out HBASE-30352. I will file a HBase JIRA; I'll pick it up there.

PR #8586, opened 2026-08-28
HBASE-30287 Filter offline and split regions from SnapshotRegionLocator (#8576)

wchevreuil:


wchevreuil:
The validations passed. Closing this and merging manually into base 2 branches to preserve proper authorship.

PR #8592, opened 2026-08-31
HBASE-30349 MobFileCleanupUtil: use human-readable date instead of timestamp in logs

guluo2016:
Details see: HBASE-30349


liuxiaocs7 (approved):

guluo2016:
Thanks @liuxiaocs7 for the review.

PR #8594, opened 2026-08-31
HBASE-30350 Fix stale Region-In-Transition tracker entries on region deletion

mnpoonia:
## Summary

`RegionInTransitionTracker` and `RegionStates` are two separate in-memory maps on the master that are not kept in sync automatically. Several code paths deleted a region from `RegionStates` directly without notifying the tracker, so a deleted region could be left behind in the tracker forever — surfacing as a permanently stuck "STUCK Region-In-Transition" warning that only clears on a full master restart (`loadMeta()` rebuilds the tracker from `hbase:meta` from scratch; a GC'd region isn't in meta, so no stale entry survives that path).

This was the root cause of a 9.8h stuck RIT incident: `GCRegionProcedure` deleted a no-longer-needed region from `RegionStates` but never told the tracker, leaving a stale entry that nothing but a restart could clear.

**Fix:** add `AssignmentManager#deleteRegion(RegionInfo)` and its plural form `#deleteRegions(List<RegionInfo>)` as the single sync point that updates both maps together, and route every region-deletion call site through it instead of calling `RegionStates#deleteRegion(s)` directly.

**Confirmed live bug, reproduced end-to-end:** `GCRegionProcedure`. Temporarily reverting its call site back to a direct `RegionStates#deleteRegion()` call reproduces the original incident symptom in `testGCRegionProcedureRemovesStaleRegionInTransitionEntry` (the tracker entry survives GC); restoring the fix makes the same test pass again.

**Plausible but unconfirmed live bug:** `AssignmentManager#deleteTable`. `DeleteTableProcedure` waits for `RegionStates` to clear RIT before calling it, but that wait only checks `RegionStates`, not the tracker — so it doesn't rule out a stale tracker entry, though it isn't proven reachable either.

**Defensive hardening, not confirmed reachable today** (the region being deleted at these sites is already untracked — either terminal/OFFLINE on a disabled table, or a non-default replica, which the tracker never tracks in the first place):
- `AssignmentManagerUtil#removeNonDefaultReplicas` (split/merge)
- `EnableTableProcedure` (replica-count-decrease cleanup)
- `RestoreSnapshotProcedure#deleteRegionsFromInMemoryStates`

## Test plan

- `testDeleteRegionRemovesStaleRegionInTransitionEntry` / `testDeleteRegionsRemovesStaleRegionInTransitionEntries` — cover the singular/plural `AssignmentManager#deleteRegion(s)` sync point directly.
- `testGCRegionProcedureRemovesStaleRegionInTransitionEntry` — runs the real `GCRegionProcedure` end-to-end through the procedure executor. Verified this test fails against the pre-fix code path and passes against the fix.
- Ran `TestAssignmentManager`, `TestAssignmentManagerUtil`, `TestEnableTableProcedure`, `TestRestoreSnapshotProcedure`, and related suites — all green, no regressions.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
@@ -1212,7 +1212,7 @@ public class AssignmentManager {
     regionStateStore.deleteRegions(regions);
     for (int i = 0; i < regions.size(); ++i) {
       final RegionInfo regionInfo = regions.get(i);
-      regionStates.deleteRegion(regionInfo);
+      deleteRegion(regionInfo);
     }
   }
 
@@ -2409,8 +2409,7 @@ public class AssignmentManager {
     RegionInfo[] mergeParents) throws IOException {
     final RegionStateNode node = regionStates.getOrCreateRegionStateNode(child);
     for (RegionInfo ri : mergeParents) {
-      regionStates.deleteRegion(ri);
-      regionInTransitionTracker.handleRegionDelete(ri);
+      deleteRegion(ri);
     }
 
     TableDescriptor td = master.getTableDescriptors().get(child.getTable());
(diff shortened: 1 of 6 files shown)


mnpoonia:
Looking for approvals for running the github actions workflow
@Apache9 @virajjasani

Umeshkumar9414:
We only delete an region once it is closed. I can think of 3 scenario of region close- 1. Table disabled 2. Merge parents 3. Split parent . In all 3 cases, regions should be out of Region-In-trnasition tracker when region was getting closed so we should never need to handle it while deleteing. 

I think bug here, that we need to find out, is why it was not removed from tracker while closing. I am opposed to the change in tracker when actually there is no state change.

mnpoonia:
Thank you @Umeshkumar9414 You are absolutely right. It is the case of missing https://issues.apache.org/jira/browse/HBASE-30169 from the code.

I will close this.

PR #8596, opened 2026-09-01
HBASE-30351 ROWS_SCANNED scan metric double-counts rows passing the filter when the scan uses the joined heap

eomiks:
Fixes the ROWS_SCANNED server-side scan metric double-counting rows that pass the filter when the scan uses the joined heap, i.e. `Scan#setLoadColumnFamiliesOnDemand(true)` with a filter marking some families as non-essential (`Filter#isFamilyEssential`).

`RegionScannerImpl#populateResult` increments the metric whenever the given heap finishes the current row, and `populateFromJoinedHeap` goes through the same method for the non-essential families, so a row passing the filter was counted twice (once for the store heap, once for the joined heap). The row-scanned event is now recorded only when populating from the store heap.

The new test in `TestServerSideScanMetricsFromClientSide` reproduces the issue with stock filters (`SingleColumnValueFilter` with `setFilterIfMissing(true)` on a two-family table): before the fix ROWS_SCANNED reported 15 for a scan over 10 rows with 5 matching, after the fix it reports 10.

Details: https://issues.apache.org/jira/browse/HBASE-30351


junegunn (approved):
LGTM, thanks!

PR #8599, opened 2026-09-01
HBASE-30352: Tolerate stale recovered.edits below durable seqid in split/merge

nirdosh0110:
## Context

Follow-up to the discussion on #8584 (HBASE-30335). @Apache9 suggested in https://github.com/apache/hbase/pull/8584#issuecomment-5487364435:

> In `MergeTableRegionsProcedure`, when we have a `recovered.edits` file, we should check if the edits are all below the persistent seqNum, if so we are OK to remove the directory and go on.

This PR implements that check for both `MergeTableRegionsProcedure` and `SplitTableRegionProcedure` (both call `AssignmentManagerUtil.checkClosedRegion`).

JIRA: https://issues.apache.org/jira/browse/HBASE-30352

## The incident this addresses

1. Region `112d9f08` was gracefully moved from RS-A → RS-B.
2. RS-B opened the region with `openSeqNum=4997750282`; all prior edits were durable in HFiles.
3. ~25 s later, RS-A was declared dead and its WAL was split.
4. The split worker produced a `recovered.edits` file for this region containing `seqId=4997750280` — an edit already flushed on HFile before RS-A closed.
5. `MergeTableRegionsProcedure` later hit `MERGE_TABLE_REGIONS_CHECK_CLOSED_REGIONS`, saw the `recovered.edits` file, and threw. The region sat in CLOSED/RIT for ~49 min until master failover cleared it.

## Change

`AssignmentManagerUtil.checkClosedRegion` now, when `hasRecoveredEdits` is true:

1. Reads the region's `lastFlushedSequenceId` from `ServerManager`.
2. For each recovered.edits file in every candidate location (WAL region dir, root region dir, legacy "wrong" WAL region dir), parses the filename — which is `formatRecoveredEditsFileName(maxEditWALSeqNum)` — to get the file's max seqid.
3. If every file's max seqid ≤ `lastFlushedSequenceId`, deletes those specific files and returns.
4. Otherwise the pre-existing abort behavior is preserved as a safe fallback (unknown/no durable info, or a file that could contain non-durable edits).

Filename parsing avoids opening/reading the WAL edits; the writer contract already encodes the max seqid in the file name via `WALSplitUtil.getCompletedRecoveredEditsFilePath`.

## Files

- `hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java` — the tolerance logic.
- `hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestAssignmentManagerUtil.java` — new test `testCheckClosedRegionDropsStaleRecoveredEdits` verifies (a) a stale file is dropped and the check passes, (b) a file with a fresh seqid still causes the abort.

## Local run

```
TestAssignmentManagerUtil - 3 tests / 0 failures / 0 errors / 0 skipped - 15.14 s
```

## Notes / open questions

- The improvement only takes effect when `ServerManager` has an authoritative `lastFlushedSequenceId` for the region. With HBASE-30335 landing, this will be the case immediately after region OPEN. Without HBASE-30335 it kicks in after the first flush heartbeat. In either case, the fallback matches today's behavior.
- Not implemented here (Apache9's other suggested item): proactively removing stale `recovered.edits` on region OPEN. Happy to file/pick that up separately if reviewers agree it should be in scope.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
@@ -24,10 +24,14 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 import java.util.ListIterator;
+import java.util.NavigableSet;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
 import java.util.stream.Stream;
 import org.apache.commons.lang3.ArrayUtils;
+import org.apache.hadoop.conf.Configuration;
+import org.apache.hadoop.fs.FileSystem;
+import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.HBaseIOException;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.hadoop.hbase.ServerName;
@@ -37,9 +41,13 @@ import org.apache.hadoop.hbase.client.RegionReplicaUtil;
 import org.apache.hadoop.hbase.favored.FavoredNodesManager;
 import org.apache.hadoop.hbase.master.RegionState;
 import org.apache.hadoop.hbase.master.procedure.MasterProcedureEnv;
+import org.apache.hadoop.hbase.util.CommonFSUtils;
+import org.apache.hadoop.hbase.util.FSUtils;
 import org.apache.hadoop.hbase.util.FutureUtils;
 import org.apache.hadoop.hbase.wal.WALSplitUtil;
 import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.apache.hadoop.hbase.shaded.protobuf.RequestConverter;
 import org.apache.hadoop.hbase.shaded.protobuf.generated.AdminProtos.GetRegionInfoRequest;
@@ -50,6 +58,7 @@ import org.apache.hadoop.hbase.shaded.protobuf.generated.AdminProtos.GetRegionIn
  */
 @InterfaceAudience.Private
 final class AssignmentManagerUtil {
+  private static final Logger LOG = LoggerFactory.getLogger(AssignmentManagerUtil.class);
   private static final int DEFAULT_REGION_REPLICA = 1;
 
   private AssignmentManagerUtil() {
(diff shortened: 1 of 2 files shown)


nirdosh0110:
@Apache9 @apurtell @virajjasani Please help with review on this PR. TIA.

virajjasani:
@nirdosh0110 I just realized this PR also points to HBASE-30335, we will need new Jira for this PR, given the 1:1 mapping that we follow for each Jira/PR.

nirdosh0110:
> @nirdosh0110 I just realized this PR also points to [HBASE-30335](https://issues.apache.org/jira/browse/HBASE-30335), we will need new Jira for this PR, given the 1:1 mapping that we follow for each Jira/PR.

Thanks @virajjasani. Some how this PR is not auto linked to open jira. Below are the details. I've updated the PR title. 
- [HBASE-30352](https://issues.apache.org/jira/browse/HBASE-30352)  PR: https://github.com/apache/hbase/pull/8599

nirdosh0110:
> > @nirdosh0110 I just realized this PR also points to [HBASE-30335](https://issues.apache.org/jira/browse/HBASE-30335), we will need new Jira for this PR, given the 1:1 mapping that we follow for each Jira/PR.
> 
> Thanks @virajjasani. Some how this PR is not auto linked to open jira. Below are the details. I've updated the PR title.
> 
>     * [HBASE-30352](https://issues.apache.org/jira/browse/HBASE-30352)  PR: [HBASE-30352: Tolerate stale recovered.edits below durable seqid in split/merge #8599](https://github.com/apache/hbase/pull/8599)

Now it is linked properly.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
```diff
@@ -294,10 +303,79 @@ static void removeNonDefaultReplicas(MasterProcedureEnv env, Stream<RegionInfo>
   }
 
   static void checkClosedRegion(MasterProcedureEnv env, RegionInfo regionInfo) throws IOException {
-    if (WALSplitUtil.hasRecoveredEdits(env.getMasterConfiguration(), regionInfo)) {
-      throw new IOException("Recovered.edits are found in Region: " + regionInfo
-        + ", abort split/merge to prevent data loss");
+    if (!WALSplitUtil.hasRecoveredEdits(env.getMasterConfiguration(), regionInfo)) {
+      return;
     }
+    // A recovered.edits file whose max seqid is <= the region's last flushed seqid is stale:
```
Apache9: I prefer we do not mention the details about how we can generate stale recoverd.edits, as in general, we should fix the problem.

Just add comment to say that this is for making the process more robust, there may be some corner cases where we left stale recovered.edits.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
```diff
@@ -294,10 +303,79 @@ static void removeNonDefaultReplicas(MasterProcedureEnv env, Stream<RegionInfo>
...
+  }
+
+  /**
+   * Try to remove recovered.edits files that are provably below the region's last flushed seqid.
+   * @return true if, after cleanup, no recovered.edits remain for the region
+   */
+  private static boolean tryDropStaleRecoveredEdits(MasterProcedureEnv env, RegionInfo regionInfo) {
+    long durableSeqId = env.getMasterServices().getServerManager()
+      .getLastFlushedSequenceId(regionInfo.getEncodedNameAsBytes()).getLastFlushedSequenceId();
+    if (durableSeqId <= 0L) {
```
Apache9, line 325: We can get this value by reading all the HFiles' metadata?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
```diff
@@ -294,10 +303,79 @@ static void removeNonDefaultReplicas(MasterProcedureEnv env, Stream<RegionInfo>
...
+    NavigableSet<Path> files = WALSplitUtil.getSplitEditFilesSorted(fs, regionDir);
+    if (files.isEmpty()) {
+      return true;
+    }
+    for (Path p : files) {
+      long fileMaxSeqId;
+      try {
+        fileMaxSeqId = Long.parseLong(p.getName());
+      } catch (NumberFormatException e) {
+        LOG.warn("Non-numeric recovered.edits filename {} for {}; not dropping", p, regionInfo);
```
Apache9: Does this really hurt? When loading recovered.edits, we will abort or just skip these files?

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
nirdosh0110: Done — reframed as a generic robustness note without the incident-specific detail in c7e4623.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
nirdosh0110: Good point. The reason for going with `ServerManager` here was consistency — SCP consults the same map when deciding what enters `recovered.edits` in the first place, so both decisions come from one source and cannot disagree. That said, HFile metadata is stronger: it is the on-disk truth and would decouple this check from HBASE-30335's seed timing (and from a cold `ServerManager` cache right after master restart).

Reading it correctly means `min across families of max HFile seqid` (the same fence `HRegion.replayRecoveredEditsIfAny` uses when skipping already-flushed edits). Roughly: `HRegionFileSystem.openRegionFromFileSystem` → per family `StoreFileTracker.load()` → wrap each `StoreFileInfo` in `HStoreFile` with `CacheConfig.DISABLED` → `getReader().getSequenceID()`. Same pattern `MergeTableRegionsProcedure.mergeStoreFiles` already uses in this package.

Would you prefer I switch entirely to the HFile-metadata source in this PR, or keep `ServerManager` primary with HFile-metadata as a fallback when the cache is empty? Happy to do either.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManagerUtil.java
nirdosh0110: You're right — `WALSplitUtil.getSplitEditFilesSorted` already restricts filenames via `EDITFILES_NAME_PATTERN` (`-?[0-9]+`), so `Long.parseLong` cannot throw here. Removed the guard in c7e4623 and left a short note pointing at the pattern.

virajjasani (approved):

virajjasani:
@nirdosh0110 could you also prepare branch-2 PR?

nirdosh0110:
> @nirdosh0110 could you also prepare branch-2 PR?

@virajjasani Raised a PR: https://github.com/apache/hbase/pull/8699 → target branch apache/branch-2.

PR #8604, opened 2026-09-02
HBASE-30355 setupMiniKdc port-conflict retry never triggers because Kerby throws KrbException, not BindException

mnpoonia:
### Problem

`TestLogLevel` fails intermittently in `setUp` when the MiniKDC port is already in use:

```
org.apache.kerby.kerberos.kerb.KrbException: Failed to start DefaultKrbServer. Address already in use (Bind failed)
    at ...AbstractInternalKdcServer.start
    at ...SimpleKdcServer.start
    at org.apache.hadoop.minikdc.MiniKdc.start(MiniKdc.java:284)
    at org.apache.hadoop.hbase.http.log.TestLogLevel.setupMiniKdc(TestLogLevel.java:155)
```

### Root cause

`setupMiniKdc` already retries on port conflicts, but only `catch (BindException e)`. The Kerby-backed `MiniKdc` does **not** throw `java.net.BindException` — it wraps the bind failure in a Kerby `KrbException`, and the cause chain carries **no** `BindException`; the conflict is only visible via the `"Address already in use"` message. So the retry is dead code and the first collision fails the test.

Same block exists in two places:
- `TestLogLevel.setupMiniKdc` (hbase-http)
- `HBaseTestingUtil.setupMiniKdc` (hbase-server) — also affects e.g. `TestSecureRESTServer`, `TestSecureExport`.

### Fix

Recognise a bind conflict regardless of wrapper type: `catch (Exception e)`, then a predicate that walks the whole cause chain and matches the `"Address already in use"` message; rethrow anything else so a genuine KDC misconfig is never masked.

Deliberately **not** `catch (BindException | KrbException)`:
- the message (not the type) is the discriminator — the cause chain has no `BindException`, and a future MiniKdc/Kerby could use a different wrapper;
- a bare `KrbException` catch would still need the message check (a real misconfig is also a `KrbException`);
- it avoids a compile-time import of a kerby type, which HBASE-29117 warns against (kerby version is unpinned; 1.x/2.x incompatible across Hadoop versions).

### Tests

- `testKdcBindConflictSurfacesAsKrbException` — deterministic reproduction: occupies the KDC port on TCP+UDP, pins MiniKdc to it, asserts the failure is a Kerby `KrbException` (not a `BindException`) recognised by the predicate.
- `testIsBindExceptionRecognizesKerbyWrappedBindFailure` — unit test for the predicate (wrapped `BindException`, message-only, and negative case).

### DIFF
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
@@ -3737,10 +3737,10 @@ public class HBaseTestingUtil extends HBaseZKTestingUtil {
         // KrbException whose shape varies by version (see isBindException), so we recognise the
         // port conflict via that predicate rather than a type. We also avoid importing kerby types
         // here (see HBASE-29117).
+        FileUtils.deleteDirectory(dir); // clean directory regardless of failure type
         if (!isBindException(e)) {
           throw e; // not a port conflict, do not mask the real failure behind a retry
         }
-        FileUtils.deleteDirectory(dir); // clean directory
         numTries++;
         if (numTries == 3) {
           LOG.error("Failed setting up MiniKDC. Tried " + numTries + " times.");
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
@@ -43,6 +43,7 @@ import java.util.Collections;
 import java.util.HashSet;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 import java.util.NavigableSet;
 import java.util.Properties;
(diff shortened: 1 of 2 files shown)


mnpoonia:
@virajjasani @Apache9 Please provide a approval to get a test run with github workflows.

mnpoonia:
@apurtell @virajjasani Can you please help in reviewing this.

hbase-server/src/test/java/org/apache/hadoop/hbase/HBaseTestingUtil.java
```diff
@@ -3731,21 +3732,51 @@ public MiniKdc setupMiniKdc(File keytabFile) throws Exception {
...
         kdc = new MiniKdc(conf, dir);
         kdc.start();
-      } catch (BindException e) {
+      } catch (Exception e) {
+        // Catch Exception, not BindException/KrbException: Kerby wraps the bind failure in a
+        // KrbException whose shape varies by version (see isBindException), so we recognise the
+        // port conflict via that predicate rather than a type. We also avoid importing kerby types
+        // here (see HBASE-29117).
+        if (!isBindException(e)) {
+          throw e; // not a port conflict, do not mask the real failure behind a retry
```
virajjasani, line 3742: Do we not need to delete dir?

mnpoonia: Good catch — fixed in 70eb936: moved `FileUtils.deleteDirectory(dir)` ahead of the `isBindException` check so the partially-initialized kdc dir is cleaned up before rethrowing on any failure, not just on a retried bind conflict.

virajjasani (approved):

PR #8608, opened 2026-09-02
HBASE-30346: Fix NPE in AbstractFSWAL WAL-roll debug logging aborting RegionServer

mnpoonia:
Backport of #8580 to branch-2.

FSHLog#getPipeline() delegated directly to Hadoop's DFSOutputStream#getPipeline(), which can legitimately return null (streamer closed, or no block pipeline currently established between blocks). The debug logging added in HBASE-28775 replaced the null-tolerant Arrays.toString(...) with Arrays.stream(...), which throws NPE on null, turning a benign transient HDFS state into a RegionServer self-abort during WAL rolling -- even though the roll itself had already succeeded.

Normalize FSHLog#getPipeline() to never return null (an empty array instead), matching the contract AsyncFSWAL#getPipeline() already honors.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
@@ -1208,13 +1208,21 @@ public class FSHLog extends AbstractFSWAL<Writer> {
   }
 
   /**
-   * This method gets the pipeline for the current WAL.
+   * This method gets the pipeline for the current WAL. Note that DFSOutputStream#getPipeline() can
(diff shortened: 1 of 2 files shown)


mnpoonia:
The previous CI run failed due to test flakiness unrelated to this change:

  - TestClientScannerTimeouts.testRetryOutOfOrderScannerNextExceptionAsync failed with Expected tryNumber=0, actual=1
  - Subsequent retries of TestClientScannerTimeouts errored with Master not initialized after 200000ms (mini-cluster startup timeout, likely CI resource contention)
  - TestAccessController and TestCanaryTool.testWebUI also flaked but passed on rerun

  None of these relate to the WAL-roll NPE fix in this patch.
@Apache9 FYI

PR #8611, opened 2026-09-02
HBASE-30341: Fix FSHLog WAL lockup on uncaught exception in publishSyncOnRingBuffer

sidkhillon:
`publishSyncOnRingBuffer` claims a Disruptor sequence via RingBuffer.next() and previously called getSyncFuture() before the try whose finally publishes the sequence. An exception from getSyncFuture (seen in production as an NPE from a corrupted Guava cache in SyncFutureCache) left the claimed slot unpublished, so the consumer could never advance and the whole WAL deadlocked.

- Move getSyncFuture inside the try so the sequence is always published.
- Make the RingBufferEventHandler tolerate the resulting empty truck by logging and falling through (matching master's AbstractFSWAL.consume) rather than failing outstanding syncs.
- Make SyncFutureCache.getIfPresentOrNew non-throwing: the cache is purely an allocation optimisation, so fall back to a new SyncFuture if it throws.

Adds a regression test for the WAL lockup and a test for the SyncFutureCache fallback. The FSHLog parts were incidentally fixed on master by HBASE-27231, never backported.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
@@ -794,8 +794,11 @@ public class FSHLog extends AbstractFSWAL<Writer> {
 
   protected SyncFuture publishSyncOnRingBuffer(long sequence, boolean forceSync) {
     // here we use ring buffer sequence as transaction id
-    SyncFuture syncFuture = getSyncFuture(sequence, forceSync);
+    // getSyncFuture must stay inside the try: the sequence is already claimed, so we must publish
+    // it even if this throws, else the consumer wedges.
+    SyncFuture syncFuture = null;
     try {
+      syncFuture = getSyncFuture(sequence, forceSync);
       RingBufferTruck truck = this.disruptor.getRingBuffer().get(sequence);
       truck.load(syncFuture);
     } finally {
@@ -1080,11 +1083,9 @@ public class FSHLog extends AbstractFSWAL<Writer> {
             entry.release();
           }
         } else {
-          // What is this if not an append or sync. Fail all up to this!!!
-          cleanupOutstandingSyncsOnException(sequence,
-            new IllegalStateException("Neither append nor sync"));
-          // Return to keep processing.
-          return;
+          // Empty truck: publishSyncOnRingBuffer claimed the sequence but threw before loading it.
+          // Fall through rather than failing syncs, so the empty slot is harmless.
+          LOG.warn("RingBufferTruck with unexpected type: {}", truck.type());
         }
 
         // TODO: Check size and if big go ahead and call a sync if we have enough data.
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
@@ -1082,10 +1082,14 @@ public class FSHLog extends AbstractFSWAL<Writer> {
           } finally {
             entry.release();
           }
+        } else if (truck.type() == RingBufferTruck.Type.EMPTY) {
+          // publishSyncOnRingBuffer claimed the sequence but threw before loading the truck.
+          LOG.warn("Empty RingBufferTruck at sequence {}", sequence);
+          return;
         } else {
-          // Empty truck: publishSyncOnRingBuffer claimed the sequence but threw before loading it.
-          // Fall through rather than failing syncs, so the empty slot is harmless.
-          LOG.warn("RingBufferTruck with unexpected type: {}", truck.type());
+          cleanupOutstandingSyncsOnException(sequence,
+            new IllegalStateException("Unexpected truck type: " + truck.type()));
+          return;
         }
 
         // TODO: Check size and if big go ahead and call a sync if we have enough data.
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/SyncFutureCache.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/SyncFutureCache.java
@@ -21,6 +21,8 @@ import java.util.concurrent.TimeUnit;
 import org.apache.hadoop.conf.Configuration;
 import org.apache.hadoop.hbase.HConstants;
 import org.apache.yetus.audience.InterfaceAudience;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.apache.hbase.thirdparty.com.google.common.cache.Cache;
 import org.apache.hbase.thirdparty.com.google.common.cache.CacheBuilder;
@@ -45,6 +47,8 @@ import org.apache.hbase.thirdparty.com.google.common.cache.CacheBuilder;
 @InterfaceAudience.Private
 public final class SyncFutureCache {
 
+  private static final Logger LOG = LoggerFactory.getLogger(SyncFutureCache.class);
+
   private static final long SYNC_FUTURE_INVALIDATION_TIMEOUT_MINS = 2;
 
   private final Cache<Thread, SyncFuture> syncFutureCache;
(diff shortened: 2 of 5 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
```diff
@@ -1080,11 +1083,9 @@ public void onEvent(final RingBufferTruck truck, final long sequence, boolean en
             entry.release();
           }
         } else {
-          // What is this if not an append or sync. Fail all up to this!!!
-          cleanupOutstandingSyncsOnException(sequence,
-            new IllegalStateException("Neither append nor sync"));
-          // Return to keep processing.
-          return;
```
hgromer, line 1087: Should we keep the `return` here? I'm pretty sure falling through isn't problematic, but it might be cleaner/easier to reason about if we simply returned here.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
```diff
@@ -1080,11 +1083,9 @@ public void onEvent(final RingBufferTruck truck, final long sequence, boolean en
...
+          // Empty truck: publishSyncOnRingBuffer claimed the sequence but threw before loading it.
+          // Fall through rather than failing syncs, so the empty slot is harmless.
+          LOG.warn("RingBufferTruck with unexpected type: {}", truck.type());
```
hgromer: Would it be better to have an explicit case for this? Maybe something like

```java
} else if (truck.type() == RingBufferTruck.Type.EMPTY) {
    LOG.warn("Empty RingBufferTruck at sequence {}", sequence);
    return;
} else {
    cleanupOutstandingSyncsOnException(sequence,
        new IllegalStateException("Unexpected truck type: " + truck.type()));
    return;
}
```

That way we still cleanup on truly exceptional cases?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
sidkhillon: Agreed, it should be the same but more readable if we return

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
sidkhillon: Sure! Just took your suggestion

hgromer:
This looks good to me, I've added Chenglei and Duo as reviewers in case they'd like to comment.

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
```diff
@@ -794,8 +794,11 @@ private SyncFuture publishSyncOnRingBuffer(boolean forceSync) {
 
   protected SyncFuture publishSyncOnRingBuffer(long sequence, boolean forceSync) {
     // here we use ring buffer sequence as transaction id
-    SyncFuture syncFuture = getSyncFuture(sequence, forceSync);
+    // getSyncFuture must stay inside the try: the sequence is already claimed, so we must publish
+    // it even if this throws, else the consumer wedges.
+    SyncFuture syncFuture = null;
     try {
+      syncFuture = getSyncFuture(sequence, forceSync);
```
Apache9, line 801: What exception will getSyncFuture throw? Seems the method only has memory operations...

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/SyncFutureCache.java
```diff
@@ -57,9 +61,15 @@ public SyncFutureCache(final Configuration conf) {
...
   public SyncFuture getIfPresentOrNew() {
-    // Invalidate the entry if a mapping exists. We do not want it to be reused at the same time.
-    SyncFuture future = syncFutureCache.asMap().remove(Thread.currentThread());
-    return (future == null) ? new SyncFuture() : future;
+    // The cache is only an allocation optimisation; never let it fail a write.
+    try {
+      // Invalidate the entry if a mapping exists. We do not want it to be reused at the same time.
+      SyncFuture future = syncFutureCache.asMap().remove(Thread.currentThread());
+      return (future == null) ? new SyncFuture() : future;
+    } catch (RuntimeException e) {
```
Apache9, line 69: Where does the RuntimeException come from?

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/FSHLog.java
sidkhillon: It's all in-memory, but the memory it mutates is Guava's `LocalCache`, and its write-order queue (present because `SyncFutureCache` uses `expireAfterWrite`) can NPE under a race. We hit it in production on branch-2.6 (`hbase.wal.provider=filesystem`):

```
java.lang.NullPointerException: Cannot invoke "org.apache.hbase.thirdparty.com.google.common.cache.ReferenceEntry.setNextInWriteQueue(org.apache.hbase.thirdparty.com.google.common.cache.ReferenceEntry)" because "previous" is null
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache.connectWriteOrder(LocalCache.java:1818)
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache$WriteQueue.remove(LocalCache.java:3725)
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache$Segment.removeValueFromChain(LocalCache.java:3249)
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache$Segment.remove(LocalCache.java:3079)
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache.remove(LocalCache.java:4273)
    at org.apache.hadoop.hbase.regionserver.wal.SyncFutureCache.getIfPresentOrNew(SyncFutureCache.java:61)
    at org.apache.hadoop.hbase.regionserver.wal.AbstractFSWAL.getSyncFuture(AbstractFSWAL.java:1093)
    at org.apache.hadoop.hbase.regionserver.wal.FSHLog.publishSyncOnRingBuffer(FSHLog.java:789)
    at org.apache.hadoop.hbase.regionserver.wal.FSHLog.publishSyncOnRingBuffer(FSHLog.java:784)
    at org.apache.hadoop.hbase.regionserver.wal.FSHLog.publishSyncThenBlockOnCompletion(FSHLog.java:801)
    at org.apache.hadoop.hbase.regionserver.wal.FSHLog.doSync(FSHLog.java:836)
    at org.apache.hadoop.hbase.regionserver.wal.AbstractFSWAL.sync(AbstractFSWAL.java:605)
    at org.apache.hadoop.hbase.regionserver.HRegion.doWALAppend(HRegion.java:7956)
    at org.apache.hadoop.hbase.regionserver.HRegion.batchMutate(HRegion.java:4597)
    ... RSRpcServices.multi -> RpcServer.call -> CallRunner.run -> RpcHandler.run
```

hbase-server/src/main/java/org/apache/hadoop/hbase/regionserver/wal/SyncFutureCache.java
sidkhillon: There can be the same NPE from Guava's `LocalCache` write-queue

```java
java.lang.NullPointerException: Cannot invoke "...ReferenceEntry.setNextInWriteQueue(...)" because "previous" is null
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache.connectWriteOrder(LocalCache.java:1818)
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache$WriteQueue.remove(LocalCache.java:3725)
    at org.apache.hbase.thirdparty.com.google.common.cache.LocalCache.remove(LocalCache.java:4273)
    at org.apache.hadoop.hbase.regionserver.wal.SyncFutureCache.getIfPresentOrNew(SyncFutureCache.java:61)
```

I'm happy to reduce this to just a NPE catch if you'd like. I kept it broad so any exception falls back to a new SyncFuture()

PR #8622, opened 2026-09-05
HBASE-30357 OpenRegionProcedure#restoreSucceedState ignores persisted transitionCode, forcing OPEN even after a real FAILED_OPEN

mnpoonia:

`OpenRegionProcedure#restoreSucceedState()` is called on master-failover restore, once
per region, via `RegionRemoteProcedureBase#stateLoaded()`. It unconditionally forced the
region into `OPEN`, regardless of the actual persisted `transitionCode`, which can be
`FAILED_OPEN`. The method's signature only received `seqId`, not `transitionCode`, so it
was structurally unable to branch on the real outcome.

Concretely: an RS reports `FAILED_OPEN`; the master persists
`state=REPORT_SUCCEED, transitionCode=FAILED_OPEN` on the `OpenRegionProcedure`, but the
in-memory `RegionState.State` is left untouched (still `OPENING`, since
`AssignmentManager#regionFailedOpen(regionNode, false)` on the live path does not update
`RegionState.State`). If the master crashes/restarts before this reaches `confirmOpened()`
and before `hbase:meta` is updated, `restoreSucceedState()` sees the region isn't `OPEN`
yet and force-transitions it to `OPEN` anyway, then persists that (false) `OPEN` state to
`hbase:meta`. There is no rollback anywhere in this procedure chain (by design -
forward-only), so nothing downstream can detect or correct it; the region silently looks
healthy in meta while no RegionServer is actually serving it.

By contrast, `CloseRegionProcedure#restoreSucceedState()` is safe doing the equivalent
unconditional force, because CLOSE has no failure-variant transition code at the master
side (`checkTransition`/`updateTransitionWithoutPersistingToMeta` both assert
`transitionCode == CLOSED`).

This is not a regression - the logic is unchanged (modulo spotless formatting) since it
was introduced in HBASE-22365 (2019-05-10).

### The fix

- Widen `RegionRemoteProcedureBase#restoreSucceedState()` to also receive the persisted
  `transitionCode`, passed through from `stateLoaded()`.
- `OpenRegionProcedure#restoreSucceedState()` now branches: `FAILED_OPEN` calls
  `AssignmentManager#regionFailedOpen(regionNode, false)`, mirroring exactly what the live
  `reportTransition`/`updateTransitionWithoutPersistingToMeta` path already does for the
  same transition code; the existing `OPENED` handling is unchanged.
- `CloseRegionProcedure#restoreSucceedState()` accepts the new parameter and ignores it,
  since CLOSE has no failure variant.

After the fix, a restore-time `FAILED_OPEN` puts the region back into the same
retryable path (`OPENING`, cleared location) that `TransitRegionStateProcedure#confirmOpened()`
already drives on the live path - the region gets reassigned/retried through the normal
flow instead of being falsely marked `OPEN`.

To prevent a region from being silently, durably marked `OPEN` in `hbase:meta` after a
master restart, when in reality no RegionServer opened it. Since HBase has no rollback
mechanism for these forward-only assignment procedures, this bug is otherwise
unrecoverable except by manual detection and intervention.

## Does this PR introduce any user-facing change?

No.

## Is there a corresponding Apache JIRA?

Yes: [HBASE-30357](https://issues.apache.org/jira/browse/HBASE-30357)

## How was this patch tested?

Added `TestOpenRegionProcedureRestoreFailedOpen`, extending `TestAssignmentManagerBase`. A
custom mock RS executor reports `FAILED_OPEN` on the first open attempt, then - while
holding the `RegionStateNode` lock, simulating the point right after a crash where the
child procedure has not yet resumed its own `execute()` - directly invokes the
`TransitRegionStateProcedure#stateLoaded()` hook that a real master restart would trigger,
and records the region's state immediately after. Confirmed the test fails against
unmodified code with the exact predicted `OPEN` state, and passes after the fix. Also ran
`TestAssignmentManager`, `TestTransitRegionStateProcedure`, `TestOpenRegionProcedureHang`,
`TestOpenRegionProcedureBackoff`, `TestRollbackSCP`, and `TestSCPGetRegionsRace` with no
regressions.

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/OpenRegionProcedure.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/OpenRegionProcedure.java
@@ -127,7 +127,12 @@ public class OpenRegionProcedure extends RegionRemoteProcedureBase {
 
   @Override
   protected void restoreSucceedState(AssignmentManager am, RegionStateNode regionNode,
-    long openSeqNum) throws IOException {
+    TransitionCode transitionCode, long openSeqNum) throws IOException {
+    if (transitionCode == TransitionCode.FAILED_OPEN) {
+      // will not persist to meta if giveUp is false, matches the live reportTransition path
+      am.regionFailedOpen(regionNode, false);
+      return;
+    }
     if (regionNode.getState() == State.OPEN) {
       // should have already been persisted, ignore
       return;
--- a/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
+++ b/hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
@@ -1,0 +1,95 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.hadoop.hbase.master.assignment;
+
+import static org.junit.jupiter.api.Assertions.assertNotEquals;
+
+import java.io.IOException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
+import org.apache.hadoop.hbase.HConstants;
+import org.apache.hadoop.hbase.ServerName;
+import org.apache.hadoop.hbase.TableName;
+import org.apache.hadoop.hbase.client.RegionInfo;
+import org.apache.hadoop.hbase.master.RegionState;
+import org.apache.hadoop.hbase.testclassification.LargeTests;
+import org.apache.hadoop.hbase.testclassification.MasterTests;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import org.apache.hadoop.hbase.shaded.protobuf.ProtobufUtil;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.AdminProtos.OpenRegionRequest.RegionOpenInfo;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.AdminProtos.OpenRegionResponse.RegionOpeningState;
+import org.apache.hadoop.hbase.shaded.protobuf.generated.RegionServerStatusProtos.RegionStateTransition.TransitionCode;
+
+/**
+ * HBASE-30357: OpenRegionProcedure#restoreSucceedState() must not force the region state to OPEN on
+ * master-failover restore when the persisted transition code is actually FAILED_OPEN.
+ */
+@Tag(MasterTests.TAG)
+@Tag(LargeTests.TAG)
+public class TestOpenRegionProcedureRestoreFailedOpen extends TestAssignmentManagerBase {
+
+  /**
+   * On the first open attempt, reports FAILED_OPEN and then immediately simulates a master restart
+   * happening before the child OpenRegionProcedure gets to run its own execute() (i.e. before it
+   * can persist anything to meta), by directly invoking the package-private stateLoaded() hook that
+   * a real restart would trigger. Records the region state right after that simulated restore.
+   * Subsequent attempts behave like a normal healthy RS so the assign procedure can converge and
+   * the test does not hang.
+   */
+  private class FailedOpenThenRestoreRsExecutor extends GoodRsExecutor {
+
+    private final AtomicBoolean firstAttempt = new AtomicBoolean(true);
+
+    final AtomicReference<RegionState.State> stateAfterRestore = new AtomicReference<>();
+
+    @Override
+    protected RegionOpeningState execOpenRegion(ServerName server, RegionOpenInfo openReq)
+      throws IOException {
+      if (!firstAttempt.compareAndSet(true, false)) {
+        return super.execOpenRegion(server, openReq);
+      }
+      RegionInfo hri = ProtobufUtil.toRegionInfo(openReq.getRegion());
+      RegionStateNode regionNode = am.getRegionStates().getRegionStateNode(hri);
+      TransitRegionStateProcedure trsp = (TransitRegionStateProcedure) regionNode.getProcedure();
+      regionNode.lock();
+      try {
+        sendTransitionReport(server, openReq.getRegion(), TransitionCode.FAILED_OPEN,
+          HConstants.NO_SEQNUM);
+        trsp.stateLoaded(am, regionNode);
+        stateAfterRestore.set(regionNode.getState());
+      } finally {
+        regionNode.unlock();
+      }
+      return RegionOpeningState.FAILED_OPENING;
+    }
+  }
+
+  @Test
+  public void testRestoreDoesNotForceOpenAfterFailedOpen() throws Exception {
+    TableName tableName = TableName.valueOf(testMethodName);
+    RegionInfo hri = createRegionInfo(tableName, 1);
+    FailedOpenThenRestoreRsExecutor executor = new FailedOpenThenRestoreRsExecutor();
+    rsDispatcher.setMockRsExecutor(executor);
+    TransitRegionStateProcedure proc = createAssignProcedure(hri);
+    waitOnFuture(submitProcedure(proc));
+
+    assertNotEquals(RegionState.State.OPEN, executor.stateAfterRestore.get());
+  }
+}
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/AssignmentManager.java
@@ -2232,8 +2232,13 @@ public class AssignmentManager {
   }
 
   // should be called under the RegionStateNode lock
-  // The parameter 'giveUp' means whether we will try to open the region again, if it is true, then
-  // we will persist the FAILED_OPEN state into hbase:meta.
+  // Called when a region open attempt has failed. The 'giveUp' parameter says whether we will try
+  // to open the region again. If true, this is a terminal failure with no more retries: we set
+  // RegionState.State to FAILED_OPEN and persist it to hbase:meta, and only remove the region from
+  // its old target server's bookkeeping once that persist succeeds. If false, we are about to
+  // retry the open with a new plan, so RegionState.State is intentionally left untouched (e.g.
+  // still OPENING), meta is not touched, and we remove the region from its old target server's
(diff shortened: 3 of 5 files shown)


mnpoonia:
@virajjasani @apurtell @Apache9 Can you please help with review.

mnpoonia:
@Apache9 can you please help in reviewing.

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/OpenRegionProcedure.java
```diff
@@ -127,7 +127,12 @@ protected void updateTransitionWithoutPersistingToMeta(MasterProcedureEnv env,
 
   @Override
   protected void restoreSucceedState(AssignmentManager am, RegionStateNode regionNode,
-    long openSeqNum) throws IOException {
+    TransitionCode transitionCode, long openSeqNum) throws IOException {
+    if (transitionCode == TransitionCode.FAILED_OPEN) {
+      // will not persist to meta if giveUp is false, matches the live reportTransition path
```
virajjasani, line 132: Is `FAILED_OPEN` the only case where we have this problem? What about split/merge reverted states?

virajjasani (commented):
Nice one, this is worth fixing

hbase-server/src/main/java/org/apache/hadoop/hbase/master/assignment/OpenRegionProcedure.java
mnpoonia: Good question — I checked both cases:

**`CloseRegionProcedure`**: no analogous bug possible. `CloseRegionProcedure.checkTransition` only ever accepts `CLOSED` (`CloseRegionProcedure.java:99-106`) — there's no `FAILED_CLOSE`/revert code in the protocol at all (the RS aborts instead of reporting a close failure). So `transitionCode` is structurally guaranteed to be `CLOSED` whenever it's persisted for this procedure, and `restoreSucceedState` can't mis-restore something that only ever has one valid value.

**Split/merge**: these don't go through `RegionRemoteProcedureBase`/`OpenRegionProcedure`/`CloseRegionProcedure`/`restoreSucceedState` at all — they're driven by `SplitTableRegionProcedure`/`MergeTableRegionsProcedure` and reported via a separate switch in `AssignmentManager.reportRegionStateTransition`, which routes `READY_TO_SPLIT`/`READY_TO_MERGE` to their own handlers. Those handlers reject `SPLIT`/`MERGED`/`SPLIT_REVERTED`/`MERGE_REVERTED` outright — those codes predate AMv2 (from the 1.x ZK-less-assignment era, HBASE-11059) and exist only as a rolling-upgrade compatibility guard against an old (<2.0) RS still reporting the legacy codes. No current RS code ever sends them.

So `FAILED_OPEN` is the only live instance of this restore-path gap — nothing further to fix here.

virajjasani (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
```diff
@@ -0,0 +1,95 @@
...
+/**
+ * HBASE-30357: OpenRegionProcedure#restoreSucceedState() must not force the region state to OPEN on
+ * master-failover restore when the persisted transition code is actually FAILED_OPEN.
+ */
+@Tag(MasterTests.TAG)
+@Tag(LargeTests.TAG)
+public class TestOpenRegionProcedureRestoreFailedOpen extends TestAssignmentManagerBase {
+
+  /**
+   * On the first open attempt, reports FAILED_OPEN and then immediately simulates a master restart
```
Apache9, line 49: So how do we simulate the master restart here? I haven't seen any restart in the test.

mnpoonia: Good question. We don’t actually restart the master in this test.

The test simulates the relevant part of a master restart by directly invoking TransitRegionStateProcedure#stateLoaded(), which is the hook reached when the procedure state is loaded during master recovery.

The important sequence we want to reproduce is:

1. RS reports FAILED_OPEN.
2. The procedure state containing REPORT_SUCCEED + FAILED_OPEN is persisted.
3. Master crashes before the child procedure gets a chance to continue and update the region state/meta.
4. On recovery, stateLoaded() restores the procedure and calls restoreSucceedState().

The test holds the RegionStateNode lock after reporting FAILED_OPEN and invokes stateLoaded() directly, so the child procedure cannot continue normally. This puts us at the same point in the procedure lifecycle as the restore path after a master restart.

I can make the test/comment more explicit about this, e.g. rename the comment from “simulates a master restart” to “simulates procedure recovery after master restart”, to avoid suggesting that the test actually restarts the master.

Apache9: I prefer we at least restart the ProcedureExecutor? There is a util for doing this.

mnpoonia: - Our test extends TestAssignmentManagerBase, which uses MockMasterServices. That class hardcodes new
    NoopProcedureStore() — nothing is actually persisted, so a genuine executor restart would just lose
    all procedure/region state. A real restart-and-reload only works with a real WALProcedureStore +
    real hbase:meta, i.e. a real mini-cluster (HBaseTestingUtil.startMiniCluster), the pattern
    TestReportRegionStateTransitionRetry/TestOpenRegionProcedureHang use.

I will rewrite the test in same way as TestRollbackSCP, TestSplitTableRegionProcedure

mnpoonia: @Apache9 I have now updated the code to use a mini cluster and a ProcedureExecutor restart as other test do.
Please review

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
```diff
@@ -0,0 +1,190 @@
...
+    PROCEED.set(proceed);
+    Future<byte[]> future = am.moveAsync(
+      new RegionPlan(region, regionNode.getRegionLocation(), regionNode.getRegionLocation()));
+    TransitRegionStateProcedure proc =
+      procExec.getProcedures().stream().filter(p -> p instanceof TransitRegionStateProcedure)
+        .filter(p -> !p.isFinished()).map(p -> (TransitRegionStateProcedure) p).findAny().get();
+    // wait until we are suspended waiting on the RS to report the open, then let it report; the
+    // AssignmentManagerForTest above rewrites that report to FAILED_OPEN and blocks the RPC handler
+    // thread, still holding the RegionStateNode lock, until we tell it to proceed
+    UTIL.waitFor(30000,
```
Apache9: Do we really need this waitFor? The below 'arrive.await' is not enough?

Apache9 (approved):
Anyway, no problems from me now. Just a minor question about the UT.

Thanks @mnpoonia !

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
mnpoonia: AGree. it can be removed.
infact even the above line is from last test. The below can be removed

`
TransitRegionStateProcedure proc =
      procExec.getProcedures().stream().filter(p -> p instanceof TransitRegionStateProcedure).  
        .filter(p -> !p.isFinished()).map(p -> (TransitRegionStateProcedure) p).findAny().get();  
    // wait until we are suspended waiting on the RS to report the open, then let it report; the.  
    // AssignmentManagerForTest above rewrites that report to FAILED_OPEN and blocks the RPC handler.   
    // thread, still holding the RegionStateNode lock, until we tell it to proceed.  
    UTIL.waitFor(30000,  
() -> proc.getCurrentStateId() == REGION_STATE_TRANSITION_CONFIRM_OPENED_VALUE);  
`


@Apache9 I can do it right away if needed. this is just dead code

mnpoonia:
>Anyway, no problems from me now. Just a minor question about the UT.

Removed the dead code from the test. Thank you @Apache9 for the review

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
```diff
@@ -0,0 +1,183 @@
...
+    PROCEED.set(proceed);
+    Future<byte[]> future = am.moveAsync(
+      new RegionPlan(region, regionNode.getRegionLocation(), regionNode.getRegionLocation()));
+    // arrive counts down only after the RS's OPENED report is intercepted, rewritten to
+    // FAILED_OPEN, and persisted, so waiting on it alone is sufficient synchronization
+    assertTrue(arrive.await(AWAIT_TIMEOUT_SECONDS, TimeUnit.SECONDS));
+
+    MasterProcedureTestingUtility.restartMasterProcedureExecutor(procExec);
+    RegionStateNode reloaded = am.getRegionStates().getRegionStateNode(region);
+    assertNotEquals(RegionState.State.OPEN, reloaded.getState());
```
Apache9: What do we expected here? FAILED_OPEN? Checking the above code, I think here we will get CLOSED? Since when restarting the procedure executor, we have not returned the FAILED_OPEN state yet?

Apache9 (approved):

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
mnpoonia: Good catch — actual result is `OPENING`, not `CLOSED`. `assertNotEquals(OPEN, ...)` was written as a minimal regression guard against the exact symptom of this bug (`restoreSucceedState()` forcing the region to `OPEN`), rather than pinning the specific correct value — that's how a wrong fix landing on e.g. `CLOSED` would still have passed it.

Restarting the ProcedureExecutor re-triggers `AssignmentManager.joinCluster()`'s meta scan, which reloads `regionNode`'s state from the persisted `hbase:meta` column (`OPENING`, since meta was never advanced past the open-dispatch point) before `stateLoaded()` runs. `restoreSucceedState(FAILED_OPEN)` then calls `regionFailedOpen(node, false)`, which only detaches from the dead server and never calls `setState()` — so it stays `OPENING`.

hbase-server/src/test/java/org/apache/hadoop/hbase/master/assignment/TestOpenRegionProcedureRestoreFailedOpen.java
```diff
@@ -0,0 +1,190 @@
...
+      // AssignmentManager#updateRegionTransition() (called from super.reportRegionStateTransition
+      // below) also locks this same RegionStateNode; that only works here because the lock is
+      // reentrant for the same thread (see RegionStateNodeLock#lock0).
+      regionNode.lock();
+      try {
+        // persists REPORT_SUCCEED/FAILED_OPEN to the real WALProcedureStore and wakes the child
+        // OpenRegionProcedure, but since we still hold the RegionStateNode lock here (reentrant,
+        // same thread), the woken child can not resume and complete its own meta update - this is
+        // exactly the window a real master crash would leave us in.
+        ReportRegionStateTransitionResponse resp = super.reportRegionStateTransition(failedOpenReq);
```
Apache9, line 110: OK, here we have finished the real report.

Apache9 (approved):

PR #8623, opened 2026-09-06
HBASE-29620 Refactor TestRSGroupsKillRS

Sigma-Ma:

Split the mixed-version scenario into `TestRSGroupsKillRSWithDifferentVersions`. Use a test `HMaster` implementation to report a lower version for the target RegionServer, and restore `VersionInfo#getVersion` to return the generated version directly.

The existing test changes process-wide `VersionInfo` state through reflection. Since MiniCluster components run in the same JVM, this may affect unrelated servers and requires production code to keep mutable state only for testing.

## How was this patch tested?

```bash
mvn -pl hbase-server -Dtest=TestRSGroupsKillRS,TestRSGroupsKillRSWithDifferentVersions -Dsurefire.failIfNoSpecifiedTests=false -Dsurefire.rerunFailingTestsCount=0 test

mvn -pl hbase-common -Dtest=TestVersionInfo -Dsurefire.rerunFailingTestsCount=0 test

### DIFF
--- a/hbase-common/src/main/java/org/apache/hadoop/hbase/util/VersionInfo.java
+++ b/hbase-common/src/main/java/org/apache/hadoop/hbase/util/VersionInfo.java
@@ -40,15 +40,12 @@ public class VersionInfo {
   // higher than any numbers in the version.
   private static final int VERY_LARGE_NUMBER = 100000;
 
-  // Copying into a non-final member so that it can be changed by reflection for testing
-  private static String version = Version.version;
-
   /**
    * Get the hbase version.
    * @return the hbase version string, eg. "0.6.3-dev"
    */
   public static String getVersion() {
-    return version;
+    return Version.version;
   }
 
   /**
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: hbase-common/src/main/java/org/apache/hadoop/hbase/util/VersionInfo.java [hunk 1]
```diff
@@ -40,15 +40,12 @@ public class VersionInfo {
...
 
-  // Copying into a non-final member so that it can be changed by reflection for testing
-  private static String version = Version.version;
-
   /**
    * Get the hbase version.
    * @return the hbase version string, eg. "0.6.3-dev"
    */
   public static String getVersion() {
-    return version;
```
Apache9, line 51: Why this change?

Sigma-Ma: This mutable field was introduced only to let the test override the version through reflection. Since the test now simulates the lower version through HMaster, the field is no longer needed, so getVersion() can return Version.version directly again.

Apache9: This is the only place where we override this field by reflection?

Sigma-Ma: Yes. I checked the repository, and TestRSGroupsKillRS#setVersionInfoVersion is the only place that modifies VersionInfo.version through reflection.

PR #8625, opened 2026-09-07
HBASE-30361 Align test tooling and documentation with JUnit 5

liuxiaocs7:
- see: HBASE-30361
- Co-authored-by: OpenAI Codex (GPT-5.6)

### DIFF
--- a/hbase-common/src/test/java/org/apache/hadoop/hbase/logging/TestJul2Slf4j.java
+++ b/hbase-common/src/test/java/org/apache/hadoop/hbase/logging/TestJul2Slf4j.java
@@ -37,8 +37,8 @@ import org.mockito.invocation.InvocationOnMock;
 import org.mockito.stubbing.Answer;
 
 /**
- * This should be in the hbase-logging module but the {@link HBaseClassTestRule} is in hbase-common
- * so we can only put the class in hbase-common module for now...
+ * This should be in the hbase-logging module but hbase-logging cannot depend on hbase-common, so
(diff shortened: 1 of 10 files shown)


hbase-common/src/test/java/org/apache/hadoop/hbase/logging/TestJul2Slf4j.java
```diff
@@ -37,8 +37,8 @@
 import org.mockito.stubbing.Answer;
 
 /**
- * This should be in the hbase-logging module but the {@link HBaseClassTestRule} is in hbase-common
```
Apache9, line 40: Since we have removed HBaseClassTestRule, is it possible to move this to hbase-logging?

liuxiaocs7: Hi, @Apache9, thanks for the comments. I had planned to address this in a follow-up issue, since retaining `HBaseJupiterExtension` would introduce a circular dependency: `hbase-common` already depends on `hbase-logging`. Would you prefer moving the tests here without the extension, or addressing this together in a follow-up?

liuxiaocs7:
related: https://github.com/apache/hbase/pull/8690

kgeisz (approved):
LGTM. I'm not sure that the change in PR #8690 is necessary since I was able to get an integration test to run with maven after removing the `surefire-junit4` dependency in `hbase-it/pom.xml` like you did in this PR

PR #8626, opened 2026-09-08
HBASE-30365 Skip the MOB pseudo-region when building snapshot input splits

junegunn:
Jira: HBASE-30365

`TableSnapshotInputFormatImpl.getRegionInfosFromManifest` creates a split for every region in the snapshot manifest, and a MOB snapshot manifest includes the MOB pseudo-region from `MobUtils.getMobRegionInfo`. Each split is then opened as a real region under the restore dir.

HBASE-30336 correctly stopped writing a `.regioninfo` for that pseudo-region under the table dir, so opening it now fails with "Region directory does not exist".

Skip it, next to the existing offline-split filter. Behaviour preserving: the pseudo-region's family files live under mobdir, so the split opened a region with no store files and returned zero rows. MOB values are read through references in the data regions, not by scanning the pseudo-region.

Covered by the existing `TestCopyTable` MOB snapshot tests, which fail without this change.

```sh
mvn -pl hbase-mapreduce test -Dtest='TestCopyTable#testLoadingSnapshotAndBulkLoadToMobTable+tsetLoadingSnapshotToMobTable'
```

### DIFF
--- a/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/TableSnapshotInputFormatImpl.java
+++ b/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/TableSnapshotInputFormatImpl.java
@@ -43,6 +43,7 @@ import org.apache.hadoop.hbase.client.Scan;
 import org.apache.hadoop.hbase.client.Scan.ReadType;
 import org.apache.hadoop.hbase.client.TableDescriptor;
 import org.apache.hadoop.hbase.io.ImmutableBytesWritable;
+import org.apache.hadoop.hbase.mob.MobUtils;
 import org.apache.hadoop.hbase.regionserver.HRegion;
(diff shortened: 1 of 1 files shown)


liuxiaocs7 (approved):
Thanks for the fix!

junegunn:
Thanks @liuxiaocs7 for the review!

liuxiaocs7:
Hi, @junegunn, do we need to backport HBASE-30365 for branch-2.5?

junegunn:
I skipped branch 2.5 because HBASE-30336 was not in it. Let me know if I'm mistaken.

liuxiaocs7:
Hi, @junegunn, when I backport HBASE-30368 for branch-2.x, I find that `TestCopyTable.testLoadingSnapshotAndBulkLoadToMobTable` and `TestCopyTable.tsetLoadingSnapshotToMobTable` seems fails on branch-2.5 now, it not introduced by HBASE-30336, let me file a new jira issue to track this.

Thanks!

PR #8628, opened 2026-09-08
HBASE-30368 TableSnapshotScanner fails on an already restored MOB tab…

liuxiaocs7:
…le snapshot

see: HBASE-30368

### DIFF
--- a/hbase-server/src/main/java/org/apache/hadoop/hbase/client/TableSnapshotScanner.java
+++ b/hbase-server/src/main/java/org/apache/hadoop/hbase/client/TableSnapshotScanner.java
@@ -26,6 +26,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.hbase.PrivateCellUtil;
 import org.apache.hadoop.hbase.client.metrics.ScanMetrics;
+import org.apache.hadoop.hbase.mob.MobUtils;
 import org.apache.hadoop.hbase.regionserver.MemStoreLAB;
 import org.apache.hadoop.hbase.snapshot.RestoreSnapshotHelper;
 import org.apache.hadoop.hbase.snapshot.SnapshotDescriptionUtils;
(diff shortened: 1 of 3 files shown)


hbase-server/src/main/java/org/apache/hadoop/hbase/client/TableSnapshotScanner.java
```diff
@@ -160,6 +161,11 @@ private boolean isValidRegion(RegionInfo hri) {
     if (hri.isOffline() && (hri.isSplit() || hri.isSplitParent())) {
       return false;
     }
+    // The mob region is a dummy region used only to organise mob files under mobdir. It has no
+    // region directory under the table dir to open, and holds no rows. See HBASE-30365.
```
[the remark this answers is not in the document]
liuxiaocs7: Updated the comment to reference both HBASE-30365 and HBASE-30368. Thanks.

hbase-server/src/main/java/org/apache/hadoop/hbase/client/TableSnapshotScanner.java
```diff
@@ -160,6 +161,12 @@ private boolean isValidRegion(RegionInfo hri) {
     if (hri.isOffline() && (hri.isSplit() || hri.isSplitParent())) {
       return false;
     }
+    // The mob region is a dummy region used only to organise mob files under mobdir. It has no
+    // region directory under the table dir to open, and holds no rows. See HBASE-30365 and
+    // HBASE-30368.
```
junegunn: Thank you. The fix looks right to me.

I verified it locally: the test fails without the `isValidRegion` change, and a full scan of a restored mob snapshot still returns the real values, so skipping the pseudo-region does not lose data.

I have one request though. I realized the last clause of the comment ("holds no rows") is not accurate.

`DefaultMobStoreFlusher` appends the original cell to the mob file, then writes a reference cell to the normal store:

```java
mobFileWriter.append(c);
...
// The key is same, the value is the filename of the mob file
ExtendedCell reference = MobUtils.createMobRefCell(c, fileName, ...);
writer.append(reference);
```

So the mob files do hold every row, with the real values. They just must not be scanned as a region, since they are reached by reference from the real regions. Simplest fix is to drop that clause:


```suggestion
    // The mob region is a dummy region used only to organise mob files under mobdir. It has no
    // region directory under the table dir to open. See HBASE-30365 and HBASE-30368.
```

That wording is mine from HBASE-30365, so the same clause is already on master in `TableSnapshotInputFormatImpl`. Could you fix that one in this PR too?

liuxiaocs7: Hi, @junegunn, thanks for catching this and for verifying the full scan. You're right—the “holds no rows” wording is inaccurate. I've removed it from both `TableSnapshotScanner` and `TableSnapshotInputFormatImpl`.

junegunn (approved):
Thanks for the update!

PR #8632, opened 2026-09-09
HBASE-30373 Expose GITHUB_REPO to external environment

ndimiduk:
Permit GITHUB_REPO variable override from execution environment. Support Github Action runner environment variable by default. Fallback to original hard-coded value.


Apache9 (approved):

apurtell (commented):
+1

PR #8638, opened 2026-09-09
HBASE-30371: Point-In-Time Restore is trying to find .backup.manifest file in the .tmp directory

kgeisz:
https://issues.apache.org/jira/browse/HBASE-30371

## AI Usage

The unit test in the pull request was generated with Claude Opus 4.6

## Summary

This pull request adds an `if` block to `HBackupFileSystem` that allows the `.tmp` directory to be skipped over when each backup root subdirectory is searched for its `.backup.manifest` file.

Before this fix, an error message would be logged during a Point-In-Time Restore saying the backup manifest could not be found in `.tmp`. The `.tmp` directory is not an actual backup directory, so it should not permanently hold this type of file to begin with.


hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/HBackupFileSystem.java
```diff
@@ -158,11 +161,15 @@ public static List<BackupImage> getAllBackupImages(Configuration conf, Path back
       }
 
       String backupId = lfs.getPath().getName();
+      if (HBASE_TEMP_DIRECTORY.equals(backupId)) {
+        continue;
+      }
+
```
wchevreuil, line 167: So why then this would find a path under hbase temp?

kgeisz: That's a good question. I'm looking into this. Based on what I've seen so far, the `backup_XXXX` directory within `backupRoot/.tmp` is getting deleted, but not the `.tmp` dir itself.

kgeisz: It looks like only the bulkoad directory (`backupRoot/.tmp/backup_XXXX`) was getting deleted when an incremental backup finished:
https://github.com/apache/hbase/blob/master/hbase-backup/src/main/java/org/apache/hadoop/hbase/backup/impl/IncrementalTableBackupClient.java#L427

I added a new commit that deletes `backupRoot/.tmp` instead. Should we still have `.tmp` get skipped over in the code here as well as a safeguard?

cc. @ankitsol

anmolnar (approved):
lgtm.

ankitsol:
LGTM

PR #8646, opened 2026-09-10
HBASE-30377 WALInputFormat drops WAL files that span the requested time range

junegunn:
Jira: HBASE-30377

The timestamp in a WAL's name is its creation time, which only bounds its entries from below. A WAL stays open until it rolls, so one created before `startTime` can still hold entries in range, and comparing the name against `startTime` threw the whole file away.

```
startTime = 100, endTime = 200
WAL created at t=50, rolled at t=150  ->  holds entries 50..150
  50 <= 200 passes, but 50 >= 100 fails  ->  file skipped
  lost: every entry in 100..150
```

### Fix

Compare against the modification time instead, which bounds the entries from above. That is only meaningful once the file is closed, since HDFS leaves mtime at the creation time through `hflush` and `hsync`, so it is gated on `DistributedFileSystem.isFileClosed`. Anything we cannot answer for, including non-HDFS filesystems, is treated as open and kept.

### End-to-end test

I verified the fix on a 6-node test cluster running on a local Kubernetes cluster. Without the 1-hour `logroll.period` workaround, WALPlayer correctly processes all WAL files.

### This is safe and conservative

With @mosmeh's suggestion applied, this patch never excludes files that the previous code included.

### DIFF
--- a/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
+++ b/hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
@@ -32,6 +32,7 @@ import org.apache.hadoop.fs.FileSystem;
 import org.apache.hadoop.fs.LocatedFileStatus;
 import org.apache.hadoop.fs.Path;
 import org.apache.hadoop.fs.RemoteIterator;
+import org.apache.hadoop.hbase.fs.HFileSystem;
 import org.apache.hadoop.hbase.regionserver.wal.WALHeaderEOFException;
 import org.apache.hadoop.hbase.util.LeaseNotRecoveredException;
 import org.apache.hadoop.hbase.wal.AbstractFSWALProvider;
@@ -41,6 +42,7 @@ import org.apache.hadoop.hbase.wal.WALEdit;
 import org.apache.hadoop.hbase.wal.WALFactory;
 import org.apache.hadoop.hbase.wal.WALKey;
 import org.apache.hadoop.hbase.wal.WALStreamReader;
+import org.apache.hadoop.hdfs.DistributedFileSystem;
 import org.apache.hadoop.io.Writable;
 import org.apache.hadoop.mapreduce.InputFormat;
 import org.apache.hadoop.mapreduce.InputSplit;
@@ -382,7 +384,7 @@ public class WALInputFormat extends InputFormat<WALKey, WALEdit> {
         // Recurse into sub directories
         result.addAll(getFiles(fs, file.getPath(), startTime, endTime, conf));
       } else {
-        addFile(result, file, startTime, endTime);
+        addFile(result, fs, file, startTime, endTime);
       }
     }
     // TODO: These results should be sorted? Results could be content of recovered.edits directory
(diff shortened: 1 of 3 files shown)


hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
```diff
@@ -391,18 +393,44 @@ List<FileStatus> getFiles(FileSystem fs, Path dir, long startTime, long endTime,
...
+      // range and must not be dropped on the strength of its name alone.
+      if (timestamp > endTime) {
+        LOG.info("Skipped {}, created after endTime [{}/{}]", lfs.getPath(), endTime,
+          Instant.ofEpochMilli(endTime));
+        return;
       }
+      // The modification time is the upper bound, but HDFS leaves it at the creation time until
+      // the file is closed, so it is only meaningful once the file is. Order the checks so the
+      // extra RPC is only paid for files that the modification time alone would prune.
+      if (lfs.getModificationTime() < startTime && isClosed(fs, lfs.getPath())) {
```
mosmeh: This can exclude WALs that the previous code retained, for example, when the NameNode clock (which determines mtime) lags behind the RegionServer clock (which determines the WAL filename timestamp).

How about applying this condition only when `timestamp < startTime`, so it only excludes WALs that the previous code would also have rejected?

```suggestion
      if (timestamp < startTime && lfs.getModificationTime() < startTime && isClosed(fs, lfs.getPath())) {
```

junegunn: Excellent point. Thank you very much.

hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java
```diff
@@ -391,18 +393,44 @@ List<FileStatus> getFiles(FileSystem fs, Path dir, long startTime, long endTime,
...
+      // range and must not be dropped on the strength of its name alone.
+      if (timestamp > endTime) {
+        LOG.info("Skipped {}, created after endTime [{}/{}]", lfs.getPath(), endTime,
+          Instant.ofEpochMilli(endTime));
+        return;
       }
+      // The modification time is the upper bound, but HDFS leaves it at the creation time until
+      // the file is closed, so it is only meaningful once the file is. Order the checks so the
+      // extra RPC is only paid for files that the modification time alone would prune.
+      if (timestamp < startTime && lfs.getModificationTime() < startTime && isClosed(fs, lfs.getPath())) {
```
mosmeh: Sorry for meat-proxying, but Codex says

> [P1] Refresh file status after confirming that the WAL is closed — hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java:372-372
>  If a WAL closes after `listLocatedStatus` captures its status but before `isClosed` runs, `lfs` still contains the open file's creation-time mtime. This check then sees a closed file and incorrectly prunes it despite in-range entries and a newer actual mtime, silently omitting edits from WALPlayer. Re-fetch the file status after confirming closure before using its mtime to prune, retaining the file if that refresh fails.

With the `timestamp < startTime` guard, this patch still is an improvement, though.

junegunn: Ah, yeah, that's possible. Pretty unlikely in real-world use cases though, but it's better to be right.

junegunn: Addressed in 5adb2557f76905aadba3dd7e99dea1996e57dc26. Could you take another look?

hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALRecordReader.java
```diff
@@ -190,9 +190,12 @@ public void testPartialRead() throws Exception {
...
     jobConf.setLong(WALInputFormat.END_TIME_KEY, ts1 + 1);
     splits = input.getSplits(MapreduceTestingShim.createJobContext(jobConf));
-    assertEquals(1, splits.size());
+    assertEquals(2, splits.size());
+    // The 1st file was created before startTime but stayed open until it rolled, so its 2nd
+    // entry, written at exactly startTime, is in-range.
+    testSplit(splits.get(0), Bytes.toBytes("2"));
     // Only the 1st entry from the 2nd file is in-range.
-    testSplit(splits.get(0), Bytes.toBytes("3"));
+    testSplit(splits.get(1), Bytes.toBytes("3"));
```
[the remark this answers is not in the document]
junegunn: WAL files are named with millisecond timestamps, and `listLocatedStatus` iterates in name order on HDFS. This ordering has been stable since the test was written. The line is also pre-existing code, not part of this change.

hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALInputFormat.java
```diff
@@ -65,34 +66,98 @@ public static void setupClass() throws Exception {
...
+    LocatedFileStatus lfs = Mockito.mock(LocatedFileStatus.class);
+    Mockito.when(lfs.getPath()).thenReturn(path);
+    Mockito.when(lfs.getModificationTime()).thenReturn(staleMtime);
+    try {
+      FileStatus refreshed = Mockito.mock(FileStatus.class);
+      Mockito.when(refreshed.getModificationTime()).thenReturn(refreshedMtime);
+      Mockito.when(fs.getFileStatus(path)).thenReturn(refreshed);
+    } catch (IOException e) {
+      throw new RuntimeException(e);
+    }
```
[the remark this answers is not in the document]
junegunn: `isKept` is a `private static` helper that does not declare `throws`. `Mockito.when(...).thenReturn(...)` requires catching the checked `IOException` inside the helper itself. Pre-existing code, not part of this change.

hbase-mapreduce/src/test/java/org/apache/hadoop/hbase/mapreduce/TestWALRecordReader.java
```diff
@@ -245,9 +248,9 @@ public void testWALRecordReader() throws Exception {
     assertEquals(1, splits.size());
     testSplit(splits.get(0), Bytes.toBytes("1"));
 
-    // now set a start time
+    // now set a start time strictly after the last WAL's modification time
     jobConf.setLong(WALInputFormat.END_TIME_KEY, Long.MAX_VALUE);
-    jobConf.setLong(WALInputFormat.START_TIME_KEY, thirdTs);
+    jobConf.setLong(WALInputFormat.START_TIME_KEY, thirdTs + 1);
```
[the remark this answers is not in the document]
junegunn: `thirdTs` is sampled immediately after `log.shutdown()`, which flushes and closes the WAL. The WAL's final modification time is set at close, so `thirdTs >= mtime` holds. Adding `+ 1` makes it strictly greater. The comment describes the relationship correctly. The variable name is pre-existing.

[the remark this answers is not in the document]
junegunn: Already answered above.

junegunn:
Copilot's test suggestions were not relevant, and I'm confident that this PR is now ready for review.

I also asked Codex to review the PR again and it commented:

> One existing limitation remains: `hbase-mapreduce/src/main/java/org/apache/hadoop/hbase/mapreduce/WALInputFormat.java:414` compares NameNode modification time against RegionServer entry timestamps. If the RegionServer clock is ahead, a closed, unarchived WAL can still lose in-range entries during filtering. This case also failed before the PR.

While this is technically possible, I wouldn't worry too much about it in real-world HBase clusters, where NTP is a requirement. More importantly, this is an existing limitation and not a regression introduced by the PR.

Could someone take a look at the patch? Thanks!

mosmeh (approved):

Apache9 (approved):

PR #8651, opened 2026-09-15
HBASE-29309 Document continuous backup and PITR in backup-restore guide

ankitsol:
Update Backup and Restore document to include continuous backup, point-in-time recovery (PITR), related configuration, commands, and administration updates.


hbase-website/app/pages/_docs/docs/_mdx/(multi-page)/backup-restore/additional-topics.mdx
```diff
@@ -201,14 +243,73 @@ HBase Bulk Load utility.
...
+
+WALs in the continuous backup directory must be retained as long as they are required to support the
+configured PITR window and the available backup chain. Backup deletion performs validation to prevent
+deletion of a backup that would leave a table without a valid PITR recovery path, and removes WALs
+that are no longer required.
 
 Like HBase Snapshots, this can have an expectedly large impact on the HDFS usage of HBase for high volume tables. Take care in enabling
 and using the backup and restore feature, specifically with a mind to removing backup sessions when they are not actively being used.
+Running incremental backups on a regular schedule reduces WAL retention on the source cluster for tables under continuous backup,
+because WALs are read from the WAL directory rather than from the source cluster.
```
anmolnar, line 312: I'm a bit confused about this part. In the previous section you mentioned

>When incremental backups are run for tables that have continuous backup enabled, the backup process
reads WALs from the WAL directory instead of collecting them from the source cluster. This eliminates
the need to retain WALs on the source cluster between incremental backup runs and reduces pressure on
the source cluster's log management.

Which means to me that with Continuous Backup we don't need to keep WAL files in the cluster at all, while the above comment suggests that we need to keep WAL files in the cluster between incrementals. Which one is true?

hbase-website/app/pages/_docs/docs/_mdx/(multi-page)/backup-restore/additional-topics.mdx
```diff
@@ -91,75 +138,70 @@ The HBase administration team uses backup sets to store data from a group of tab
 application called green. In this example, one table contains transaction records and the other contains customer details. The
 two tables need to be backed up and be recoverable as a group.
 
-The admin team also wants to ensure daily backups occur automatically.
-
 ![Tables Composing The Backup Set](/docs-images/backup-app-components.png)
 
-The following is an outline of the steps and examples of commands that are used to backup the data for the _green_ application and
-to recover the data later. All commands are run when logged in as HBase superuser.
+The following workflow outlines the steps to back up and recover the _green_ application data. All
```
anmolnar, line 143: Instead of overwriting this section with the more advanced Continuous Backup case, I think we might want to keep the original one as well and outline the differences between the two workflows in a separate example for the same _green_ scenario.

PR #8661, opened 2026-09-16
HBASE-30392 Exclude checksum from onDiskDataSizeWithHeader

Huginn-kio:
### Summary

Keep the size metadata of cache-on-write HFile blocks consistent with the bytes stored in the cache. See [HBASE-30392](https://issues.apache.org/jira/browse/HBASE-30392).

### Root cause

`HFileBlock.Writer#getBlockForCaching` creates the cache buffer with `cloneOnDiskBufferWithHeader()`, which contains only the block header and on-disk data. The metadata was nevertheless populated with `onDiskBlockBytesWithHeader.size() + onDiskChecksum.length`, so `onDiskDataSizeWithHeader` described bytes that are not present in the cache buffer.

### Fix

- Set `onDiskDataSizeWithHeader` to the header-plus-data buffer length.
- Keep `onDiskSizeWithoutHeader` unchanged because the physical on-disk block size still includes checksums.

### Compatibility

- No public API or configuration change.
- No HFile on-disk format change.
- The change only corrects in-memory cache metadata for cache-on-write blocks.

### Testing

- `git diff --check` passed.
- `mvn -pl hbase-server -am -DskipTests -Dcheckstyle.skip=true -Drat.skip=true -Dspotbugs.skip=true compile` passed with JDK 17.


wchevreuil (changes requested):
The name of the property here is "onDiskDataSizeWithHeader" so it should record the proper block size on disk. If we apply this change, we won't be recording the actual block size on disk.

wchevreuil:
> The resulting size mismatch can make cache validation or later block unpacking use an incorrect length, causing cache mismatches or read/decompression failures.

Can you prove this? 

Also, there are other parts of the code that set "onDiskDataSizeWithHeader" within the checksum. If we would apply this proposal, we would probably break block comparison when checking the cache, for cacheOnRead, for example, potentially caching an already cached block again.

Huginn-kio:
> The name of the property here is "onDiskDataSizeWithHeader" so it should record the proper block size on disk. If we apply this change, we won't be recording the actual block size on disk.

Thanks for the comment. I think the distinction between the two header fields was not clear in the PR.

For an HFile block:

- `onDiskSizeWithoutHeader` is the physical on-disk block size excluding the header and including trailing checksum bytes (`D + C`).
- `onDiskDataSizeWithHeader` is the header plus on-disk data before the checksum region (`H + D`).

This is also documented by `Writer#putHeader()`. `cloneOnDiskBufferWithHeader()` returns exactly `H + D` and does not copy `onDiskChecksum`.

Therefore, changing `onDiskDataSizeWithHeader` to `onDiskBlockBytesWithHeader.size()` does not lose the physical block size. The physical size remains represented by `onDiskSizeWithoutHeader`; this change only makes the checksum boundary metadata consistent with the buffer.

The metadata has an extra section the length of the checksum compared to the actual buffer, causing the buffer boundary and header metadata to be inconsistent, which led to out-of-bounds access when we implemented fast-path decompression.

Whether on cache-on-read or cache-on-write, the actual cached buffer does not contain a checksum. I agree with your point about needing to be compatible with already cached blocks, but I think this semantics ultimately still needs to be revised.

PR #21, opened 2015-09-17
Upgrade to a recent version of Guava (18.0)

oliviertoupin:
I updated Guava to version 18. We had a version clash in a project we wanted to include hbase-client in.

You might not want to merge this particular maintenance branch, but I thought it might interests you.

PR #8676, opened 2026-09-17
Create separate Jenkinsfile for Read-Replica feature

kgeisz:
- Removes Read-Replica stage from `dev-support/Jenkinsfile`
- Creates a new `dev-support/read-replica/Jenkinsfile` for running the read-replica integration tests
- Moves `dev-support/hbase_nightly_read_replica_test.sh` -> `dev-support/read-replica/hbase_nightly_read_replica_test.sh`
- Updates paths in `hbase_nightly_read_replica_test.sh` accordingly since the script was moved
- Removes `SCRIPT_DIR` variable from `hbase_nightly_read_replica_test.sh` since it is now redundant with `REPLICA_DIR` variable 
- Updates documentation in `dev-support/README.md` and `dev-support/read-replica/README.md`


dev-support/read-replica/.env
```diff
@@ -22,7 +22,7 @@ HBASE_CONTAINER_NAME=hbase-docker
 # This is the host running the hbase-docker containers. Use localhost if the containers
 # are running locally. If they are started by another container, such as a Jenkins
 # container in a Docker-out-of-Docker setup, then try setting this to host.docker.internal.
-HBASE_HOST=${HBASE_HOST:-localhost}
+HBASE_HOST=${HBASE_HOST:-host.docker.internal}
```
kgeisz, line 25: @taklwu this needs to be set to `host.docker.internal` in order to get it running in my local Jenkins setup. We will see how this goes with the upstream Jenkins.

`HBASE_HOST` is used for curling the HBase UI to see if it's up. It's part of the process of verifying HBase has started. ([here it is in the code](https://github.com/kgeisz/hbase/blob/HBASE-30087-read-replica-jenkinsfile-test/dev-support/read-replica/python/src/hbase_docker_client.py#L146))

The only problem with this is if I want to run the tests locally with `hbase_nightly_read_replica_test.sh` (without Jenkins) then `HBASE_HOST` should be `localhost`. We may want to think of a way to make it easier for a user to run this without having to know they should modify `HBASE_HOST` to make it `localhost`.

taklwu: can we use something like `hostname -a` then it could auto detect the hostname ?

taklwu: seems like you can use it in `build-images.sh` line #21  and overwrite the hostname? or is it how the `.env` is being used?

```
. ./.env
```

dev-support/read-replica/.env
taklwu, line 19: so, are we always use default? or `kgeisz` is being used? 

I'm thinking if we should make it more like `default` instead of using a user like prefix.

taklwu (changes requested):

dev-support/read-replica/hbase_nightly_read_replica_test.sh
```diff
@@ -68,13 +66,6 @@ echo "DOCKER_COMPOSE_FILE=${DOCKER_COMPOSE_FILE}"
 echo "HBASE_DATA_STORE_ROOT=${HBASE_DATA_STORE_ROOT}"
 echo "realpath of HBASE_DATA_STORE_ROOT=$(realpath ${HBASE_DATA_STORE_ROOT})"
 
-echo "Removing HBase log directories from mounted volumes that may exist from a previous test run:"
```
taklwu, line 71: why don't we need this now?

dev-support/Jenkinsfile
```diff
@@ -323,7 +321,7 @@ pipeline {
                 else
                   echo "No archiver directory, skipping compressing."
                 fi
-              '''
+'''
```
taklwu: Do we need this indentation change? if not please remove them and below

taklwu (changes requested):
please remove those minor indentation change, otherwise LGTM

dev-support/read-replica/python/src/hbase_docker_client.py
```diff
@@ -141,26 +136,27 @@ def _get_pid_from_jps(self, process_name: str) -> int | None:
...
     def wait_for_hbase_ui(self) -> bool:
-        """Checks for a 200 OK on the HBase Master UI."""
-        # Read HBASE_HOST from environment, falling back to 'localhost' for host-native execution
-        url = f"http://{self._hbase_host}:{self._hbase_ui_port}"
-        logger.info(f"Waiting for HBase UI: {self._cluster_name} on {url}")
-        last_exception = None
+        """
+        Checks for a 200 OK on the HBase Master UI inside the container using curl.
+        """
+        check_cmd = f"curl -s -f --max-time 3 http://{self._container_name}:{self._hbase_ui_port} > /dev/null"
```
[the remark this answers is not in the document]
kgeisz: @taklwu, IMO this can be ignored. `curl` is a dependency of the `rpm` package, which will always be included in the image.

From inside the Docker container:
```
$ rpm -q --whatrequires curl
rpm-4.14.3-32.el8_10.aarch64
```

dev-support/read-replica/hbase_nightly_read_replica_test.sh
kgeisz: Originally, before I was using `pytest`, I was saving the HBase logs regardless of if the test passed or failed. We decided I don't need save any logs anymore since the user can reproduce any errors locally by running `hbase_nightly_read_replica_test.sh`.

If you think it would be good to save the logs, then I can add it back in. I could maybe even save separate logs for each unit test.

dev-support/Jenkinsfile
kgeisz: No, I didn't realize I still had that change there. I'll remove it. Thanks

taklwu (approved):
LGTM

PR #8677, opened 2026-09-18
HBASE-30357 OpenRegionProcedure#restoreSucceedState ignores persisted transitionCode, forcing OPEN even after a real FAILED_OPEN

mnpoonia:
## Summary
- Backport of the master fix (HBASE-30357) to branch-2.
- `OpenRegionProcedure#restoreSucceedState` was ignoring the persisted `TransitionCode`, always forcing the region state to OPEN on master-restart restore even when the persisted transition was actually `FAILED_OPEN`. Now checks `transitionCode == FAILED_OPEN` and calls `am.regionFailedOpen(regionNode, false)`, matching the live `reportTransition` path.
- Threaded `TransitionCode` through `RegionRemoteProcedureBase#restoreSucceedState`/`stateLoaded` so both `OpenRegionProcedure` and `CloseRegionProcedure` receive it (close has no failure variant, so it's unused there).
- Added `TestOpenRegionProcedureRestoreFailedOpen`, which restarts a real `ProcedureExecutor` to reproduce the crash window.

## Test plan
- [x] Ported test compiles against branch-2 APIs (`HBaseTestingUtility`, JUnit5 `HBaseJupiterExtension`).
- [ ] Note: local run of the new test on this branch-2 checkout (JDK8 + `-Dhadoop.profile=3.0`) is flaky due to a pre-existing race in the test's synchronization (a non-interruptible `regionNode.lock()` held across an RPC call can deadlock with `ProcedureExecutor#stop()/join()` until a 30s await times out); this race is inherited from the already-merged master version of this test and is being tracked separately rather than fixed in this backport.


mnpoonia:
branch-2 test is failing because branch-2 doesnot have  HBASE-28199 HBASE-28240.
have to rethink how to write it for branch-2
@virajjasani

mnpoonia:
@virajjasani is it okay if we don't include test in branch-2?

virajjasani:
> @virajjasani is it okay if we don't include test in branch-2?

We can include test and keep it and add `@ignored` with reason

Apache9:
So what is the problem here? The test can not work on branch-2?

mnpoonia:
@Apache9 Yes, the test can not run reliably on branch-2.

This test reproduces a real crash window: the RPC-handler thread holds the `RegionStateNode` lock while the master's `ProcedureExecutor` restarts, simulating a failover mid-transition.

On master, HBASE-28199/HBASE-28240 replaced `RegionStateNode`'s plain `ReentrantLock` with a suspend-capable `RegionStateNodeLock`, so a contended lock suspends the procedure instead of blocking the worker thread - `ProcedureExecutor.stop()` joins cleanly.

Branch-2 never got that backport, so the same setup deadlocks for real: the `PEWorker` blocks on the lock, `stop()` blocks on the worker, and the RPC-handler blocks on the main thread - broken only by the test's 30s timeout unwinding it. The test "passing" on branch-2 depends on hitting that timeout every run, not on any deterministic fix logic, so it isn't a meaningful regression test there.

Backporting the suspend-lock redesign is out of scope for this JIRA (it spans `RegionStateNode`, `RegionRemoteProcedureBase`, `TransitRegionStateProcedure`, likely SCP too). Restructuring the test to dodge the deadlock would exercise branch-2's old synchronous meta-update path concurrently - an unverified path master never hits, so it wouldn't provide equivalent coverage.

Per @virajjasani's suggestion, I've added `@Disabled` on the branch-2 test with a comment documenting the missing backport, rather than keeping a test whose pass/fail is really just a timeout race.

PR #8681, opened 2026-09-18
HBASE-30387: Support SingleEKU certificates

lucakovacs:


PDavid:
The newly added `org.apache.hadoop.hbase.rest.TestRESTServerSSL.testClientAuthNeedRejectsClientWithoutCert` test seems to fail. Can you please have a look?

PDavid (approved):
Many thanks, looks really good. 👍

hbase-common/src/main/java/org/apache/hadoop/hbase/io/crypto/tls/X509Util.java
```diff
@@ -175,20 +288,27 @@ public static SslContext createSslContextForClient(Configuration config)
...
 
-    String trustStoreLocation = config.get(TLS_CONFIG_TRUSTSTORE_LOCATION, "");
-    char[] trustStorePassword = config.getPassword(TLS_CONFIG_TRUSTSTORE_PASSWORD);
-    String trustStoreType = config.get(TLS_CONFIG_TRUSTSTORE_TYPE, "");
+    String trustStoreLocation = resolveConfig(config, TLS_CONFIG_CLIENT_TRUSTSTORE_LOCATION,
+      TLS_CONFIG_TRUSTSTORE_LOCATION, "");
+    char[] trustStorePassword = resolvePassword(config, TLS_CONFIG_CLIENT_TRUSTSTORE_PASSWORD,
+      TLS_CONFIG_TRUSTSTORE_PASSWORD);
+    String trustStoreType =
+      resolveConfig(config, TLS_CONFIG_CLIENT_TRUSTSTORE_TYPE, TLS_CONFIG_TRUSTSTORE_TYPE, "");
```
petersomogyi: The fallback happens per config key. Setting only `hbase.rpc.tls.client.keystore.location` means that legacy password and type will be matched with it causing an error. These should not be combined.

hbase-http/src/main/java/org/apache/hadoop/hbase/http/InfoServer.java
```diff
@@ -83,6 +94,16 @@ public InfoServer(String name, String bindAddress, int port, boolean findPort,
...
         .setIncludeCiphers(getTLSProperty(c, "include.cipher.list"))
         .setExcludeCiphers(getTLSProperty(c, "exclude.cipher.list"));
+
+      // Activate mutual TLS if configured. Default is NONE, which preserves today's behavior of
+      // never requesting a client certificate on the UI connector (leaving any configured
+      // truststore inert for peer verification). Set hbase.ui.ssl.server.client.auth.mode to
+      // WANT or NEED to opt in. The client.auth.mode key is looked up directly on
+      // HBASE_WEB_TLS_SERVER_CONFIG_PREFIX; there is no legacy or Hadoop-prefixed fallback.
+      X509Util.ClientAuth clientAuth = X509Util.ClientAuth
+        .fromPropertyValue(c.get(HBASE_UI_SSL_CLIENT_AUTH_MODE, X509Util.ClientAuth.NONE.name()));
```
petersomogyi: It might be an edge case but when the config is in `hbase-site.xml` with empty value then it will use `NEED` not `NONE` when the config is not there at all.

```
  /**
     * Converts a property value to a ClientAuth enum. If the input string is empty or null, returns
     * <code>ClientAuth.NEED</code>.
     * @param prop the property string.
     * @return the ClientAuth.
     * @throws IllegalArgumentException if the property value is not "NONE", "WANT", "NEED", or
     *                                  empty/null.
     */
```

hbase-thrift/src/main/java/org/apache/hadoop/hbase/thrift/ThriftServer.java
```diff
@@ -413,16 +423,58 @@ protected void setupHTTPServer() throws IOException {
...
       sslCtxFactory.setKeyManagerPassword(keyPassword);
       sslCtxFactory
-        .setKeyStoreType(conf.get(THRIFT_SSL_KEYSTORE_TYPE_KEY, THRIFT_SSL_KEYSTORE_TYPE_DEFAULT));
+        .setKeyStoreType(X509Util.resolveConfig(conf, THRIFT_SSL_SERVER_KEYSTORE_TYPE_KEY,
+          THRIFT_SSL_KEYSTORE_TYPE_KEY, THRIFT_SSL_KEYSTORE_TYPE_DEFAULT));
+
+      // Truststore is entirely new for Thrift — no legacy fallback because there is no historical
+      // hbase.thrift.ssl.truststore.* configuration. When left unset, no truststore is configured
+      // on the connector and any hbase.thrift.ssl.server.client.auth.mode = WANT/NEED setting
+      // will fail the handshake for lack of a peer-cert trust root.
```
petersomogyi: Opus 5.5 called out this: Jetty falls back to the JVM cacerts, so any publicly-CA-signed client cert is accepted and it will not cause a handshake fail. It should fail at startup if configs are missing or WARN about it to let the operator aware of the configuration issue.

petersomogyi: Same problem is present in REST and InfoServer.

hbase-http/src/test/java/org/apache/hadoop/hbase/http/TestInfoServerTLSConfig.java
```diff
@@ -0,0 +1,100 @@
...
+    c.unset(HADOOP_KEY.replace("keystore.location", "keystore.password"));
+    assertNull(InfoServer.getTLSPassword(c, "keystore.password"));
+  }
+
+  @Test
+  public void testClientAuthModeKeyIsRoleScoped() {
+    // Guard against a "double server.server." regression: the client-auth-mode config key must
+    // resolve to the single role-scoped key, not to a nested/prefixed form.
+    assertEquals("hbase.ui.ssl.server.client.auth.mode", InfoServer.HBASE_UI_SSL_CLIENT_AUTH_MODE);
+  }
```
petersomogyi: What is the point of this test?

hbase-common/src/test/java/org/apache/hadoop/hbase/io/crypto/tls/TestX509Util.java
```diff
@@ -375,4 +376,142 @@ public void testLoadPKCS12TrustStoreWithWrongPassword() {
...
+  public void testCreateSSLContextForClientFallsBackToLegacyKeystore() throws Exception {
+    // The base setUp() only sets the legacy TLS_CONFIG_KEYSTORE_* / TLS_CONFIG_TRUSTSTORE_* keys.
+    // The role-scoped keys are intentionally unset; the context must still build using the legacy
+    // values (backward-compat regression guard).
+    conf.unset(X509Util.TLS_CONFIG_CLIENT_KEYSTORE_LOCATION);
+    conf.unset(X509Util.TLS_CONFIG_CLIENT_KEYSTORE_PASSWORD);
+    conf.unset(X509Util.TLS_CONFIG_CLIENT_KEYSTORE_TYPE);
+    conf.unset(X509Util.TLS_CONFIG_CLIENT_TRUSTSTORE_LOCATION);
+    conf.unset(X509Util.TLS_CONFIG_CLIENT_TRUSTSTORE_PASSWORD);
+    conf.unset(X509Util.TLS_CONFIG_CLIENT_TRUSTSTORE_TYPE);
```
petersomogyi: Isn't this a no-op?
Apart from that this test is the same as `testCreateSSLContextWithoutCustomProtocol` but that also validates the TLS protocols, not just if it is not empty.

hbase-common/src/test/java/org/apache/hadoop/hbase/io/crypto/tls/TestX509Util.java
```diff
@@ -375,4 +376,142 @@ public void testLoadPKCS12TrustStoreWithWrongPassword() {
...
+    conf.unset(X509Util.TLS_CONFIG_KEYSTORE_TYPE);
+
+    SslContext sslContext = X509Util.createSslContextForServer(conf);
+    ByteBufAllocator byteBufAllocatorMock = mock(ByteBufAllocator.class);
+    assertTrue(
+      sslContext.newEngine(byteBufAllocatorMock).getSSLParameters().getProtocols().length > 0);
+  }
+
+  @TestTemplate
+  public void testCreateSSLContextForServerFallsBackToLegacyKeystore() throws Exception {
```
petersomogyi: Same comment as testCreateSSLContextForClientFallsBackToLegacyKeystore

petersomogyi (changes requested):
Could you extend the TLS documentation?
https://github.com/apache/hbase/blob/master/hbase-website/app/pages/_docs/docs/_mdx/(multi-page)/security/tls.mdx

hbase-rest/src/test/java/org/apache/hadoop/hbase/rest/TestRESTServerSSL.java
```diff
@@ -160,6 +161,98 @@ public void testSslConnectionUsingKeystoreFormatPKCS12() throws Exception {
...
+  /**
+   * With {@code client.auth.mode=NEED}, an anonymous client (no client cert) is rejected by the
+   * server. The base {@link Client} configures truststore-only, so it presents no key material to
+   * the server.
+   */
+  @Test
+  public void testClientAuthNeedRejectsClientWithoutCert() throws Exception {
+    conf.set(Constants.REST_SSL_CLIENT_AUTH_MODE, "NEED");
+    startRESTServerWithDefaultKeystoreType();
+    assertThrows(SSLException.class, () -> sslClient.get("/version"));
```
petersomogyi: This should be changed to IOException like in TestThriftServerSSLMutualAuth.

PR #8682, opened 2026-09-18
HBASE-30220: A replica cluster can have read-only mode disabled even when another active cluster already exists

kgeisz:
https://issues.apache.org/jira/browse/HBASE-30220

HBASE-30220 was merged into `master` in #8377.  This PR back-ports HBASE-30220 into `branch-3`.


anmolnar (approved):
lgtm.

anmolnar:
Merged. Thanks @kgeisz !

